PR #2, opened 2023-02-03
add rust toolchain file

franziskuskiefer:
Adding a rust toolchain file to make it possible (easier) to build the thir-export.
Unfortunately it looks like the crate itself doesn't work yet. So I'll refrain from digging deeper until we had a chat about what works and what needs to happen to make it work.


W95Psp:
Hi Franziskus,

About the `Cargo.lock` file, we actually need to commit it, Nix uses the lock information so that the thing is always built the same way with the exact same dependencies.

Oh, what is not working? It's not compiling?

franziskuskiefer:
> About the `Cargo.lock` file, we actually need to commit it, Nix uses the lock information so that the thing is always built the same way with the exact same dependencies.

As long as the libraries are only used in the local binary that's fine. Otherwise it's dangerous because you won't know whether it's actually building for other versions of the dependencies.

> Oh, what is not working? It's not compiling?

It's compiling but it's only throwing errors.
Can you start by adding documentation here on how you expect this to be used?

PR #5, opened 2023-03-03
Coq backend

cmester0:


karthikbhargavan (approved):
Looks good for now. Of course, we'll want to iterate on this and the other backends before a proper release.

PR #9, opened 2023-03-23
Do not call union_spans for empty tuple

### LINKED ISSUE #8: Calling function with unit return type
There seems to be an issue with the unification of spans. The following code fails in the "Desugar_direct_and_mut" phase.
```rust
pub fn a_test () {

}

pub fn b_test () {
    a_test()
}
```
This is because `List.reduce_exn` is given an empty list.

cmester0:
Fix #8

### DIFF
--- a/thir-elab/lib/ast_utils.ml
+++ b/thir-elab/lib/ast_utils.ml
@@ -288,8 +288,11 @@ module Make (F : Features.T) = struct
 
   let make_tuple_pat (pats : pat list) : pat =
     let len = List.length pats in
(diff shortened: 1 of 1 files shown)


W95Psp:
Hi Lasse, thanks!
Though I think calling `union_span` with zero spans should return the `Dummy` range, so I'd rather fix that function directly.
(and that removes a call to an exception-yielding function, which always feels nice!)

W95Psp:
just pushed a quick commit that patches the function directly. Closing this PR thus.

Issue #4, opened 2023-02-07
Add CI


franziskuskiefer:
Add CI for all parts in this repository to ensure things don't break.

PR #41, opened 2023-04-13
Add CI

W95Psp:
Related to #4


.github/workflows/specs.yml
```diff
@@ -0,0 +1,81 @@
+name: Test extractions for `hacpsec/specs`
+
+on:
+  push:
+    branches:
+    - '*'
```
franziskuskiefer, line 6: Let's run this on PRs and on demand as well.
At some point this shouldn't run on branches anymore, but that's ok for now.

```suggestion
  pull_request:
  workflow_dispatch:
```

.github/workflows/specs.yml
```diff
@@ -0,0 +1,81 @@
...
+    - uses: cachix/cachix-action@v12
+      with:
+        name: hacspec
+        authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+    - name: Build
+      run: nix build
+
+    - name: Install
+      run: |
+        nix profile install nixpkgs#yq
```
franziskuskiefer, line 23: Please document in #4 that this is using nix for now and that we need to add this without nix as well.

.github/workflows/compile_engine.yml
```diff
@@ -0,0 +1,53 @@
...
+jobs:
+  release:
+    if: startsWith(github.ref, 'refs/tags/')
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - macos-latest
+          - ubuntu-latest
+          # - windows-latest
```
franziskuskiefer, line 17: Windows isn't working?

.github/workflows/compile_engine.yml
```diff
@@ -0,0 +1,53 @@
...
+      contents: write
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - uses: ocaml/setup-ocaml@v2
+        with:
+          ocaml-compiler: ${{ matrix.ocaml-compiler }}
+
+      - uses: actions-rs/toolchain@v1
```
franziskuskiefer, line 31: `actions-rs` is unmaintained. You can use `dtolnay/rust-toolchain` if you want to use a toolchain action.

.github/workflows/compile_engine.yml
```diff
@@ -0,0 +1,53 @@
...
+        id: rust_toolchain
+        with:
+          toolchain: stable
+          
+      - run: cargo install --path cli
```
franziskuskiefer, line 36: Why do we need the cli when building the engine?

.github/workflows/compile_engine.yml
```diff
@@ -0,0 +1,53 @@
...
+
+      - run: opam install . --deps-only
```
franziskuskiefer, line 38: How do we compile for m1 here? Is there some `--target` in opam/dune?

franziskuskiefer (changes requested):
Thanks!
Generally lgtm with a few nits. The unmaintained dependency is the main point that needs to get fixed.
Can you also update #4 with what's in here and what still needs to be done?

.github/workflows/specs.yml
W95Psp: (I edited #4)

.github/workflows/compile_engine.yml
W95Psp: It was not due to incompatibility of some library, but I might I have fixed that, will check

.github/workflows/compile_engine.yml
W95Psp: The Rust part (aka the cli) gives the types to the ML part, it provides a binary `cargo-circus-export-json-schema` (or something) that outputs JSON schema, which is then processed by a script (the dirty JS one) called by dune to create a ML module of input types

.github/workflows/compile_engine.yml
W95Psp: I have absolutely no idea :/
Also, the binaries I produce are not static, and it seems not that easy to make OCaml output fully static binaries...

That's why for now I just compile to JS, I think that's a good fallback solution (`nodejs` is available basically on every plateform)

.github/workflows/compile_engine.yml
W95Psp: So, turns out Opam is not able to install `gmp` or `pkgconf`, I'll mark that as a TODO in the issue #4

karthikbhargavan:
Let's merge this soon and then improve over time.

.github/workflows/compile_engine.yml
franziskuskiefer: ok, let's use this for now and discuss how to do this.

.github/workflows/specs.yml
franziskuskiefer: We'll need to make sure that things work for everyone. But let's discuss and do something in a follow up.

franziskuskiefer (approved):
Let's get this.

PR #56, opened 2023-04-20
circus linter

franziskuskiefer:
Infrastructure and first basic linter.


cli/driver/src/driver.rs
```diff
@@ -60,15 +64,36 @@ fn main() {
...
+        }
+        None => {
+            // hacspec lint
+            let mut callbacks = LinterCallbacks {};
+
+            std::process::exit(rustc_driver::catch_with_exit_code(move || {
+                rustc_driver::RunCompiler::new(&rustc_args, &mut callbacks).run()
+            }))
+        }
+    };
```
W95Psp, line 98: Why not:
```rust
    let mut callbacks: Box<dyn rustc_driver::Callbacks + Send> = match options.command {
        Some(Command::ExporterCommand(command)) => Box::new(exporter::ExtractionCallbacks {
            output_dir: options.output_dir,
            inline_macro_calls: options.inline_macro_calls,
            command,
        }),
        Some(Command::LintCommand(circus_cli_options::LinterCommand::Rust)) => {
            Box::new(LinterCallbacks {})
        }
        Some(Command::LintCommand(circus_cli_options::LinterCommand::Hacspec)) | None => {
            Box::new(LinterCallbacks {})
        }
    };

    std::process::exit(rustc_driver::catch_with_exit_code(move || {
        rustc_driver::RunCompiler::new(&rustc_args, &mut *callbacks).run()
    }))
```

W95Psp (approved):
I've got only one comment about `driver.rs`

Otherwise, I guess in another PR we'll have to transfer `error.rs` error cases into the `kind` enum & add pretty printing to it

cli/driver/src/driver.rs
franziskuskiefer: Re `Box`: I guess I was lazy ... done that now.
But we have to have an option here if `options.command` should be a subcommand. I don't think we can do a default subcommand (only args). So the options needs to stay for the default value.
Note that there's no config here yet. But this will change once the linter knows about the different things it should lint for.

PR #64, opened 2023-04-21
Engine: improve diagnostics

W95Psp:
This PR:
 - fixes the JSON Schema script for enum serialization;
 - adds some tests for the script;
 - simplify & change the phases and subtype module a bit to propagate spans everywhere so that we can report errors nicely


engine/utils/ocaml_of_json_schema/test/rust/src/lib.rs
```diff
@@ -0,0 +1,39 @@
+use schemars::JsonSchema;
```
franziskuskiefer, line 1: Please add some comments to all files in this crate what they are doing.

engine/utils/ocaml_of_json_schema/test/rust/src/tester.rs
```diff
@@ -0,0 +1,33 @@
+mod lib;
+// use proptest::arbitrary::*;
```
franziskuskiefer, line 2:
```suggestion
```

engine/utils/ocaml_of_json_schema/test/rust/Cargo.toml
franziskuskiefer, line 1: Can you add a readme.md or something here to say something about what it is and how it should be used?

engine/utils/ocaml_of_json_schema/test/ocaml/bin/dune
```diff
@@ -0,0 +1,20 @@
...
+   types.ml
+   (pipe-stdout
+        (run ocaml_of_json_schema-schema)
+        (run node %{ocaml_of_json_schema} - -)
+   ))))
+
+(env
+ (_
+  (flags
+   (:standard -w -A -w +8))))
```
franziskuskiefer, line 20:
```suggestion
   (:standard -w -A -w +8))))

```

franziskuskiefer (approved):
Thanks, lgtm to me after adding a little more docs.

engine/utils/ocaml_of_json_schema/test/rust/src/lib.rs
W95Psp: I added a README that describes the thing, I think that's enough, what do you think?

franziskuskiefer: Generally all the code needs comments as well (readme is high level doc also for someone just using it, comments in the code is for anyone working on the code including future selves).
I'll complain about any undocumented code but won't block on it for now in favour of speed.

W95Psp: Yes, I agree, I basically wrote no comment while rushing adding features.
Ok, nice, thanks!

Issue #42, opened 2023-04-14
Replace `ocaml_of_json_schema.js` with something cleaner


W95Psp:
`ocaml_of_json_schema.js` is a script that takes a JSON Schema created by [schemars](https://docs.rs/schemars/latest/schemars/) and outputs an OCaml module with type definitions, [(yo)JSON](https://github.com/ocaml-community/yojson) serializers and deserializers.

We would like to drop the dependency on NodeJS at build time, so the idea here would be to replace that script with an OCaml (or Rust) program.

PR #68, opened 2023-04-21
`ocaml_of_json_schema`: remove tests

W95Psp:
Seems like any `dune` project in any directory under `engine` gets compiled with `dune build`...

This test folder is not very important, and we want to replace that script (see #42) at some point anyway.


franziskuskiefer (approved):
😔

PR #70, opened 2023-04-24
start tests

franziskuskiefer:
We need tests for each part. I started a generic `tests` directory. We might want to split this into frontend/engine tests.
There's also a test case in there that is failing right now.
This should get added to the CI as well.

I also enabled logging to allow debugging.

### DIFF
--- a/cli/subcommands/src/cargo_circus.rs
+++ b/cli/subcommands/src/cargo_circus.rs
@@ -1,6 +1,9 @@
 use circus_cli_options::Options;
 use clap::Parser;
-use std::process::Command;
+use std::{
+    io::{self, Write},
+    process::Command,
+};
 
 /// Return a [Command] for [cargo]: when the correct nightly is
 /// already present, this is just the command [cargo], otherwise we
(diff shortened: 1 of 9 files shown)


cli/subcommands/src/cargo_circus.rs
```diff
@@ -38,20 +41,18 @@ fn main() {
...
+            circus_cli_options::ENV_VAR_OPTIONS_FRONTEND,
+            serde_json::to_string(&options)
+                .expect("Options could not be converted to a JSON string"),
+        )
+        .spawn()
+        .unwrap();
+    let output = child.wait_with_output().unwrap();
+    io::stdout().write_all(&output.stdout).unwrap();
+    io::stderr().write_all(&output.stderr).unwrap();
+    std::process::exit(output.status.code().unwrap_or(254))
```
W95Psp, line 57: This will run `RUSTC_WORKSPACE_WRAPPER=... cargo build ...`, wait (`wait_with_output` being synchronous) for it to run entirely, and then, will write stdout & stderr atomically.
I think this is not what we want right?
The default behavior of spawn is to redirect stderr & stdout to the parent process, I think we don't need to play manually (& we should not) with stderr and stdout here

franziskuskiefer: Right now you don't get any output because you just spawn it.
Feel free to improve on this. But we need to get the output from the binary and this was the easiest to make it happen at all.

W95Psp: But wait, currently, `cargo circus` does redirect outputs: we get both the diagnostics from the linter and the errors from the engine (which is one process deeper)
(and the [doc of spawn](https://doc.rust-lang.org/std/process/struct.Command.html#method.spawn) says the default behavior is to inherit stdout & stderr, so that should be okay)

franziskuskiefer: Oh, well I guess the fact that circus only runs after cleaning makes this a little hard to test ...

W95Psp (approved):

PR #72, opened 2023-04-25
Fix `cargo` and `rustc` cache: factorize caching logic

W95Psp:


franziskuskiefer (approved):
Thanks!
It appears to fix the caching issues.

bors r+

PR #73, opened 2023-04-25
Add use statement

cmester0:
Add `use` statements (with resolution information). Imports in coq still need some generalization (currently assume imports are in the `Examples` package), and the updated coq library for hacspec needs to be put somewhere.

### DIFF
--- a/engine/lib/ast.ml
+++ b/engine/lib/ast.ml
@@ -553,7 +553,11 @@ functor
           of_trait : (global_ident * generic_value list) option;
           items : impl_item list;
         }
-      | Use of string list * string * bool * string option
+      | Use of {
+          path : string list;
+          is_external : bool;
+          rename : string option;
+        }
       | NotImplementedYet
 
     and item = {
--- a/engine/lib/ast.ml
+++ b/engine/lib/ast.ml
@@ -519,38 +519,6 @@ functor
       constraints : generic_constraint list;
     }
 
-    and def_path_item =
-      | PathCrateRoot
-      | PathImpl
-      | PathForeignMod
-      | PathUse
-      | PathGlobalAsm
-      | PathClosureExpr
-      | PathCtor
-      | PathAnonConst
-      | PathImplTrait
-      | PathTypeNs of string
-      | PathValueNs of string
-      | PathMacroNs of string
-      | PathLifetimeNs of string
-
-    and def_id = { krate : string; def_path : def_path_item list }
-
-    and use_res =
-      | ToolMod
-      | Err
-      | Def of (string * def_id)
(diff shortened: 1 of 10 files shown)


cmester0:
Implements part of the coq backend for #74.

W95Psp:
Thanks for the PR! :)

I'm looking into this, and I'm wondering: do you need `use` items because you actually need to generate Requires in Coq? (i.e. otherwise modules are not in scope)

In this case, `use` items is not what you want I think. I mean, in Rust one might just write fully qualified paths without any use.
I think we should have a phase/util that browses an AST and lists all the items used that live outside of the current crate.

Also, it seems that at the HIR/THIR phase, since name resolution is already there, every use .. as .. items are just gone :(

cmester0:
I think the information is in the resolution objects that are part of `UsePath`s? There at least you can check that it is a `use` statement.

W95Psp:
Well, what I mean is that all user-defined naming is gone at HIR/THIR time.
So, when you do `use foo::bar as hello`, at HIR we actually get `use foo::bar`. 
This is a bit of an issue, because we usually do renaming when there are name clashes or to make things clearer.

But anyway, these kinds of issues are not very important, it's only about making paths nicer.

The important thing is: do you need the list of external crate used in an AST so that you can generate `Require`s in Coq? (because in this case, the current phases / utils / AST doesn't give enough information indeed!)

cmester0:
I found a way to get the `rename` information back, it is stored as the ident on item objects representing `use` statements (https://github.com/rust-lang/rust/blob/8b8110e1469d459a196f6feb60d82dec48c3cfc2/compiler/rustc_ast_lowering/src/item.rs#L563-L569), will try and add it to the path objects.

engine/lib/ast.ml
```diff
@@ -553,6 +553,7 @@ functor
           of_trait : (global_ident * generic_value list) option;
           items : impl_item list;
         }
+      | Use of string list * string * bool * string option
```
W95Psp, line 556: For uniformity, I'd prefer a record payload here, i.e.
```suggestion
      | Use of { path: string list; external: bool; rename: string }
```

About the `string` field for Single, Glob, etc.: I don't really see how we could use this information. Do you have ideas in mind?

W95Psp (commented):
just leaving some quick comment (even if it's still draft)

engine/lib/ast.ml
cmester0: I agree, I just did not want to throw away potential useful information. However this can be re-added if it is ever needed.

W95Psp (approved):
LGTM, thanks! 😃

PR #75, opened 2023-04-25
Update coq backend

cmester0:


engine/backends/coq/coq_backend.ml
```diff
@@ -472,32 +474,31 @@ module CoqBackend = struct
...
-          pre ^ ty_name ^ " " ^ ":" ^ " " ^ ty_str ^ post ^ variants_str )
-    | _ -> ([], "")
+    List.fold_left
+      ~init:([],"")
+      ~f:(function | (variant_definitions, variants_str) ->
+          function | (ty_name, ty) ->
+                      let ty_definitions, ty_str = ty_to_string ty in
+                      ( ty_definitions @ variant_definitions,
+                        pre ^ ty_name ^ " " ^ ":" ^ " " ^ ty_str ^ post ^ variants_str ))
+      variants
```
W95Psp, line 501: why not this:
```suggestion
      ~f:(fun (variant_definitions, variants_str) (ty_name, ty) ->
                      let ty_definitions, ty_str = ty_to_string ty in
                      ( ty_definitions @ variant_definitions,
                        pre ^ ty_name ^ " " ^ ":" ^ " " ^ ty_str ^ post ^ variants_str ))
      variants
```

engine/backends/coq/coq_backend.ml
```diff
@@ -8,6 +8,8 @@ module CoqBackend = struct
   module InputLanguage = struct
     open Features
     include Off
+    (* include On.Monadic_action *)
+    include On.Early_exit
```
W95Psp, line 12: Are you translating early exits for real, or is that just for debugging purposes?

W95Psp (approved):
I've got some remarks, but otherwise, that looks good to me!

Can you run `dune fmt` on the engine code?

W95Psp:
Looking a bit more into this:
do you need `use` items because you actually need to generate `Require`s in Coq? (i.e. otherwise modules are not in scope)

In this case, `use` items is not what you want I think. I mean, in Rust one might just write fully qualified paths without any `use`.
I think we should have a phase/util that browses an AST and lists all the items used that live outside of the current crate.  

Also, it seems that at the HIR/THIR phase, since name resolution is already there, every `use .. as ..` items are just gone :(

engine/backends/coq/coq_backend.ml
cmester0: I wanted to start trying, however it seems there are still things to do, before we can actually try translating to the backends?

engine/backends/coq/coq_backend.ml
```diff
@@ -835,21 +837,24 @@ module CoqBackend = struct
...
+         [ C.AST.Notation (pglobal_ident name ^ "_t", pty ty) ]
       (* record *)
       | Type
           {
             name;
             generics;
             variants = [ { name = record_name; arguments } ];
             record = true;
-          } ->
+          } when name == record_name ->
```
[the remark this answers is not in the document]
cmester0: Hmm, this is equality between `global_ident`s so I think it might work, but I will do something better.

engine/backends/coq/coq_backend.ml
W95Psp: Well, yeah, I'm currently writing a phase that transforms `return`s, `continue`s and `break`s into monadic let bindings.
I think that should really be done in phases, not in backends (backends should be as simple as possible)

Actually, I would like to have multiple phases for addressing this:
 1. one fully monadic
 2. one that just rewrite the control flow, basically that tries to remove early `return`s by pushing them at the end of the function (this phase might fail, sometimes)
 3. one that choose between those two

I'm currently thinking about (1), but if you're interested in writing phases, maybe you could help me with (2) :)
If you're interested we can chat about that tomorrow at the meeting

engine/backends/coq/coq_backend.ml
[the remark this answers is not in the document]
W95Psp: Oh, sorry, indeed!
But then you can just use `GlobalIdent.(name = record_name)`

(otherwise, two equal `global_ident`s that comes from different place in the AST would not be equal `==`-wise)

engine/backends/coq/coq_backend.ml
cmester0: I just wanted to start writing the translation into monadic binds, as it currently only works for result and option types, however V2 allows more general early returns, which I could handle directly as a state monad in Coq.

W95Psp: Oh right! Then you might want to disable `early_exit` but enable `monadic_binding`.
That allows for expr `Let` to be monadic.
(But fow now, it's draft and no phase actually spit out `monadic_binding` nodes)

engine/backends/coq/coq_backend.ml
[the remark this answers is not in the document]
cmester0: Yes, that is probably better, even though that should never be the case, as these are type declarations?

engine/backends/coq/coq_backend.ml
W95Psp: oh, you mean that you want to translate directly early returns into monadic stuff in Coq directly in the backend?
I think this is a nontrivial thing that should be a phase (and I hope the phase I'm writing corresponds to what you describe)

engine/backends/coq/coq_backend.ml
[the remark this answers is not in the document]
cmester0: I mean for the specific rust code, we should not have 2 type definitions with the same global identifier, as that would mean you define the same type twice?

engine/backends/coq/coq_backend.ml
W95Psp: Nice! That's what I plan to output! (with explicit binds though)

engine/backends/coq/coq_backend.ml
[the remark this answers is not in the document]
cmester0: I seems you cannot do `GlobalIdent.(name = record_name)` but have to do `GlobalIdent.equal name record_name`, as the derive macro produces an `equal` function and not `(=)`?

engine/backends/coq/coq_backend.ml
W95Psp: The complicated part is to make breaks, continues, returns, local mutation and functionalization of loops all play nicely together I think

cmester0: I have done that already for ssprove, so that is why I could try and play with it directly ..

engine/backends/coq/coq_backend.ml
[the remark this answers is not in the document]
W95Psp: Ok, that's a bit sad the PPX doesn't derive `=`

engine/backends/coq/coq_backend.ml
cmester0: The interesting part of the implementation can be found here:
https://github.com/cmester0/hacspec/blob/7c6d21cd45ec2625d8dfe32505388cde6c5d496c/coq_ssprove/src/Hacspec_Lib.v#L1458-L1553
https://github.com/cmester0/hacspec/blob/7c6d21cd45ec2625d8dfe32505388cde6c5d496c/coq_ssprove/src/Hacspec_Lib.v#L1718-L1726

PR #83, opened 2023-05-02
Start linter for hacspec

franziskuskiefer:
This PR adds
- test infrastructure for the linter
- infrastructure for the hacspec linter
- CI for the linter


franziskuskiefer:
@W95Psp something in nix is failing because it's trying to run a test that won't work the way it's trying to run it. Please fix it when doing the review.

frontend/diagnostics/src/error.rs
```diff
@@ -74,6 +74,18 @@ pub fn no_mut_self(session: &Lrc<Session>, span: Span) {
...
+    session.span_warn_with_code(
+        span,
+        "[Circus] Mutable arguments are not supported",
+        DiagnosticId::Lint {
+            name: "Mutability".to_string(),
+            has_future_breakage: false,
+            is_force_warn: true,
+        },
+    );
+}
```
W95Psp, line 87: At some point, we need to replace those functions and calls to those functions to something using the `Diagnostics` datatype

frontend/lint/tests/hacspec.rs
```diff
@@ -0,0 +1,54 @@
...
+            "--bin",
+            "cargo-circus",
+            "--",
+            "-C",
+            "--manifest-path",
+            test.manifest_path,
+            ";",
+            "lint",
+            "hacspec",
+        ]);
```
W95Psp, line 44: `cargo run` calls `cargo-circus` with env vars `DYLD_FALLBACK_LIBRARY_PATH` and `LD_LIBRARY_PATH` set to specific values, which messes up things in a rustup-free environment

frontend/lint/tests/util.rs
```diff
@@ -0,0 +1,39 @@
...
+    let mut cmd = Command::new("cargo");
+    cmd.current_dir("../../");
+
+    let status = cmd
+        .args(&["install", "--path", "cli/driver"])
+        .output()
+        .unwrap()
+        .status;
+    assert!(status.success());
+}
```
W95Psp, line 22: I think we should not install a binary while running test, it's a side effect I would not expect when running tests locally

frontend/lint/hacspec_tests/v1-lib/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "v1-lib"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+hacspec-lib = { git = "https://github.com/hacspec/hacspec.git" }
```
W95Psp, line 9: I know that will be a very sad and concerning news to you Franziskus, but... this is what breaks tests in Nix! Since there is no `Cargo.lock` committed in the Hacspec v1 repo, the dependencies are not known and thus no reproducible build (Nix or not) can be provided :(

Quoting [Cargo doc](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries):
> Users dependent on the library will not inspect the library’s Cargo.lock (even if it exists)

Thus, for non-Nix users, there is no harm in committing a `Cargo.lock`, which will be ignored anyway. So I think we should add `Cargo.lock` in hacspec v1, or move the library to some other repo with `Cargo.lock` committed.

For now, I fixed it on a `with-lockfile` branch.

W95Psp (commented):
I left some comments, and I fixed tests in Nix, this dynamic paths + rustup issue thing and a few other things in [branch `lucas/rust-linter-1`](https://github.com/hacspec/hacspec-v2/commits/lucas/rust-linter-1).
Can you look at it / rebase stuff that makes sense to you?

frontend/diagnostics/src/error.rs
franziskuskiefer: Yes!

frontend/lint/hacspec_tests/v1-lib/Cargo.toml
franziskuskiefer: Commiting lock files is not good practice for libraries. So it's nothing that we should do.
I'm afraid nix will need to find a different way to make this work. If it can't handle this, it won't be able to handle most Rust repositories.

frontend/lint/tests/hacspec.rs
franziskuskiefer: So `cargo run` doesn't work without rustup here? But I think it's generally fine to require `rustup` for this repo.

frontend/lint/tests/util.rs
franziskuskiefer: I agree, but this required rewriting the binaries. `cargo-circus` currently calls the driver binary, which requires that it's installed. This is generally bad, also for local development because you end up getting an old installed version of part of the repo when using `cargo run`.
If you want to fix this before getting this in, we can do that. But otherwise I'd prefer getting this merged and then getting it fixed in a follow up.

frontend/lint/tests/hacspec.rs
W95Psp: Yes, without rustup something is going wrong.
Well, I fixed it in 42c2744c8fb806ebc10baaeeb2dc8d75db5947c1 by adding an environment variable the circus driver is sensible to, so that we can inject the absolute path to the local cargo binary

If that change is fine to you, feel free to cherry-pick it in this branch :) (I can also do it if you want)

frontend/lint/tests/util.rs
W95Psp: Yes, so I fixed it in 42c2744c8fb806ebc10baaeeb2dc8d75db5947c1 by adding an environment variable the circus driver is sensible to, so that we can inject the absolute path to the local cargo binary

If that change is fine to you, feel free to cherry-pick it in this branch :) (I can also do it if you want)

frontend/lint/tests/hacspec.rs
W95Psp: Yes, without rustup something goes wrong.

42c2744c8fb806ebc10baaeeb2dc8d75db5947c1 fixes this issue as well, by explicitly remove such env vars Cargo adds

frontend/lint/hacspec_tests/v1-lib/Cargo.toml
W95Psp: Oh, my bad. Actually, that was a bug some while ago in `crane`, the Nix helper for building Cargo stuff, but it's fixed now.
So, forget about this, Nix (well, crane) handles that alright.

frontend/lint/tests/hacspec.rs
```diff
@@ -0,0 +1,46 @@
...
+        cmd.current_dir("../");
+
+        let output = cmd.output().unwrap();
+        eprintln!("{:?}", output);
+        let output = cmd.output().unwrap();
+        eprintln!("{:?}", output);
```
W95Psp, line 41: Why twice the same code?
```suggestion
        let output = cmd.output().unwrap();
        eprintln!("{:?}", output);
```

frontend/lint/tests/rust.rs
```diff
@@ -0,0 +1,57 @@
...
+#[test]
+fn lint_rust() {
+    for test in TESTS {
+        let mut cmd = Command::circus(&["-C", "--manifest-path", test.manifest_path]);
+        cmd.current_dir("../");
+
+        let output = cmd.output().unwrap();
+        eprintln!("{:?}", output);
+        let output = cmd.output().unwrap();
+        eprintln!("{:?}", output);
```
W95Psp, line 52: same here

W95Psp (approved):
LGTM, there's just one strange thing with duplicated code, but I'm fine saying we don't care

PR #89, opened 2023-05-11
Add custom test harness

W95Psp:
*(paraphrasing `test-harness/README.md`)*

> This ~crate~ PR defines a custom test harness[^1][^2] that scans for packages in the Cargo workspace `../tests/Cargo.toml`.
> 
> Each package in that workspace should define a sequence of tests to be run in the `package.metadata.circus-tests` dictionary of its `Cargo.toml` manifest.
> 
> ## Format for `package.metadata.circus-tests`
> `package.metadata.circus-tests` is a map from a target (e.g. `into fstar` or `lint hacspec`) to a **test specification** (see below).
> 
> `package.metadata.circus-tests` is expected to be a **dictionary** with the following optional fields:
>  - `lint`, a map from a **linter name** to a **test specification**.
>  - `into`, a map from a **backend name** to a **test specification**.
>    
> Note that instead of linter or backend names, conjunction are allowed, for instance `fstar+coq`.
>    
> ### Test specifications
> A **test specification** is a dictionary with the following fields:
>  - `positive: bool`: is the test positive (i.e. expected to succeed) or negative?
>  - `optional: bool`: is the test optional? (useful for slow tests for instance)
>  - `broken: bool`: is this test broken because of some feature not being implemented?
>  - `issue_id: u64`: when the test has a companion issue on GitHub (closed or not)
>  
> ### Linter names
> The available linters can be listed by running `cargo circus lint --help`.
> 
> ### Backend names
> The available backends can be listed by running `cargo circus into --help`.
> 
> [^1]: https://doc.rust-lang.org/cargo/reference/cargo-targets.html#the-harness-field
> [^2]: https://nexte.st/book/custom-test-harnesses.html

The test is disabled by default for now, since it requires building both the Rust part and the OCaml engine. Though, it can be run using `cargo test --test full`.

Also, this uses the library insta, that provides the subcommand `cargo insta`, which is a nice way of reviewing changes.


test-harness/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "test-harness"
+version = "0.1.0"
+edition = "2021"
+
+
+[[test]]
+name = "full"
```
franziskuskiefer, line 8: I'm not sure if full is the best name... But I also don't have a good idea what to use ... maybe "all"?

franziskuskiefer (changes requested):
Thanks, that's pretty nice!

We should use this in the litner test as well.

Can you add a CI job to run this test so we make sure it's working?

test-harness/Cargo.toml
W95Psp: Yeah, I was very happy either with `full`...
`all` is a bit better. Just got an idea: what about `toolchain` (so that we type `cargo test --test toolchain`)?

test-harness/README.md
```diff
@@ -21,16 +25,40 @@ for instance `fstar+coq`.
    
 ### Test specifications
 A **test specification** is a dictionary with the following fields:
- - `positive: bool`: is the test positive (i.e. expected to succeed) or negative?
- - `optional: bool`: is the test optional? (useful for slow tests for instance)
- - `broken: bool`: is this test broken because of some feature not being implemented?
- - `issue_id: u64`: when the test has a companion issue on GitHub (closed or not)
- 
+
+<!-- | Field | Type | Def. | Description | -->
```
franziskuskiefer, line 29: Remove?

tests/lint/hacspec/mut_args/Cargo.toml
```diff
@@ -6,3 +6,5 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+
+[package.metadata.circus-tests.lint.hacspec]
```
franziskuskiefer, line 10:
```suggestion
[package.metadata.circus-tests.lint.hacspec]

```

franziskuskiefer (approved):
Thanks, only a few nits

W95Psp:
bors +r

W95Psp:
bors cancel

PR #92, opened 2023-05-12
Add phases for translating side-effects (local mutation, early returns & possibly more)

### LINKED ISSUE #16: Improve OCaml compile time
Currently, the compile times are pretty bad, and there's almost no incrementally while compiling.
There are a lot of ML modules lacking a signature, so that is probably the reason.

W95Psp:
This PR adds the following phases:
 - `HoistSideEffects`
 - `Local_mutation`
 - `Cf_into_monads`.

This PR is about translating chunks of code that contain side effects.
The side effects considered are:
 - reading a mutable binding;
 - mutating a mutable binding;
 - mutating a `&mut`;
 - control flow expressions like `return`, `continue` or `break`.

The phase `HoistSideEffects` transforms expression into hoist side effects as let bindings.
The invariant that should hold on any expression after that phase is (more or less) that the expression is either:
 1. a sequence of non-monadic let bindings whose RHS might contain *direct* side-effects. An expression is a "direct side effect" when it is directly an (unwrapped) `return`, `continue`, `assign` etc. node.
 2. a if or a match whose structenees are pure.
 3. a loop (when it's a for loop, the range expression should be pure).
 4. a closure.

After this hoisting phase, dealing with side effect is quite easy.

Phase `LocalMutation` transform mutation of simple (not borrowed) `mut` bindings with shadowing.

Phase  `Cf_into_monads` transforms some side-effects leaves (such as `return`) as monadic expressions, and then elaborates monadic binds and lifts when needed.

This PR also adds signatures for OCaml modules implementing phases (that fixes #16).


cli/driver/src/exporter.rs
```diff
@@ -214,7 +214,9 @@ impl Callbacks for ExtractionCallbacks {
 
                     let out = engine_subprocess.wait_with_output().unwrap();
                     if !out.status.success() {
-                        panic!("{} exited with non-zero code", ENGINE_BINARY_NAME);
+                        eprintln!("{} exited with non-zero code", ENGINE_BINARY_NAME);
```
franziskuskiefer, line 217: We should probably move to logging for these things at some point.

cmester0 (approved):
Everything seems fine for the coq backend!

Issue #28, opened 2023-04-12
Port specifications from repository `hacspec/specs`


W95Psp:
- [ ] `tls_cryptolib` (fstar: ❌CE0001[^1], coq: ❌CE0001[^1])
 - [x] `hacspec-curve25519` (fstar: ✅, coq: ✅)
 - [x] `hacspec-chacha20` (fstar: ✅, coq: ✅)
 - [x] `hacspec-ristretto` (fstar: ✅, coq: ✅)
 - [x] `hacspec-linalg` (fstar: ✅, coq: ✅)
 - [ ] `hacspec-merlin` (fstar: ❌CE0002[^2], coq: ❌CE0002[^2])
 - [x] `hacspec-poly1305` (fstar: ✅, coq: ✅)
 - [x] `hacspec-chacha20poly1305` (fstar: ✅, coq: ✅)
 - [x] `hacspec-gimli` (fstar: ✅, coq: ✅)
 - [x] `hacspec-sha1` (fstar: ✅, coq: ✅)
 - [x] `hacspec-sha256` (fstar: ✅, coq: ✅)
 - [x] `hacspec-sha3` (fstar: ✅, coq: ✅)
 - [x] `hacspec-ntru-prime` (fstar: ✅, coq: ✅)
 - [x] `hacspec-riot-bootloader` (fstar: ✅, coq: ✅)
 - [x] `hacspec-riot-runqueue` (fstar: ✅, coq: ✅)
 - [x] `hacspec-hmac` (fstar: ✅, coq: ✅)
 - [x] `hacspec-hkdf` (fstar: ✅, coq: ✅)
 - [x] `hacspec-p256` (fstar: ✅, coq: ✅)
 - [x] `hacspec-bls12-381` (fstar: ✅, coq: ✅)
 - [x] `hacspec-ecdsa-p256-sha256` (fstar: ✅, coq: ✅)
 - [x] `hacspec-aes` (fstar: ✅, coq: ✅)
 - [x] `hacspec-aes-jazz` (fstar: ✅, coq: ✅)
 - [x] `hacspec-gf128` (fstar: ✅, coq: ✅)
 - [x] `hacspec-aes128-gcm` (fstar: ✅, coq: ✅)
 - [x] `hacspec-bls12-381-hash` (fstar: ✅, coq: ✅)
 - [x] `hacspec-ed25519` (fstar: ✅, coq: ✅)
 - [x] `hacspec-edwards25519` (fstar: ✅, coq: ✅)
[...]

PR #98, opened 2023-05-25
Backends: pull some macro processing out, add errors

W95Psp:
This PR:
 - move the processing of `secret_bytes` and `secret_arrays` from backends to a proper phase;
 - gives a `mli` signature to backends;
 - flatten a bit backends;
 - adds better error handling facilities for backends.

Makes issue #28 much more green!

@cmester0: I did some changes to the Coq backend, basically I removed a layer of module (which was necessary at some point but now it's not any longer), and by default transformed the TODO nodes into errors, so that it's easier to track what is not actually extracting. Wdyt?


W95Psp:
I'm thinking a bit about those inline macros: I think we should actually let Rust expand them... Those are generating code we (almost) support.

Expansions contain calls to `vec!`, which produces code we don't support, but I think we should support `vec!` instead of supporting specific hacspeclib macros

franziskuskiefer:
> I'm thinking a bit about those inline macros: I think we should actually let Rust expand them... Those are generating code we (almost) support.
> 
> Expansions contain calls to `vec!`, which produces code we don't support, but I think we should support `vec!` instead of supporting specific hacspeclib macros

Yes! I generally agree. Not expanding them would be a stopgap measure until it's all supported. But if you think it is easy enough to support the code in the macros, we should do that.

engine/backends/coq/coq_backend.ml
cmester0, line 1: These changes seems nice. At some point we would probably want to report the errors, and still continue the translation. However this allows you to kinda choose either, so it is a good alternative for now, and reduces confusion.

cmester0 (approved):

W95Psp:
I've removed all code relative to hacspec-specific inline macros, in favor of their Rust expansion as we discussed

This PR now also adds support for more fundamental rust stuff:
 - (some) pointer expressions, which are actually coercions between internal representations (e.g. sized [u8; N] to unsized slice [u8] or function signatures);
 - repeat expressions, for instance `[123; 18]`;
 - indirectly, this makes `vec!` work properly;
 - translate ZstLiteral nicely, those are actually used (mostly?) for unapplied functions that are specialized with generic types (one could see that as a type-only partial application)

PR #100, opened 2023-05-30
Rename `circus` to `hax`

W95Psp:


karthikbhargavan (commented):
What are all the commands available now, could you put them in README.md?
Do we really need the "into" prefix?
`cargo hax fstar` may well be enough?

W95Psp:
I updated the README with an *Usage* section to document a bit the commands.

Concerning the `into` subcommand, I think that's useful and nicer for user experience: you have to choose between the actions `lint`, `into` or `json`, and each of those takes different kinds of options (i.e. there is `cargo hax into --output-dir` but no `cargo hax lint --output-dir`)

karthikbhargavan (approved):

PR #104, opened 2023-05-31
Add `--force` flag to `cargo install`

W95Psp:
This PR adds `--force` flag by default to `setup.sh` so that people can update Hax properly.
Otherwise, `cargo install` sees a binary with the same name, and skips the installation.


karthikbhargavan (approved):
At some later stage, we should use proper version numbers to address this.

PR #94, opened 2023-05-23
Temporary fix for `derive`d code (fixes #78)

W95Psp:
We want to skip derived implementations for `PartialEq`, `Eq`, `Clone`, `Debug`. This patch filters out every item that was automatically derived. This is a too coarse filter, we need to rework that.

Issue #45, opened 2023-04-14
Add option for marking items as non-export


W95Psp:
We should have both:
 - CLI options
 - Rust attributes

We should be able to express "everything but foo" and "nothing but foo" (at the module level)

franziskuskiefer:
@W95Psp how does this relate to #211 ?

W95Psp:
I think that's actually a duplicate of https://github.com/hacspec/hacspec-v2/issues/223, which I just added...

W95Psp:
Closed #223 in favor of that one

W95Psp:
Closing this, the feature is there in main.

PR #106, opened 2023-05-31
Add support for attributes & expn_data on items (fixes #49, fixes #88)

### LINKED ISSUE #49: Translate attributes

### LINKED ISSUE #88: Extract proc macros informations
For now, the THIR we export has no information (or more poorly shaped information) about which block of code was expanded by which proc-macro.
We have no support for attributes in general.

This is required for #78

W95Psp:
This also makes #45 trivial to implement

This is draft because I removed the quick & dirty filter implemented in #94, thus I now need to restore the functionality.
 - [ ] when we see an `impl` for a `Clone` derived by `core::clone::Clone`, replace it with an identity impl
 - [ ] when we see an `impl` for a `Debug` derived by `core::fmt::macros::Debug`, replace it with a dummy impl (e.g. a constant string)


frontend/exporter/src/all.rs
```diff
@@ -677,13 +694,18 @@ pub struct ExpnData {
...
+#[derive(Clone, Debug, Serialize, Deserialize, JsonSchema)]
 pub struct Span {
     lo: Loc,
     hi: Loc,
     filename: FileName,
     // expn_backtrace: Vec<ExpnData>,
-    // macro_call: Option<MacCall>,
+}
+
+impl PartialEq for Span {
```
franziskuskiefer, line 705: This look equivalent to the derived implementation. Any reason you don't derive this?

franziskuskiefer (approved):
@karthikbhargavan do you want to review?
We should get this in.

frontend/exporter/src/all.rs
W95Psp: Oops, indeed, I wrote this imp for debugging purposes at some point, we should drop it (I don't have my laptop with me now though)

karthik-cryspen (approved):
This looks very useful. An example/test would have been handy.

PR #107, opened 2023-05-31
Small fixes

W95Psp:
Small fixes after #98, notably, update the list of examples that are still failing, so that the CI and #28 reflect the same thing.
Now, only two v1 examples are not extracting: `tls_cryptolib` and `hacspec-merlin`.


karthikbhargavan (approved):
Looks good. This is needed for some examples.

PR #110, opened 2023-06-01
Coq backend restructure

cmester0:
Restructure to allow multiple backends for Coq, and get better separation.


W95Psp:
Looks nice, though I'm wondering: why do you need the secret arrays & secret bytes special handling?
It should be fine now, those should actually get expanded by Rust, right?
(that's what currently happens in `main`, after PR #98, that finally get rids of macros we can just let Rust expand)

cmester0:
Yes, you are right. I was just running the command with incorrect arguments to `-i`.

franziskuskiefer (approved):
:shipit:

W95Psp (approved):
LGTM, one thing though: could you restore a mli interface to `coq_backend.ml`?

PR #111, opened 2023-06-01
Ssprove backend

cmester0:
Add ssprove backend skeleton, still missing mutable variable analysis.


W95Psp:
That's great, thanks! We should chat about which interface should mutable variable analysis have, and how we could factor this so that it's usable by any other backend.
Maybe we should set up a chat or something to discuss that live, wdyt?

cmester0:
Yes, this should definitely be reusable by any backend. The analysis I need is what functions are used in any given scope, and what mutable variables are used in any given scope. I did this as a simple two phase parse over the AST, saving the data as an extra field on `expr`. I think something similar could work for hax.

cmester0:
Hmm, yes that seems correct, however I am not sure how you would map from `expr` as there is nothing unique in any given expression, and the context matters for what the values of any given expression is. 

I would expect that in general we would want some way of getting information relevant to any expression, for a static analysis done. Usually I would expect this information to be decorated onto `expr`, however this can also be done by adding an `id`, and then save the information in a map. I am not sure what is the cleanest solution.

W95Psp:
> Hmm, yes that seems correct, however I am not sure how you would map from `expr` as there is nothing unique in any given expression, and the context matters for what the values of any given expression is.

What exactly is that context?

> I would expect that in general we would want some way of getting information relevant to any expression, for a static analysis done. Usually I would expect this information to be decorated onto `expr`, however this can also be done by adding an `id`, and then save the information in a map. I am not sure what is the cleanest solution.

What I dislike in attaching data to nodes directly (in a type-safe manner) is that:
 - this doesn't scale well (e.g. if we want to add other such additional information, that means we need more fields)
 - phases might break the coherency of those information

cmester0:
For any given expression it is the locations and functions used by the sub-expression / variables. So for every expression you need to accumulate anything that the expression might have been affected by.

Yes, I agree that it is much cleaner to have maps with the information, this is also how the rust compiler does it.

W95Psp:
Ok!
But then we can memoize by hashing expressions, that should be fine. 
Hashing will be a bit slow, but I think that's really alright, the bottleneck will always be the Rust JSON thing, and I prefer the tradeoff of performance for cleanness for now

W95Psp:
Also: is this analysis exactly useful for encoding mutation in the state monad, or for something else?

cmester0:
It is useful for assigning locations/memory to functions/variables. This is what allows us to reason about how functions interact, if they share or do not share some memory locations (e.g. if they call the same helper function).

cmester0:
I have changed the `local mutation` phase such that it does not remove mutable variables (this should probably be seperated into two phases anyways).

W95Psp:
Do you need to transform "standard" for loops into state-passing ones so that you can apply the loop functionalization phase afterward?
I don't get why we should have a phase that transform standard for loops into state passing ones if we don't disable mutation in the same time

cmester0:
Hmm, seems I do not want to functionalize the for loops, so you might be correct. However I still want the for loops to have an accumulator / state, as part of the loop is still pure/functional.

W95Psp:
Consider the following loop:
```rust
for pat in it {
  BODY
}
```
where `BODY` mutates the *local* variables `v₁`...`vₙ`.

You want to transform it as:
```rust
(for⟨state=(mut v₁, …, mut vₙ)⟩ pat in it {
   BODY // untouched BODY, doing some local mutation
   (v₁, …, vₙ)
})⟨(v₁, …, vₙ)⟩
```

Is that what you want?

PR #129, opened 2023-06-07
AST: add exprs in types, so that we have const generics

W95Psp:


franziskuskiefer (approved):
Thanks! lgtm. I didn't look too deep into the engine code.
❤️ for more docs.

franziskuskiefer (approved):
Thanks! lgtm. I didn't look too deep into the engine code.
❤️ for more docs.

W95Psp:
Yes, doc is lacking I agree! ^^
Thanks for the review!

PR #132, opened 2023-06-08
Add array update

franziskuskiefer:
This is not complete but is a step in the right direction.


karthikbhargavan:
Field update is coming next.

W95Psp (approved):
I think this PR is ready.
What do you think @karthikbhargavan, @franziskuskiefer, should we `bors r+`?

karthikbhargavan (approved):
I think this is a useful feature to have in main.

PR #135, opened 2023-06-09
Generalized `for` loops

W95Psp:
This PR:
 - [x] makes `ForLoop` by default be a for loop over iterators (i.e. `for i in it {body}`)
 - [x] adds `ForIndexLoop`
 - [x] adds missing type information in some phases
 - [ ] resugars the `ForLoop`s that deal with ranges into `ForIndexLoop`
 - [x] removes rejection phases from backends

Note that the `for i in it {body}` loops are now translated using the standard rust iterator `fold` method


franziskuskiefer:
Nice! Can you add a couple tests?

W95Psp:
Done, added some tests.

Issue #90, opened 2023-05-11
Error `UnallowedMutRef` when it should not


W95Psp:
Reported by @linesthatinterlace on [this zulip message](https://hacspec.zulipchat.com/#narrow/stream/269544-general/topic/Porting.20a.20Rust.20program.20to.20HacSpec/near/357604302).

Minimal example:
```rust
fn f() {
    let mut vec = Vec::new();
    vec.push(1);
    vec.push(2);
    vec.swap(0, 1);
}
```

Here, `vec.swap(0, 1)` (really it is `(&mut vec).swap(0, 1)`) should be translated as a functional call plus a local mutation, i.e. `vec = vec.swap(0, 1)`.

W95Psp:
I investigated a bit into this: the issue is that here we deal with smart pointers with implicit coercions.

`vec.swap(0, 1)` is actually transformed by  the rust compiler into something like:
```rust
 core::slice::swap(
   &mut (*(core::ops::deref::DerefMut::deref_mut(&mut vec))),
   0, 1,
 )
```

This happens because `swap` is a method that comes from `Deref<Vec<_>>`.

@franziskuskiefer, I cannot reproduce `vec.append(something)`. For instance, the following snippet works:
```rust
fn _f() {
    let mut vec = Vec::new();
    vec.push(1);
    vec.push(2);
    let mut test = vec![12];
    vec.append(&mut test);
    // vec.swap(0, 1);
}
```

PR #136, opened 2023-06-10
vec append test

franziskuskiefer:
Test for #90

### DIFF
--- a/tests/mut-ref-functionalization/src/lib.rs
+++ b/tests/mut-ref-functionalization/src/lib.rs
@@ -4,3 +4,16 @@ fn _f() {
     vec.push(2);
     vec.swap(0, 1);
 }
+
+fn build_vec() -> Vec<u8> {
+    vec![1, 2, 3]
+}
+
+fn _append() {
+    let mut vec1 = Vec::new();
+    let mut vec2 = vec![1u8, 2, 3];
+    // This works
+    vec1.append(&mut vec2);
+    // This doesn't work
+    vec1.append(&mut build_vec());
+}


W95Psp:
I think this example is actually a *legitimate* place where Hax throws a "UnallowedMutRef": the second argument of `vec.append(&mut )` is actually a `&mut something` with `something` being `&mut build_vec()`, an arbitrary expression.

For now, the rule is really that `something` should exactly be a local mutable identifier.
Though, we could also allow any expression that binds no mutable bindings. (that's fine until someone uses interior mutability, but that's illegal to do so --there's no pure model of interior mutability)

Should I open a new issue to support a bigger variety of `&mut` arguments?

franziskuskiefer:
> Should I open a new issue to support a bigger variety of `&mut` arguments?

Yes. Not allowing this forces people to write ugly (and probably less efficient) Rust code.

karthikbhargavan:
I think this is part of a longer and longer-term discussion on the language. For now, all our treatment of `&mut` is for super-simple cases. The `vec` library is stretching the limits. Our choice are:
1) We forbid all `&mut` outside a small range, even if it means telling people clearly that certain functions on `Vec<T>` are not supported
2) We do something specific for `Vec<T>` because of its ubiquity and forbid these `&mut` use-cases elsewhere
3) We extend the language to support a larger set of mutation patterns.

I can see the attraction of (3) but it will very soon expand to the full Rust language, which will require us to learn the lessons that many other Rust tools are grappling with, and to have a more principled treatment of the Rust semantics and our phase transformations. A worthwhile endeavour but not one to be done in a rush.

W95Psp:
Yeah, I agree with you @karthikbhargavan.
We should only support a clear and limited set of `&mut` use cases which are:
 - simple enough to identify for the users;
 - straightforward to implement;
 - straightforward to justify in terms of semantics.

Maybe the best move forward here is actually documentation: I think I'm going to take some time to write some markdown to explain what is the scope of our `&mut` transformations, what is already done, what is left to do, and what we don't want to address at all.

karthikbhargavan:
> Maybe the best move forward here is actually documentation: I think I'm going to take some time to write some markdown to explain what is the scope of our `&mut` transformations, what is already done, what is left to do, and what we don't want to address at all.

Good idea, we should be documenting all the phase transformations, but this is a good one to start with. 
But instead of saying "we don't want to address at all", I would rather say "we don't address in the current version" and invite people to give us examples of where this is limiting them, so that we can use them as motivation for our next version.

linesthatinterlace:
For what it's worth, good support for Vec feels like it would be a big win without having to do as much of this grander scope thinking.

PR #138, opened 2023-06-13
Remove specific code for `[x; LEN]`, use general const to expr

W95Psp:
Fixes the construction of arrays `[x; LEN]` when `LEN` is not a literal.


franziskuskiefer (approved):
:shipit: 

bors r+

Issue #27, opened 2023-04-12
Fix `attempted to read from stolen value`


W95Psp:
The frontend browses items of a crate in an order that causes this bug.
Some data in the Rust compiler is kept in a `Steal`, which is basically a box whose content can be stolen, that is pulled out of the memory (rustc uses [arenas](https://doc.rust-lang.org/stable/nightly-rustc/rustc_arena/index.html#:~:text=Arenas%20are%20a%20type%20of,arena%20itself%20is%20still%20alive.))

Thus, the order in which we browse the crate to export every item matters: sometimes I inspect things that happens to be stolen before because of some previous inspections.
This is particularly true for constants, it seems.

**Status: mostly fixed**
 - @geonnave's bug (https://github.com/hacspec/hacspec-v2/issues/27#issuecomment-1568365024) is fixed
 - @franziskuskiefer's bug (https://github.com/hacspec/hacspec-v2/issues/27#issuecomment-1586990315) is fixed

*However*, I tried Hax on `core`, and I get new stolen stuff. Thus, there exists some scenarios in which this still fails.

geonnave:
For the record (context from #99):
- this issue also affects translation of the [`edhoc-rs`](https://github.com/openwsn-berkeley/edhoc-rs) crate into fstar.
- in #99 there is a `How to reproduce` section, that could be validated against once this issue is addressed.

franziskuskiefer:
Another code snippet that panics

```rust
#[repr(u16)]
pub enum ProtocolVersion {
    Mls10 = 1,
}

impl ProtocolVersion {
    #[allow(non_upper_case_globals)]
    fn tls_deserialize(bytes: &[u8]) -> core::result::Result<(Self, &[u8]), tls_codec::Error> {
        const __TLS_CODEC_Mls10: u16 = ProtocolVersion::Mls10 as u16;
        let (discriminant, remainder) =
            <u16 as tls_codec::DeserializeBytes>::tls_deserialize(bytes)?;
        match discriminant {
            __TLS_CODEC_Mls10 => {
                let result = ProtocolVersion::Mls10 {};
                Ok((result, remainder))
            }
            _ => Err(tls_codec::Error::UnknownValue(discriminant.into())),
        }
    }
}
```

W95Psp:
Another stealing bug (from issue #474):

https://github.com/hacspec/hax/blob/aa14fb139280e7fcd28acb45db27b01b53c9de93/tests/rust-ast/src/lib.rs#L14-L20

[...]

PR #139, opened 2023-06-13
Fix steal bug

W95Psp:
This is an attempt at fixing the frontend exporter bug described in #27.

The fix is really simple: it's all about processing `const`-like items first.

@franziskuskiefer: I tested a bit this fix on a reduced version of the example you posted in #27, but I think you've been hitting that much more, can you try this patch?


franziskuskiefer (approved):
I ran it on a couple larger code bases where it was crashing before and it appears to go through everywhere now 🥳

PR #143, opened 2023-06-13
Analysis intro

cmester0:
Here is an initial attempt at structuring the analyses. Structured based on suggestions in  #111

### DIFF
--- a/engine/lib/analyses/flatten_ast.ml
+++ b/engine/lib/analyses/flatten_ast.ml
@@ -1,0 +1,49 @@
+open Base
+open Utils
+
+module%inlined_contents Make (F : Features.T) = struct
+  module FA = F
+  module A = Ast.Make (F)
+  open Ast
+
+  let rec flatten_ast (x : A.expr) : A.expr list =
+    x
+    ::
+    (match x.e with
+    | If { cond; then_; else_ } -> (
+        flatten_ast cond @ flatten_ast then_
+        @ match else_ with None -> [] | Some x -> flatten_ast x)
+    | App { f; args } -> flatten_ast f @ List.concat_map ~f:flatten_ast args
+    | Literal l -> []
+    | Array elements -> List.concat_map ~f:flatten_ast elements
+    | Construct { constructor; constructs_record; fields; base } ->
+        List.concat_map ~f:(snd >> flatten_ast) fields
+    | Match { scrutinee; arms } ->
+        flatten_ast scrutinee
+        @ List.concat_map
+            ~f:(fun { arm = { arm_pat; body } } -> flatten_ast body)
+            arms
+    | Let { monadic; lhs; rhs; body } -> flatten_ast rhs @ flatten_ast body
+    | LocalVar l -> []
+    | GlobalVar g -> []
+    | Ascription { e; typ } -> flatten_ast e
+    (* Macro *)
+    | MacroInvokation { macro; args; witness } -> []
+    (* Mut *)
+    | Assign { lhs; e; witness } -> flatten_ast e
+    (* Loop *)
+    | Loop { body; kind; state; label; witness } -> flatten_ast body
+    (* ControlFlow *)
+    | Break { e; label; witness } -> flatten_ast e
+    | Return { e; witness } -> flatten_ast e
+    | QuestionMark { e; converted_typ; witness } -> flatten_ast e
+    | Continue { e; label; witness } -> (
+        match e with Some (_, v) -> flatten_ast v | None -> [])
+    (* Mem *)
+    | Borrow { kind; e; witness } -> flatten_ast e
+    (* Raw borrow *)
+    | AddressOf { mut; e; witness } -> flatten_ast e
+    | Closure { params; body; captures } ->
+        flatten_ast body @ List.concat_map ~f:flatten_ast captures
+    | EffectAction { action; argument } -> flatten_ast argument)
+end
--- a/engine/lib/analyses/flatten_ast.ml
+++ b/engine/lib/analyses/flatten_ast.ml
@@ -16,7 +16,7 @@ module%inlined_contents Make (F : Features.T) = struct
     | App { f; args } -> flatten_ast f @ List.concat_map ~f:flatten_ast args
     | Literal l -> []
     | Array elements -> List.concat_map ~f:flatten_ast elements
-    | Construct { constructor; constructs_record; fields; base } ->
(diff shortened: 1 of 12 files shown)


cmester0:
I ran into issues with the analysis wanting to run in some specific phases / set of features, while we want to use the analysis in the final phase, however this might just be an issue with how the features are set for `SSProve` currently, as we would like mutable variables.

cmester0:
Another point is currently all backend are curried for the translation one item at a time, however when doing static analysis, we will need the entire AST.

engine/lib/analyses/flatten_ast.ml
```diff
@@ -0,0 +1,49 @@
+open Base
+open Utils
+
+module%inlined_contents Make (F : Features.T) = struct
+  module FA = F
+  module A = Ast.Make (F)
+  open Ast
+
+  let rec flatten_ast (x : A.expr) : A.expr list =
```
W95Psp, line 9: Here, `flatten_ast` can be implemented directly using [visitors](http://gallium.inria.fr/~fpottier/visitors/manual.pdf), which are derived for the entire AST:
```ml
let rec flatten_ast : A.expr -> A.expr list =
    (object
       inherit [_] A.expr_reduce as super
       inherit [_] list_monoid as m
       method visit_t _ _ = m#zero
       method visit_mutability (_f : unit -> _ -> _) () _ = m#zero
       method visit_expr f e = e :: super#visit_expr f e
    end)#visit_expr ()
```

given some `list_monoid` like:
```ml
class ['s] list_monoid =
    object
      inherit ['s] VisitorsRuntime.monoid
      method private zero = []
      method private plus = List.append
    end
```

The nice thing with this is that the code will be robust against AST changes.

engine/lib/analyses/mutable_variables.ml
```diff
@@ -0,0 +1,91 @@
...
+  module FA = F
+  module A = Ast.Make (F)
+  open Ast
+
+  type pre_data =
+    ( concrete_ident,
+      concrete_ident list,
+      Concrete_ident.comparator_witness )
+    Map.t
+    * (concrete_ident, int, Concrete_ident.comparator_witness) Map.t
```
W95Psp, line 14: I think here a few things would help:
 1. Use `Map.M`, to get rid of the `comparator_witness` noise:
    ```suggestion
       type analysis_data =
           concrete_ident list Map.M(Concrete_ident).t * int Map.M(Concrete_ident).t
    ```
 2. Make a type synonym and use it instead of `int` (I guess it's some kind of id?)

engine/lib/analyses/mutable_variables.ml
```diff
@@ -0,0 +1,91 @@
...
+
+  type analysis_data =
+    ( concrete_ident,
+      local_ident list * ((local_ident * A.ty) * int) list,
+      Concrete_ident.comparator_witness )
+    Map.t
```
W95Psp, line 20: Same here.
Actually, maybe you could define a `'t analysis_data` of type `(local_ident list * 't) Map.M(Concrete_ident).t`?
In my opinion that would make the aim of the code more clear: your `analyse` function would just add information from a `unit analysis_data` to a `(TypedLocalIdent.t * int) list analysis_data`, wdyt?

(note `TypedLocalIdent` is defined in `Ast_utils`)

engine/lib/analyses/mutable_variables.ml
```diff
@@ -0,0 +1,91 @@
...
+           ~f:(fun x ->
+             match x.e with
+             | Assign { lhs = LhsLocalVar { var; typ }; witness } ->
+                 Some (var, typ)
+             | Let { lhs = { p = PBinding { mut = Mutable _; var; typ } }; _ }
+               ->
+                 Some (var, typ)
+             | _ -> None)
+           (Flatten.flatten_ast x))
+    in
```
W95Psp, line 80: This is exactly a computing a map from `A.expr` to a list of typed ident that were assigned, right?
`Ast_utils.free_assigned_variables` do something a bit different (it's taking care of scoping), but I think you could implement that with a visitor in a very simple way (**without flattening exprs**):
```ml
object
        inherit [_] expr_reduce as super
        inherit [_] Sets.TypedLocalIdent.monoid as m
        method visit_t _ _ = m#zero
        method visit_mutability (_f : unit -> _ -> _) () _ = m#zero
            
        method visit_Assign _env lhs e _wit =
          let rec visit_lhs lhs =
            match lhs with
            | LhsLocalVar { var; _ } ->
                Set.singleton (module TypedLocalIdent) (var, e.typ)
            | LhsFieldAccessor { e; _ } -> visit_lhs e
            | LhsArrayAccessor { e; index; _ } ->
                Set.union (super#visit_expr () index) (visit_lhs e)
            | LhsArbitraryExpr { witness; e } -> m#zero
          in
          visit_lhs lhs
      end
```

(note that this codes handles all kinds of LHSs)

W95Psp (commented):
Starting some reviews

engine/lib/analyses/mutable_variables.ml
cmester0: It will not always be the case that the pre and post analysis data types can be described using the same dependent type. So I think it might be better to keep it general and have a type for both from and to.

engine/lib/analysis_utils.ml
```diff
@@ -0,0 +1,11 @@
...
+
+module type ANALYSIS = sig
+  (* module FA : Features.T *)
+  module A : Ast.T
+
+  type pre_data
+  type analysis_data
+
+  val analyse : pre_data -> A.item list -> analysis_data
+end
```
W95Psp, line 11: Do we actually need to force analyses to be modules with a specific type?
An analyse could also operate on a different level (i.e. expressions), so maybe it'd be nicer to let analyses be just functions without constraints on their type signatures, wdyt?
Backends or phases can then freely call them in a natural way

W95Psp (approved):
Seems fine to me, just one or two comments.
Those analyses and things in `ast_utils` should be the same thing I think. Though, we can think about that later.

engine/lib/analyses/mutable_variables.ml
W95Psp: I added a comment about the `ANALYSIS` type module below: I'm not sure we should constraint the type signature of such analysis functions. (then `ANALYSIS` is basically encoding a function of arity two where the second argument is more or less fixed, it's not bringing much value)

PR #148, opened 2023-06-15
Make `Steal` error recoverable, make more errors non-fatal

W95Psp:
I found some other code that triggers the steal error.
This PR makes the error recoverable using some unsafe things: we report an error but keep going.

### DIFF
--- a/cli/driver/src/exporter.rs
+++ b/cli/driver/src/exporter.rs
@@ -59,7 +59,6 @@ fn convert_thir<'tcx>(
         .map(|did| {
             let span = hir.span(hir.local_def_id_to_hir_id(did));
             let mk_error_thir = || {
-                // let ty = tcx.ty_error(error);
                 let ty = tcx.mk_ty_from_kind(rustc_type_ir::sty::TyKind::Never);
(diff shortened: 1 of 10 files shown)


cli/driver/src/exporter.rs
```diff
@@ -57,11 +57,48 @@ fn convert_thir<'tcx>(
...
     > = bodies
         .map(|did| {
-            let (thir, expr) = tcx
-                .thir_body(did)
-                .expect("While trying to reach a body's THIR defintion, got a typechecking error");
-            // Borrowing `Thir` from a `Steal`!
-            (did, (thir.borrow().clone(), expr))
+            let span = hir.span(hir.local_def_id_to_hir_id(did));
+            let mk_error_thir = || {
+                // let ty = tcx.ty_error(error);
```
franziskuskiefer, line 62: remove?

franziskuskiefer (approved):
Great 👍🏻  (I don't see unsafe things here 😉 )

W95Psp:
Thanks for the review :)
Yes, well, that's not exactly `unsafe`, it's the `std::panic::AssertUnwindSafe` bit.
Though, I think that's totally fine, it's a good trade-off (so that we can actually debug when the steal error occurs!)

PR #149, opened 2023-06-15
Inherent `impl`s (fix #115)

### LINKED ISSUE #115: support inherent `impl`s (`impl` without trait)
Consider the following [inherent trait](https://doc.rust-lang.org/reference/items/implementations.html):
```rust
struct Foo { x: u8 }
impl Foo {
  fn bar(&self) -> u8 {
    self.x
  }
}
```

We can translate that either as: (note: this is Rust notation, but that's just to illustrate the translation)
1. ```rust
   trait some_unique_generated_name { fn bar(&self) -> u8 }
   impl Foo { fn bar(&self) -> u8 { self.x } }
   ```
2. ```rust
   fn bar(&self) -> u8 { self.x }
   ```

(2) feels a bit nicer in terms of simplicity of extracted code, but that requires looking up for each function application whether we apply a method that comes from a impl without trait or not

With (1) we stick more to what exists in the original code, and the transformation is easy and local: for every "`impl`-without-trait" we see, we produce a `trait` and an `impl`.

W95Psp:
Status: ready!


W95Psp:
I made progress on this PR.
As we discussed on Wednesday @cmester0, I made quite a lot of changes in this PR, so that impacted the Coq backend.
I fixed some stuff, but not everything and I left some comments as TODOs in the code (see commit 5a762d932b153b4770b77d1339737f69ef39483f).
Could you look at those?

karthikbhargavan (approved):
Much needed changes.

PR #157, opened 2023-06-27
Structure output directory

pnmadelaine:
Remove the `--output-dir` option of `cargo hax into` and instead extract to `{crate}/hax/{backend}/extraction`


cli/driver/Cargo.toml
```diff
@@ -25,3 +25,4 @@ log.workspace = true
 pretty_env_logger.workspace = true
 inquire = "0.6"
 const_format = "0.2"
+cargo_metadata = "0.15"
```
W95Psp, line 28: Can you move that dependency to the root `Cargo.toml`, and then use `cargo_metadata.workspace = true` here and in `test-harness/Cargo.toml`?

cli/driver/src/exporter.rs
```diff
@@ -320,29 +320,34 @@ impl Callbacks for ExtractionCallbacks {
...
-                                    ))
-                                })
-                            }
-                        }
+                    let metadata = cargo_metadata::MetadataCommand::new().exec().unwrap();
+                    let crate_name = std::env::var("CARGO_CRATE_NAME").unwrap();
+                    let package = metadata
+                        .packages
+                        .iter()
+                        .rfind(|pkg| pkg.name == crate_name)
```
W95Psp, line 328: why not just find?

W95Psp (approved):
Thanks :) jsut a few things

W95Psp:
bors cancel

franziskuskiefer (approved):
The output structure needs to be described somewhere (Readme or book or something). Otherwise this looks god to me.

PR #165, opened 2023-06-29
Disable `halo2` and `weierstrass` in the tests of Hax

W95Psp:
Those are not supported yet, PR #149 and #160 will fix those.
But for now, that blocks all the PRs, so we should disable those.


franziskuskiefer (approved):
bors r+

Issue #96, opened 2023-05-24
Transformation of control flow into monads: support `continue`, `break` and `return`


W95Psp:

W95Psp:
This is related to #28: causes an issue with `tls_cryptolib`

lassebramer:
Ran into this error: 
```
error[CE0001]: (Diagnostics.Context.Phase CfIntoMonads): something is not implemented yet. This is discussed in issue 96: please upvote or comment this issue if you see this error message.
               TODO: Monad for loop-related control flow
   --> pasta/src/pasta.rs:783:5
    |
783 | /     for _ in 0..loop_upper_bound {
784 | |         if check_not_zero(r.clone()) && degree_polyx(r.clone()) >= degree_polyx(d.clone()) {
785 | |             let t: Polyx = divide_leading_terms(r.clone(), d.clone());
786 | |
...   |
790 | |         }
791 | |     }
    | |_____^
```

Full function here:

```
pub fn divide_poly(n: Polyx, d: Polyx) -> (Polyx, Polyx) {
    let mut q: Polyx = Seq::<FpVesta>::create(n.len());
    let mut r: Polyx = n.clone();

    let mut loop_upper_bound = d.len();
    if q.len() > d.len() {
        loop_upper_bound = q.len();
    }
    for _ in 0..loop_upper_bound {
        if check_not_zero(r.clone()) && degree_polyx(r.clone()) >= degree_polyx(d.clone()) {
            let t: Polyx = divide_leading_terms(r.clone(), d.clone());

            q = add_polyx(q, t.clone());
            let aux_prod: Polyx = mul_polyx(d.clone(), t.clone());
            r = sub_polyx(r, aux_prod);
        }
    }

    (trim_poly(q), trim_poly(r))
}
```

W95Psp:
Thanks for your bug report! This for loop contains no `break` or `continue`; thus I suspect hax is not reconstructing properly a for loop. I opened #105 for tracking that specifically.

W95Psp:
Closing in favor of #15

PR #170, opened 2023-06-30
Introduce feature `break`

W95Psp:
The `Break` expression constructor of the AST was gated only by the feature `loop`. This commit introduces the feature `break` and makes the `Break` constructor gated by the witness tuple `break * loop`.

Thanks to that, the phase `Cf_into_monads` now can remove `Break`-nodes in a type-safe manner. (though, for now, it removes `Break`s in favor of a runtime error, see issue #96)

~**PR #169 should be merged before this PR**~ :heavy_check_mark:


W95Psp:
There's not really something to test actually: this is an addition at the type level only, that should just prevent us shooting in the leg later on

franziskuskiefer (approved):
Untestable changes are never great ... but let's get this in then.
bors r+

PR #171, opened 2023-07-03
Switch from `cachix` to `DeterminateSystems/magic-nix-cache`

W95Psp:
Simplifies a bit the CI & should improve performances.
This replaces [Cachix](https://app.cachix.org/cache/hacspec) with https://github.com/DeterminateSystems/magic-nix-cache, that uses Github Actions cache.
That's less configuration and fetching cache should be faster.


franziskuskiefer (approved):
bors r+

PR #174, opened 2023-07-03
add `--deps` option

### LINKED ISSUE #166: CLI: add a `--deps` flag
Have a `--deps` flag (or another name) to the `cargo hax` subcommand. (this concerns the linter, the JSON and the backends)
By default, `cargo hax into BACKEND` would translate/lint/* only the package (i.e. a folder with `Cargo.toml` that is not a workspace) it's called on, with `--deps`, it would also translate/lint/* any other local dependent package.

Plan:
According to [the docs](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates), cargo gives us `CARGO_PRIMARY_PACKAGE`:

> CARGO_PRIMARY_PACKAGE — This environment variable will be set if the package being built is primary. Primary packages are the ones the user selected on the command-line, either with -p flags or the defaults based on the current directory and the default workspace members. This environment variable will not be set when building dependencies. This is only set when compiling the package (not when running binaries or tests).

this variable is exactly what we need: in `driver/driver.rs`, we shall translate/lint/* (i.e. run rustc with our callbacks) only when `CARGO_PRIMARY_PACKAGE` is non-empty or `--deps` is on.

### LINKED ISSUE #167: Build scripts (aka `build.rs`) are being translated/linted/exported-to-json
A custom build script is a particular target for Cargo: it's a `custom-build` target.
Before diving into this, let's see if this issue is not solved by #166 (or: does cargo consider a custom build being a dependency)

pnmadelaine:
fixes #166 
fixes #167

### DIFF
--- a/cli/options/src/lib.rs
+++ b/cli/options/src/lib.rs
@@ -11,18 +11,24 @@ pub struct ForceCargoBuild {
     pub data: u128,
 }
 
+impl std::default::Default for ForceCargoBuild {
+    fn default() -> Self {
+        ForceCargoBuild { data: 0 }
+    }
+}
+
 impl std::convert::From<&std::ffi::OsStr> for ForceCargoBuild {
     fn from(s: &std::ffi::OsStr) -> Self {
-        ForceCargoBuild {
-            data: if s == "false" {
-                use std::time::{SystemTime, UNIX_EPOCH};
-                SystemTime::now()
+        use std::time::{SystemTime, UNIX_EPOCH};
+        if s == "false" {
+            ForceCargoBuild {
+                data: SystemTime::now()
                     .duration_since(UNIX_EPOCH)
                     .map(|r| r.as_millis())
-                    .unwrap_or(0)
-            } else {
-                0
-            },
+                    .unwrap_or(0),
+            }
+        } else {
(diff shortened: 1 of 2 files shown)


W95Psp (approved):

cli/options/src/lib.rs
```diff
@@ -200,6 +206,10 @@ pub struct Options {
     /// [cargo] caching is disabled by default, this flag enables it back.
     #[arg(long="enable-cargo-cache", action=clap::builder::ArgAction::SetTrue)]
     pub force_cargo_build: ForceCargoBuild,
+
+    /// Apply the command to the dependencies of the primary package
```
W95Psp, line 210: The description is a bit terse, what about:
```suggestion
    /// By default, the action ([translate], [lint], or [json]) is applied only on the package selected by the user (i.e. the package or virtual workspace of the current directory, or the one selected with [-C -p package ;] or [-C --manifest-path=/some/path/Cargo.toml ;]. With `--deps`, the action will be applied to every local package in the dependency closure.
```

W95Psp:
bors cancel

PR #177, opened 2023-07-04
Concrete idents: per-backend name policies & reserved words for F*

W95Psp:
Inspecting a concrete ident is now done up to a *name policy*.
For now, a name policy only consists in a set of reserved words that should be escaped.


franziskuskiefer (approved):
Thanks, much needed :)
There's no test here again.
Couldn't we have something like `val` etc. in a test that we extract? We don't run F* on it yet, but will be soon. Then this should get tested.

PR #179, opened 2023-07-04
Fix missed `pp` (used when `[%show: Concrete_ident.t]`ing)

W95Psp:
PR #177 introduced a bug in the rust formatter: the concrete idents were printed using the wrong printer


franziskuskiefer (approved):
bors r+

W95Psp:
(note: this was merged, but something went wrong with the status of this PR)

PR #181, opened 2023-07-04
Add logs to Hax engine

W95Psp:


franziskuskiefer (approved):
Can you add logging to the Rust pieces as well?
https://crates.io/crates/log is the default way to go (it might be used in some places already, not sure).
The log levels should be harmonised. Maybe mimic the error, warn, info, debug, trace?

W95Psp:
Will do so in another PR

bors r+

PR #182, opened 2023-07-04
Fix rust printer for index fields

W95Psp:


franziskuskiefer (approved):
Thanks!
I guess I haven't noticed what looks like debug output before (because there was too much output here). But this looks like it should go as well?
```
# Entering phase (Reject RawOrMutPointer)
# Exiting phase (Reject RawOrMutPointer)
```
There's a bunch of these phase outputs.

bors r+

PR #185, opened 2023-07-05
Use `Re.Pcre` instead of `Re.Str`

W95Psp:
It's nowhere in `Re`'s doc, but it seems like `Re.Str` is a [reimplentation/a copy pase](https://github.com/ocaml/ocaml-re/blob/ddbc60f8e37e85f6b53b55fad1853cc97db93912/lib/str.ml) of OCaml's standard library, which is basically broken.

Using `Re.Str` resulted in... a stack overflow, when matching on big strings...

Using `Re.Pcre`, the issue is now gone.


franziskuskiefer (approved):
🙏🏻  Thanks for the quick fix

bors r+

PR #189, opened 2023-07-05
Explicitly disable the (required) jobs `docker` and `setup_sh` for

W95Psp:
anything else but merge groups.

The idea comes from https://github.com/bevyengine/bevy/issues/7824#issuecomment-1562380615.


pnmadelaine (commented):
lgtm

PR #192, opened 2023-07-06
Add `Ast_utils.add_typ_ascription`

W95Psp:
This utils adds type ascription nodes in nested function calls. This helps type inference in the presence of associated types in backends that don't support them well (F* for instance).

Examples:
  - `f (g x)` is turned into `f (g x <: τ)`;
  - `f (let x = g x in x)` is left unchanged (the F* backend annotates `let` already);
  - `f x + g x` is turned into `(f x <: τ₁) + (g x <: τ₂)`;
  - `f (g x <: τ)` is left unchanged (`g x` already has a type ascription).

(edit: there is no test for this since; that's backend-specific and it's noticeable only while F* typechecking)


franziskuskiefer (approved):
:shipit:

PR #194, opened 2023-07-06
Accept zero-arm matches, insert `never_to_any` coercition 

W95Psp:
With the never type (noted `!` in Rust), it's possible to match against empty enums. The engine used to assume a match had at least one arm. Now, we pick the first arm not typed `!`, and resort to `!` if no such arm exists.

Status: this PR introduces some TFalse nodes, which are not supported by Coq, gotta fix that

### DIFF
--- a/engine/lib/concrete_ident/names.rs
+++ b/engine/lib/concrete_ident/names.rs
@@ -105,8 +105,10 @@ mod hax {
 
     fn repeat() {}
     fn update_at() {}
-    // TODO: Should that live here: this is F* specific
+    // TODO: Should that live here? (this is F* specific)
     fn array_of_list() {}
+
+    fn never_to_any() {}
 
     mod control_flow_monad {
         trait ControlFlowMonad {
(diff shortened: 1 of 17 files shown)


engine/lib/concrete_ident/names.rs
```diff
@@ -105,9 +105,11 @@ mod hax {
 
     fn repeat() {}
     fn update_at() {}
-    // TODO: Should that live here: this is F* specific
+    // TODO: Should that live here? (this is F* specific)
```
franziskuskiefer, line 108: Can you add an issue here?

franziskuskiefer (approved):
Thanks, lgtm with one nit.

engine/lib/concrete_ident/names.rs
W95Psp: done #195

W95Psp:
Tests are not passing, actually. I have to fix some stuff

Issue #197, opened 2023-07-10
Memory issues in OCaml build


franziskuskiefer:
The build of the engine requires too much memory and runs out of memory on some machines because opam spawns too many processes.

To fix this:
 - [x] the `setup.sh` should start with fewer jobs and a `-j` option for faster builds should be added.
The readme needs to be updated with this information.

In addition:
 - [x] the build should be checked to understand why it requires so much memory.

W95Psp:
Fixed by #504!

PR #198, opened 2023-07-10
Improve setup and dependencies

W95Psp:
This PR:
 - refactors `setup.sh`;
 - adds checks for system dependencies (currently: node, rustup, opam, jq);
 - adds env var `OPAMERRLOGLEN` so that `opam` display error logs when an error occurs (otherwise users are facing useless error messages, for them as well as for us);
 - adds env var `OCAMLRUNPARAM` and `OPAMJOBS` to limit the amount of memory used, fixing #197.

Note regarding #197: other people are facing high memory consumption while compiling OCaml stuff, see https://discuss.ocaml.org/t/how-to-limit-the-amount-of-memory-the-ocaml-compiler-is-allowed-to-use/797/2.

### DIFF
--- a/engine/lib/concrete_ident/generate.sh
+++ b/engine/lib/concrete_ident/generate.sh
@@ -21,7 +21,7 @@ for json in $(
                     | sort -u
                ); do
     name=$(
-        echo "$json" | base64 --decode \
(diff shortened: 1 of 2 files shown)


franziskuskiefer (approved):
lgtm

PR #201, opened 2023-07-12
Cleanup frontend

W95Psp:


franziskuskiefer (changes requested):
I didn't look at the code yet but this doesn't actually terminate when running on a relatively small crate.

W95Psp:
uh, gonna debug that

W95Psp:
I added logging at various places.
I don't observe non-termination on my side though, I tried a few things, including Hax itself and openmls (that fails at some point).

franziskuskiefer (approved):

PR #202, opened 2023-07-13
implement `hax check`

### LINKED ISSUE #173: New subcommand `cargo hax check`
The subcommand `check`, just like `into` or `lint`, runs on a selection of crates.

For each selected crate `CRATE`, `check BACKEND`:
 1. make sure the current crate was extracted (maybe just a check the folder `CRATE-PATH/proof/BACKEND/extraction` exists)
 2. produces a list of paths for the fstar/coq/BACKEND modules from the dependencies of `CRATE`:
    - lists all the dependencies (indirect ones as well as direct ones);
    - make sure `cargo` downloaded the dependencies;
    - figure out where each dependency is located;
    - filter out those that have no `proof` directory;
    - build up a list of paths pointing to the `proof/BACKEND/{extraction,model,.}` directory of each dependency.
 3. calls `make` in folder `CRATE-PATH/proof/BACKEND/`, with `HAX_BACKEND_PATHS` env var set.

### Note about selecting crates
The selection is made by the user:
 - with `-C -p somepackage ;`, `-C --manifest-path some/path/to/Cargo.toml`, or any other option `cargo build` accepts.
 - or according to the current folder the command was ran into.

`check` however doesn't hook inside rustc, thus we need to figure out what crates are selected.
I think we can use something in the spirit of:
`cargo build OPTIONS_PASSED_WITH_C -Z unstable-options --build-plan | jq '.invocations | .[] | select(.env.CARGO_PRIMARY_PACKAGE) | .env.CARGO_MANIFEST_DIR'`

pnmadelaine:
fixes #173

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -419,7 +419,6 @@ dependencies = [
 name = "hax-driver"
 version = "0.1.0"
 dependencies = [
- "cargo_metadata",
  "clap",
  "const_format",
  "hax-cli-options",
@@ -473,6 +472,7 @@ dependencies = [
 name = "hax-subcommands"
 version = "0.1.0"
 dependencies = [
+ "cargo_metadata",
  "clap",
  "colored",
  "hax-cli-options",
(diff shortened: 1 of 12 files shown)


W95Psp:
Closing for now since it is stale, ping if you feel like reopening it.

PR #209, opened 2023-08-03
Exporter: disable the translation of `UserTy` for now

W95Psp:
@franziskuskiefer found the following bug:
```rust
pub const MY_VAL: u16 = 5;
pub type Alias = MyStruct<MY_VAL>; // Using the literal 5, it goes through

pub struct MyStruct<const VAL: u16> {}

impl<const VAL: u16> MyStruct<VAL> {
    pub const MY_CONST: u16 = VAL;
}

pub fn do_something() -> u32 {
    u32::from(Alias::MY_CONST)
}
```

The `Alias::MY_CONST` in `do_something` yields a `UserTy` node in the translated AST, which contains a Debruijn index, that we don't handle properly. Since we never use `UserTy`, this PR just disable its translation, thus fixing the issue.

### DIFF
--- a/frontend/exporter/src/types/copied.rs
+++ b/frontend/exporter/src/types/copied.rs
@@ -369,8 +369,29 @@ pub struct UserSubsts {
 #[derive(AdtInto, Clone, Debug, Serialize, Deserialize, JsonSchema)]
 #[args(<'tcx, S: BaseState<'tcx>>, from: rustc_middle::ty::UserType<'tcx>, state: S as gstate)]
 pub enum UserType {
-    Ty(Ty),
-    TypeOf(DefId, UserSubsts),
+    // TODO: for now, we don't use user types at all.
+    // We disable it for now, since it cause the following to fail:
+    //
+    //    pub const MY_VAL: u16 = 5;
+    //    pub type Alias = MyStruct<MY_VAL>; // Using the literal 5, it goes through
+    //
+    //    pub struct MyStruct<const VAL: u16> {}
+    //
+    //    impl<const VAL: u16> MyStruct<VAL> {
+    //        pub const MY_CONST: u16 = VAL;
+    //    }
+    //
+    //    pub fn do_something() -> u32 {
+    //        u32::from(Alias::MY_CONST)
+    //    }
+    //
+    // In this case, we get a [rustc_middle::ty::ConstKind::Bound] in
+    // [do_something], which we are not able to translate.
+
+    // Ty(Ty),
+    // TypeOf(DefId, UserSubsts),
+    #[todo]
+    Todo(String),
 }
 
 #[derive(AdtInto, Clone, Debug, Serialize, Deserialize, JsonSchema)]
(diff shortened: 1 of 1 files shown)


frontend/exporter/src/types/copied.rs
```diff
@@ -369,8 +369,29 @@ pub struct UserSubsts {
...
+    //    impl<const VAL: u16> MyStruct<VAL> {
+    //        pub const MY_CONST: u16 = VAL;
+    //    }
+    //
+    //    pub fn do_something() -> u32 {
+    //        u32::from(Alias::MY_CONST)
+    //    }
+    //
+    // In this case, we get a [rustc_middle::ty::ConstKind::Bound] in
+    // [do_something], which we are not able to translate.
```
franziskuskiefer, line 389:
```suggestion
    // [do_something], which we are not able to translate.
    // See: https://github.com/hacspec/hacspec-v2/pull/209
```

franziskuskiefer (changes requested):
Hm, on the original code I get another panic now in ocaml

```
Fatal error: exception Failure("parsing error: parse_user_type LINE=6112 JSON={ \"Todo\": \"Ty(u32)\" }")
Raised at Stdlib.failwith in file "stdlib.ml", line 29, characters 17-33
Called from Hax_engine__Types.parse_canonical_for__user_type in file "lib/types.ml", line 1448, characters 11-34
Called from Hax_engine__Types.parse_canonical_user_type_annotation in file "lib/types.ml", line 1367, characters 11-49
Called from Hax_engine__Types.parse_ascription in file "lib/types.ml", line 998, characters 11-55
Called from Hax_engine__Types.parse_pat_kind in file "lib/types.ml", line 4796, characters 11-35
Called from Hax_engine__Types.parse_decorated_for__pat_kind in file "lib/types.ml", line 1599, characters 11-33
Called from Hax_engine__Types.parse_stmt_kind in file "lib/types.ml", line 5504, characters 11-48
Called from Hax_engine__Types.parse_stmt in file "lib/types.ml", line 5430, characters 11-34
Called from Stdlib__List.map in file "list.ml", line 92, characters 20-23
Called from Hax_engine__Types.parse_expr_kind in file "lib/types.ml", line 2317, characters 11-238
Called from Hax_engine__Types.parse_decorated_for__expr_kind in file "lib/types.ml", line 1546, characters 11-34
Called from Hax_engine__Types.parse_fn_def in file "lib/types.ml", line 2995, characters 11-49
Called from Hax_engine__Types.parse_item_kind in file "lib/types.ml", line 3907, characters 11-31
Called from Hax_engine__Types.parse_item in file "lib/types.ml", line 3805, characters 11-34
...
Re-raised at Lib.main in file "bin/lib.ml", line 80, characters 6-42
Called from Dune__exe__Native_driver in file "bin/native_driver.ml", line 1, characters 8-19
error: hax-engine exited with non-zero code 2
       stdout:
        stderr:
```
I can try to make another reproducer for this. The small example here goes through.

W95Psp:
Oh, that's odd! To me, that sounds like the generated OCaml module for parsing the types from Rust is outdated. Sometimes dune misses changes... can you try removing the `hax-engine` binary (somewhere under `.opam` I guess?), `dune clean` in `engine` and recompile everything?

(also: if you used `rebuild.sh`, it's installing hax-engine under `~/.cargo/bin`, so maybe it's picking that old one)

franziskuskiefer (approved):
Indeed, it looks like the `setup` script should `reinstall` the opam package rather than `install`.?

PR #211, opened 2023-08-10
Add attributes `#[skip]` and refinement attributes

W95Psp:
Example:
```rust
use hax_lib_macros::*;

#[skip]
pub fn skip_me() {}

#[hax]
pub struct Hello {
    pub x: u32,
    #[refine(y > 3)]
    pub y: u32,
    #[refine(y + x + z > 3)]
    pub z: u32,
}
```

Translates into the following F*:
```fstar
module Attributes
open Core

let v___: Prims.unit = ()

let refinement_under___ (x y: u32) : bool = y >. 3ul

let v___1: Prims.unit = ()

let refinement_under___1 (x y z: u32) : bool = ((y +. x <: u32) +. z <: u32) >. 3ul

type t_Hello = {
  f_x:u32;
  f_y:f_y: u32{f_y >. 3ul};
  f_z:f_z: u32{((f_y +. f_x <: u32) +. f_z <: u32) >. 3ul}
}
```

I need to add more attributes and clean up the `refinement_under__*` things

TODO (follow up PRs):
 - [ ] a `#[hax::late_skip]` (internal thing: translate to the backend, but don't print)
 - [ ] refinements on functions inputs & outputs (follow up PR)
 - [ ] pre/post on functions, contract style
 - [ ] `#[refine(...)]` on enum fields
 - [ ] `#[hax::lemma]` for marking a function and express that this function is a lemma statement we want to prove on the proof assistant side 
 - please comment if other ideas


tests/attributes/src/lib.rs
```diff
@@ -1,47 +1,13 @@
...
-       diam.
-    */
-    #[hax::a::path(Bar:field2 attr)]
-    field2: u32,
+use hax_lib_macros::*;
+
+#[skip]
+pub fn skip_me() {}
+
+#[hax]
```
franziskuskiefer, line 6: What's the semantic of these two attributes?
We need to write this down somewhere. At least with doc comments on the attributes.

For `skip` there's probably only one possibility what it should do (still needs to be written down).
`hax` for me only makes sense if everything else in the crate without the attribute is ignored for translation.
With this semantic this test doesn't do much though.

tests/raw-attributes/src/lib.rs
franziskuskiefer, line 1: Please add comments to tests with what they are supposed to test.

hax-lib-macros/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+version = "0.1.0"
+edition = "2021"
+
+[lib]
+proc-macro = true
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+lazy_static = "1.4.0"
```
franziskuskiefer, line 12: This appears not to be used.

engine/utils/ocaml_of_json_schema/ocaml_of_json_schema.js
```diff
@@ -208,6 +220,10 @@ let parse_type_name = s => {
     return s.split('/').slice(-1)[0];
 };
 
+let int_repr_of_format = format => 
+  (format.endsWith('int128') || format == 'uint64' || format == 'uint' /*`uint`s are `usize`s actually, so that's safer to assume it's a uint64, see https://github.com/GREsau/schemars/blob/386e3d7f5ac601795fb4e247291bbef31512ded3/schemars/src/json_schema_impls/primitives.rs#L85C16-L85C21*/)
```
franziskuskiefer, line 224: 🙈

franziskuskiefer (approved):
Thanks, I think there's enough good stuff in here to get it in.

Did you remove the `hax` attribute again? We should have it at some point. But `skip` is more important for now I think.

engine/utils/ocaml_of_json_schema/ocaml_of_json_schema.js
W95Psp: Ahaha, yes, that is really terrible... That caused me quite a lot of pain

W95Psp:
I just renamed that `hax` attribute to `hax_attributes`: when an item is marked with this attribute, anything below it can use the `refine` attribute (where it's allowed --for now only on struct fields).

franziskuskiefer:
> I just renamed that `hax` attribute to `hax_attributes`: when an item is marked with this attribute, anything below it can use the `refine` attribute (where it's allowed --for now only on struct fields).

Right, but we want a `hax` attribute that disables everything in a crate except for the annotated functions to extract code from large crates. But that's not urgent yet. Can you file a follow up for it?

W95Psp:
Oh, right! Sure, filed #223 :)

franziskuskiefer:
@W95Psp this needs manual conflict resolution

Issue #152, opened 2023-06-21
Invalid "cycle detected when building THIR"


franziskuskiefer:
The following code

```rust
fn inner_fun<const LEN: usize>() -> [u8; LEN] {
    [0u8; LEN]
}

pub fn fun() {
    const LEN: usize = 5;
    inner_fun::<LEN>();
}
```

generates the error

```
error[E0391]: cycle detected when building THIR for `fun::{constant#0}`
   --> src/lib.rs:166:17
    |
166 |     inner_fun::<LEN>();
    |                 ^^^
    |
note: ...which requires type-checking `fun::{constant#0}`...
   --> src/lib.rs:166:17
    |
166 |     inner_fun::<LEN>();
    |                 ^^^
...
```

when running `cargo hax into fstar`.

Since `cargo +nightly-2023-06-02 build` works fine this doesn't appear to be an actual `rustc` issue but a `hax` bug.

Note that the error goes away when using the literal `5` directly instead of defining the `const LEN` first.

W95Psp:
After investigating, that's indeed related to the way the exporter browses & typecheck the crate.

This is related to [feed_anon_const_type](http://localhost:8000/rustc_middle/ty/context/struct.TyCtxt.html#method.feed_anon_const_type)

franziskuskiefer:
Do you have any updates on this @W95Psp ?

W95Psp:
Status: this really occurs less, but still, it occurs

PR #212, opened 2023-08-11
Exporter: misc. improvements

W95Psp:
This PR fixes a few things:
 - more `steal`ing issues (but not all, I found some, one in `Core`, notably);
 - misc. bugs with char consts, opaque types and generic bounds;
 - we now translate correctly `PhantomData`s.

Those fixes are for bugs I stumbled upon while looking at big crate (`Core`, but not only), thus that's generally pretty hard to minimize. (that's why this PR doesn't ship unit tests)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -422,6 +422,7 @@ dependencies = [
  "hax-frontend-exporter-options",
  "hax-lint",
  "inquire",
+ "itertools 0.11.0",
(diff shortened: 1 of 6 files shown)


franziskuskiefer (approved):
Thanks! Looks like we are good for most stealing issues now 🥳 

Is #152 resolved with this as well?

PR #213, opened 2023-08-17
Rework frontend errors

W95Psp:
This PR provides a more uniform way of reporting errors in the exporter.

### DIFF
--- a/frontend/exporter/src/utils.rs
+++ b/frontend/exporter/src/utils.rs
@@ -57,11 +57,9 @@ macro_rules! report {
     };
 }
 
-#[allow(unused_macros)]
 macro_rules! error { ($($tt:tt)*) => {$crate::utils::report!(error, $($tt)*)} }
 #[allow(unused_macros)]
 macro_rules! warning { ($($tt:tt)*) => {$crate::utils::report!(warn, $($tt)*)} }
-#[allow(unused_macros)]
 macro_rules! fatal { ($($tt:tt)*) => {$crate::utils::report!(fatal, $($tt)*)} }
 
 pub(crate) use format_with_context;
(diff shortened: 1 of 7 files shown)


frontend/exporter/src/utils.rs
```diff
@@ -1,57 +1,107 @@
...
+            $verb,
+            $s,
+            $s.base().opt_def_id.map(|did| $s.base().tcx.def_span(did)),
+            $crate::utils::format_with_context!($($tt)*)
+        )
+    };
+}
 
-Please report that message along with the crate!
+#[allow(unused_macros)]
```
franziskuskiefer, line 60: Are these all unused or why do we need the `allow`s here?

frontend/exporter/src/utils.rs
```diff
@@ -1,57 +1,107 @@
...
-        }}
+    macro_rules! _span_verb_base {
+        ($verb:ident, $s:ident, $span:expr, $message:expr) => {{
+            let mut builder = $crate::utils::_verb!($verb, $s.base().tcx.sess, $message);
+            if let Some(span) = $span {
+                builder.set_span(span.clone());
+            }
+            builder.code(rustc_errors::DiagnosticId::Error(format!("HaxFront")));
+            builder.note(
+                "⚠️ This is a bug in Hax's frontend.
```
franziskuskiefer, line 35: Add a linkt to the issues in this crate.

franziskuskiefer (approved):
Thanks

frontend/exporter/src/utils.rs
W95Psp: Only `warning` is unused, actually, removing those other superfluous `allow`s!

PR #217, opened 2023-08-18
allow setting -j in setup and uninstall before installing

franziskuskiefer:


W95Psp (approved):
Thanks, LGTM :)

franziskuskiefer:
Ok... looks like opam on Linux doesn't like the uninstall 🙄 
I'll try something else.

W95Psp:
Maybe we can `opam uninstall ... || true`?

PR #228, opened 2023-08-22
Fix #225

### LINKED ISSUE #225: foreach loops not generating unit return value
Seems there are some issues with return values of foreach loops, as this
```
for zkp in zkps {
    check_valid(zkp);
}
```
results in the body
```
check_valid zkp
```
while
```
for zkp in zkps {
    check_valid(zkp); ()
}
```
generates
```
let _ := check_valid zkp in
tt
```
there should not be any difference returning unit or having `;` on the last line.

W95Psp:
I think this should fix your issue @cmester0 :)

### DIFF
--- a/engine/lib/phases/phase_reconstruct_for_loops.ml
+++ b/engine/lib/phases/phase_reconstruct_for_loops.ml
@@ -190,28 +190,7 @@ struct
                                                                           _;
                                                                         };
                                                                   };
-                                                                body =
-                                                                  ( {
-                                                                      e =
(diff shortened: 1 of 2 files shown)


franziskuskiefer (approved):
lgtm if @cmester0 is happy with it.

W95Psp:
Merging this, @cmester0, tell me if there's anything off with this PR, if that's the case, we'll fix that in another PR
(I really think that solves the issue, should be fine!)

PR #234, opened 2023-08-24
Additions to "Implement better support for MIR"

W95Psp:


frontend/exporter/src/types/copied.rs
```diff
@@ -353,55 +205,38 @@ pub struct Scope {
...
-                let kind = ConstantExprKind::GlobalName { id };
-                Decorated {
-                    ty: ty.sinto(s),
-                    span: span.sinto(s),
-                    contents: Box::new(kind),
-                    hir_id: None,
-                    attributes: vec![],
+                if is_anon_const(ucv.def, tcx) {
+                    let evaluated_c = self.eval(tcx, get_param_env(s));
+                    if &evaluated_c == self {
```
sonmarcho, line 223: Are you sure this works? Shouldn't we do `if let ...::Unevaluated = evaluated_c { ... }`?

frontend/exporter/src/constant_utils.rs
```diff
@@ -204,76 +192,72 @@ pub(crate) fn scalar_to_constant_expr<'tcx, S: BaseState<'tcx>>(
...
-        ty: ty.sinto(s),
-        span: cspan,
-        contents: Box::new(kind),
-        hir_id: None,
-        attributes: vec![],
-    }
+    kind.decorate(ty.sinto(s), cspan)
 }
 
-/// We evaluate the anonymous constants
```
sonmarcho, line 236: The reason why I had put written this function was to make sure the evaluation policy was consistent across the code base.

W95Psp: Makes sense indeed! Just factored out the handling of unevaluated constants in 0514b83b4bd2965e9569a745919eb3a9b492999b

frontend/exporter/src/types/copied.rs
W95Psp: Yes, the implementation of the `eval` methods returns the constant as it is

frontend/exporter/src/types/copied.rs
```diff
@@ -1791,6 +1581,7 @@ pub struct VariantInformations {
 
     pub typ: DefId,
     pub variant: DefId,
+    pub variant_index: VariantIdx,
```
sonmarcho, line 1584: I need this to be an option : `None` if the ADT is a structure

sonmarcho (approved):
Looks good! You can merge whenever you want, modulo the comment I left for `VariantInformation` :)

frontend/exporter/src/types/copied.rs
W95Psp: I removed that option because of the field `typ_is_struct` below: isn't that what you need?

sonmarcho: Indeed, I hadn't noticed. That works for me.

Issue #231, opened 2023-08-24
`&mut` bugs


franziskuskiefer:
This issue is collecting bugs in the handling of `&mut`.

The following code is throwing errors.

```rust
let mut bytes = vec![0u8, 1, 2]; // if this is an array, this works
bytes[0] ^= 0xFF; // bytes[0] = bytes[0] ^ 0xFF is accepted
```

```
error[HAX0010]: (RefMut) At this position, Hax was expecting an expression of the shape `&mut _`. Hax forbids `f(x)` (where `f` expects a mutable reference as input) when `x` is not a place expression[1] or when it is a dereference expression.
```

Another instance of this is the following
```
error[HAX0002]: (AndMutDefSite) Fatal error: something we considered as impossible occurred! Please report this by submitting an issue on GitHub!
                Details: Place.of_expr: got `None` for: core::ops::bit::BitAnd::bitand(core::ops::index::Index::index(out, 31), 127)
   --> src/ecdh.rs:243:27
    |
243 |                 out[31] = out[31] & 127u8;
    |                           ^^^^^^^^^^^^^^^
```

W95Psp:
The first snippet of code is failing to extract because the type of `bytes[0]` is an associated type. For now, Hax is not able to look those up. Though, associated types will soon be resolved thanks to the great work of @sonmarcho :smiley:

franziskuskiefer:
> or now, Hax is not able to look those up.

That's a pretty big issue. This is used all over the place.

W95Psp:
Yes, that's the long-standing #20 issue

sonmarcho:
> > or now, Hax is not able to look those up.
> 
> That's a pretty big issue. This is used all over the place. Setting this to P1

Yes, not having support for traits is a big issue, because traits appear everywhere. But I'm almost done with extracting trait information from the Rust compiler, so that should be gone soon. I will keep both of you posted.

W95Psp:
(well, we had support for traits via typeclass resolution in the backends, but we had no pre-resolved instance info :smiley:)

W95Psp:
Update on that:

Now we have associated types correctly resolved, this bug still persist, and is related to to the way `x <OP>= e` is desugared in [import thir](
https://github.com/hacspec/hax/blob/eebfe9b38a621d924794379553a9b263ba7d32bd/engine/lib/import_thir.ml#L499-L501C11).

[...]

PR #240, opened 2023-08-31
Fix uncatched errors in the engine

W95Psp:
This should fix the fatal error you observed in the engine @franziskuskiefer


franziskuskiefer (approved):
Thanks, this fixes the error.
The culprit is the code from #231 .

Issue #121, opened 2023-06-06
Debugging facilities: improve `--debug_engine`


W95Psp:
Currently, enabling the flag `--debug_engine` makes the engine export all items at every phase.
This is quite slow, hence we need to:
 - [ ] add a flag to filter out/in phases
 - [ ] add a flag to filter out/in items/namespaces

PR #242, opened 2023-08-31
Make `--engine-debug` easy to use

W95Psp:
This PR:
 - rewrites the tiny webserver serving the webapp for debugging the engine in Rust;
 - embed this webserver directly into the Rustc driver;
 - when `--engine-debug` (or `-d`) is passed to the CLI, the driver now spawns a webserver directly so that debugging is very easy.

(this is somehow related to #121)


engine/utils/phase_debug_webapp/src/main.rs
```diff
@@ -0,0 +1,5 @@
+use phase_debug_webapp::*;
+
+fn main() {
+    run(|| std::fs::read_to_string("/tmp/hax_debug/debug-hax-engine.json").unwrap())
```
franziskuskiefer, line 4: A hard-coded path? I'm not sure that's a good idea....

franziskuskiefer (changes requested):
Yay, only one comment

engine/utils/phase_debug_webapp/src/main.rs
W95Psp: Oops, that's was not intended

W95Psp: Fixed it: I actually removed this binary entirely, it has no purpose now

engine/DEV.md
```diff
@@ -46,10 +46,7 @@ You can also use `dune utop` and then `#show_type Hax_engine.Types.SOME_TYPE` an
...
-Just add the flag `--debug-engine DIR` to the `into` subcommand, where
-`DIR` is a directory where will be outputted debug informations. Make
-sure to create the folder beforehand.
-
-To browse the debug information, run
-`./engine/utils/phase_debug_webapp/server.js DIR`. Note you can change
-to port by setting the environment variable `PORT`.
+Just add the flag `--debug-engine` (or `-d`) to the `into` subcommand.
+At the end of the translation, `cargo hax` will spawn a webserver that
+let you browse the debug information. Note you can change to port by
```
franziskuskiefer, line 51:
```suggestion
lets you browse the debug information. Note, you can change to port by
```

franziskuskiefer (approved):

PR #243, opened 2023-09-03
Extract trait resolution information from Rustc

sonmarcho:
The following PR extracts the results of trait resolution from the Rust compiler. I only touched the Rust code for MIR so there is work to do to propagate the changes to THIR and ML, and I haven't merged the recent changes (in particular, I haven't merged https://github.com/hacspec/hacspec-v2/pull/234 yet). Still, I want to proceed with submitting the PR so that the reviewing process can start and the addition of traits is not blocked on waiting for me. I'm currently finishing to propagate the changes to Charon and Aeneas, but the code on the hacpsec side now seems stable.

More specifically, this PR implements helpers which query the trait resolution mechanism to recursively solve trait obligations (a function may have trait bounds that we need to solve at call site, but the implementations which satisfy those bounds may themselves have bounds, etc.). It uses those helpers to add trait resolution information to the MIR and to `Ty`. Adding trait resolution for THIR should mostly be a matter of introducing calls to those helpers at the appropriate locations. All the code is in `frontend/exporter/src/types/traits.rs`.

Below, I review the most important information that code maintainers and users should be aware of.

# Representation
When adding information about trait resolution, I pretend that traits are "functor" instances that we explicitly reference and instantiate. For instance:
```rust
trait Trait<T, U> {
  fn f<V>(self, ...) -> (T, V);
}

impl<T> Trait<T, usize> for Foo { ... } // let's call this implementation Impl1

fn f<T, V>(x : Foo) {
  let y : (T, V) = x.f(...) // We desugar this to: `Impl1<T>::f<V>(x, ...)`
  ...
}
```

Importantly, I pay attention to splitting the generics given to `f` between generics which apply to the trait instance (`Impl1<T>`) and generics which actually apply to the method (`f<V>`). This may seem like a technical detail but is not for code maintainers because the Rust compiler doesn't provide the information like in the above shape: if you query the Rust compiler, you get that all the generics apply to `f` - I thus manually split them.

# Trait parameters
One issue remains with trait parameters, that I refer to as "local clauses". If the result of trait resolution is a top-level implementation Rust gives us its exact id. However, if it is a trait parameter, Rust doesn't tell us which exact clause solves the trait obligation (it just tells us it is one of them). This makes sense in the context that trait resolution is used by Rustc to check that trait obligations are satisfied and for codegen, but it is a big pain on our side because we later on have to redo a limited trait resolution for the parameters (I do it in Charon). In this regard, one big issue is that a trait obligation may be solved not only by the available trait clauses but also by a clause in their bounds. With an example:

```rust
trait ToU64 { fn to_u64(&self) -> u64; }
trait Foo : ToU64 { type S : ToU64; }
//          ^^^^^^           ^^^^^^
//          bound 1          bound 2
//      (parent bound)     (bound for item S)

fn f<T : Foo>(x : T, y : (T as Foo)::S) -> u64 {
  //     ^^^
  //   This is the only "local" clause we directly see
  let a = x.to_u64(); // This requires bound 1
  let b = y.to_u64(); // This requires bound 2
  a + b
}
```

In order to make the work easier, I extended the `TraitPredicate` with what I call the "parent predicates" (bound 1 above) and the "item predicates" (bound 2 above). This way, it is possible to later solve the parameter traits without doing any lookup.

**Remark (trait parameters):** if we forbid higher-order types (having universal quantifiers on, say, lifetimes), trait type constraints (`where T::S = usize`) and specialization on trait parameters (though I'm not completely sure we need to forbid this one), we already support a quite expressive subset for which solving the parameter trait obligations is just a matter of doing equality checks (this is what I do in Charon). If we add trait type constraints, we may need to perform those equality checks modulo equivalence classes (but maybe not: it depends on whether the types in the trait obligations have been normalized - for this we need to check in detail what the output of hacspec is on specific examples, and we may update the code to adjust this output).

**Remark (`Self`):** it often happens that trait definitions refer to themselves, like so:
```rust
trait Trait {
  type S;

  fn f(x : Self::S);
  //         ^^^^^^
}
```
When querying Rust, the type for `x` we see is `Trait<T>::S` (where `T` is automatically introduced by rustc and refers to the `Self` type). As a result the trait obligation it generates is not fundamentally different from other trait obligations that are solved by using trait parameters. In order to correctly handle this, it is useful to have a `Self` clause which gives us exactly that `T : Trait` (or, without sugar, a clause which gives us an instance of `Trait<T>`): if whenever we solve a trait obligation it is actually matched by this clause, we know we should use `Self` (this is what I do in Charon). Actually, this clause is exactly the first trait clause returned by `TyCtxt::predicates_of` which, according to the rustc documentation, returns the same predicates as `TyCtxt::predicates_defined_on` with, for trait items, an additional predicate corresponding to `Self : Trait`.


franziskuskiefer:
Nice, thanks!

@sonmarcho there are some conflicts (probably from merging #234). Can you fix them?

@W95Psp do you want to get this in as is and then do the thir/ml parts or do you want to put them on top of this first? In other words, shall I review this or wait for changes from you?

W95Psp:
I haven't reviewed it either, so I don't know yet, but I think we should merge the PR as it is, just fixing whatever broke in the engine before.
The plan is the following:
 - merge main in #221 (this will require a bit of work),
 - review and fix engine in this PR,
 - merge main in this PR (I hope this should be fast),
 - merge that PR

franziskuskiefer (approved):
Generally looks good to me. I'll leave the detailed review to Lucas.

W95Psp:
There's a bug when traits are sort of recursive:
```rust
trait Foo: Bar {
    type U;
}
trait Bar<R = <Self as Foo>::U> {}
```

This yields a stack overflow.

However I think we should just never translate default types: Rustc inline them for us in `impl`s that omit them

W95Psp:
Closing this: traits were added in parallel to this PR in https://github.com/hacspec/hax/pull/249.
@sonmarcho and I are merging hax and charon traits in an upcoming PR

PR #252, opened 2023-09-12
Generic printer

W95Psp:
Start an exploration of a generic printer, aiming at factorizing backends a little.
In this PR the default printer is a Rust-like printer for the whole AST; the idea is to have a printer per family of languages.
For instance, Lean, Coq and F* share a lot in terms of structure.

This implementation uses OCaml classes (surprisingly!). This is somehow a bit nice here because we want to be able to refine an printer in another printer: we could have a `FunctionalFlavorPrinter` as a basis for `CoqPrinter`, `FStarPrinter` an d `LeanPrinter` that would override some parts of `FunctionalFlavorPrinter`, but inherit other parts.
In OCaml, I think OOP is the only way to achieve that. In type-class capable languages, we could use a double-dispatch thing with typeclasses and default implementation of methods. But in OCaml, I think OOP is the only sane way.

WDYT @cmester0?
 
(cc @R1kM also, since you're looking into having a Lean backend)

### DIFF
--- a/engine/lib/diagnostics.ml
+++ b/engine/lib/diagnostics.ml
@@ -54,6 +54,7 @@ module Context = struct
     | Backend of Backend.t
     | ThirImport
     | DebugPrintRust
+    | GenericPrinter of string
     | Other of string
   [@@deriving show, eq, yojson, compare]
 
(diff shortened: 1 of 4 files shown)


W95Psp:
The generic printer is working fine, I rebased it on main.
I think we should merge it; for now it's deadcode, it's used nowhere, but having it in main would be nice so that we can update it as the AST changes, and that will make things easier to switch from existing backends or printers to this generic printer. (I'm especially thinking about the Rust printer)

franziskuskiefer (approved):
Ok, lets get this in and make sure that we put a printer on top soon (Rust or PV or something else) to test the code as well.

PR #257, opened 2023-09-19
Update README.md

franziskuskiefer:
This shouldn't be needed anymore


W95Psp (approved):
Nice, that's great if it's not needed anymore :partying_face:

PR #261, opened 2023-09-25
Revive JS

W95Psp:
This PR:
 - Revives the JS extraction of the Hax engine:
   + was extremely slow due to poor string performance (I wrote a small helper that made parsing JSON go from one minute to instant on big JSON payloads...);
- Add an option `--hax-engine-path` to the `into` command so that one can specify a path to:
   + a native hax engine;
   + a JS hax engine;
   + auto detect.
- Make auto detect able to download from dists on the internet:
   + for now, this points to a place that doesn't exists.

Proposal for JS dists:
 - **Side note:** the (rust) Hax frontend defines the types, and thus sets the tone. To ensure compatibility, this PR introduces `RUST_SRC_HASH`, a hash that captures all the rust sources of the frontend. `RUST_SRC_HASH` is inlined in the frontend binary (it can be queried via `cargo hax -V`).
 - In the frontend, when "resolving" the engine, the goal is to find a binary or a JS dist built in an environment with a particular `RUST_SRC_HASH`.
 - Thus, my proposal is to upload `${RUST_SRC_HASH}.js` each time a commit is pushed to `main` to some location we can query via HTTP. A simple thing would be github pages, with a dedicated repo.


franziskuskiefer (changes requested):
As discussed offline

W95Psp:
Closing, instead, I'm going to:
 - open a PR for just reviving the JS compilation;
 - open (later) a PR for release / download of JS dists.

PR #265, opened 2023-10-02
Revive js engine

W95Psp:
First half of #261:
 - Add a few missing features to the JSON schema to OCaml script;
 - Use native JS JSON parser while compiling the engine with JS.

### DIFF
--- a/engine/bin/js_driver.ml
+++ b/engine/bin/js_driver.ml
@@ -1,1 +1,85 @@
-let _ = Lib.main (Lib.read_options_from_stdin Yojson.Safe.from_string))
+open Base
+open Js_of_ocaml
+
+(* Strings are slow with js_of_ocaml. Thus, parsing a string into a
+   `Yojson.Safe.t` is extremly slow using yojson itself. Instead, I
+   wrote a very simple and stupid `yojson_of_string_via_js` that (1)
+   parses the json out of a JS string into a JS object (2) make a
+   yojson AST. This is something like x100 faster. Without this hack,
+   the performance is too bad to be bearable. *)
+let yojson_of_string_via_js (s : string) : Yojson.Safe.t =
+  let f =
+    Js.Unsafe.js_expr
+      {js|
+(function (mkNull, mkBool, mkBigint, mkInt, mkFloat, mkString, mkDict, mkArray){
+  function isInt(n) {
+    return n % 1 === 0;
+  }
+  function f(x){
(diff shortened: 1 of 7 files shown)


engine/bin/js_driver.ml
```diff
@@ -1 +1,85 @@
...
+open Base
+open Js_of_ocaml
+
+(* Strings are slow with js_of_ocaml. Thus, parsing a string into a
+   `Yojson.Safe.t` is extremly slow using yojson itself. Instead, I
+   wrote a very simple and stupid `yojson_of_string_via_js` that (1)
+   parses the json out of a JS string into a JS object (2) make a
+   yojson AST. This is something like x100 faster. Without this hack,
+   the performance is too bad to be bearable. *)
+let yojson_of_string_via_js (s : string) : Yojson.Safe.t =
```
franziskuskiefer, line 10: So this is only used when using the JS engine, right? Ideally we get rid of `ocam_of_json_schema.js` soon to get rid of the JS dependency. Then we shouldn't introduce a new one here. That is unless someone runs the engine in JS instead of ocaml of course.

franziskuskiefer (approved):

engine/bin/js_driver.ml
W95Psp: This chunk of code (and the whole `js_driver.ml` module really) is only the entry point to the engine when compiled with OCaml then js_of_ocaml.
When compiling, either the entry point is `native.ml` or `js_driver.ml`.

That's useful only when running the engine with NodeJS instead of running it natively. This scenario is interesting only for easy distribution (since JS engines are virtually available absolutely everywhere without any install complexity while that's far from being true for a good OCaml building env).
But in that case, the dependency to NodeJS is runtime.

The `ocaml_of_json_schema.js` script is a build-time-only dependency and is completely orthogonal to that ML module.

So I don't think that's adding any dependency.

Also, concerning getting rid of `ocaml_of_json_schema`, that would mean rewriting it, this will take some time and effort while it's working very well (it basically never broke), so I don't see the point of doing that any time soon. Especially, if we distribute the engine as JS, that makes only:
  - rust toolchain as a build (& runtime) dependency;
  - node as a runtime dependency.

franziskuskiefer: 👍🏻

W95Psp:
Thanks for the review & the merge :)

PR #268, opened 2023-10-04
fix(front/adt-into): make `map` work nicely with `not_in_source`

W95Psp:


frontend/exporter/adt-into/src/lib.rs
```diff
@@ -226,6 +226,11 @@ fn variant_to_arm(
         // TODO: refactor
```
franziskuskiefer, line 226: While you're here ... Can you add more information here and an issue or remove it?

franziskuskiefer (approved):

frontend/exporter/adt-into/src/lib.rs
W95Psp: I removed those, I don't recall what I meant by that :/

PR #270, opened 2023-10-04
refactor(engine/ast): `*i_name: string` -> `*i_ident: concrete_ident`

W95Psp:
As @cmester0 noticed some while ago, `ti_name` and `ii_name` are strings while any other global identity is usually a `concrete_ident` (or a `global_ident` when that makes sense).
This PR renames and retypes the fields `ti_name` and `ii_name` into `ti_ident` and `ii_ident`.


engine/lib/import_thir.ml
```diff
@@ -1253,6 +1255,13 @@ module Make (Opts : OPTS) : MakeT = struct
                  items =
                    List.map
                      ~f:(fun (item : Thir.impl_item) ->
+                       (* TODO: introduce a Kind.TraitImplItem or
```
franziskuskiefer, line 1258: Can this get an issue?

franziskuskiefer (approved):

PR #275, opened 2023-10-06
F* backend: always add type annotations on `cast`s

W95Psp:


franziskuskiefer (approved):
I assume `into`/`from` casts are handled by the corresponding functions?

PR #284, opened 2023-10-11
Semantic

cmester0:
Trying to make hax/hacspec semantic explicit.


W95Psp:
Hi @cmester0,

I guess we'll discuss that this afternoon, but the idea here is to define an executable semantic? i.e. target some AST and give it a semantic with an interpreter?
Would be awesome :)

cmester0:
Attempt at defining a common (deeply embedded) semantic to enable cross backend reasoning. This has been put on hold for now, so we will close this PR for now.

PR #286, opened 2023-10-13
Improve namings for `impl` blocks

W95Psp:
The main change with this PR is described in commit message of "feat(engine/idents): improve `impl` naming":
> This commit introduces some smarter naming for non-inherent `impl`s.
> Consider a `impl<> Trait<> for Type<>`. Note by `<>` I do mean empty generics.
> 
> If:
> 1. the `impl` block and the `Trait` trait lives under the same
>       namespace;
> 2. the type `Type` is simpleⁱ or is an user-defined enum or struct
>      without generics that live in the same namespace as the `impl`
>       block.
> 
> Then:
>       the name of the `impl` block is not `impl_DISABIGUATOR` but
>       `impl_<Trait>_for_<Type>`.
> 
> ⁱ: a simple type is a type we can "serialize" to a string
>     identifier. Typically, a (unique) identifier representation of
>     `(Vec<(A, B)>, Fn(A) -> C)` would be cumbersome. We allow only
>     applications of types of arity 1 or less.

Other than that, this PR also:
 - adds structure and information to the `impl_infos` exported by the frontend;
 - improves a bit the test harness;
 - adds tests;
 - fix a bug in the `refmut` phase (see the comment added by commit 
`fix(engine/phase/refmut): preserve original type`).


tests/naming/src/lib.rs
```diff
@@ -63,3 +63,14 @@ fn f(x: Foobar) -> usize {
 fn reserved_names(val: u8, noeq: u8, of: u8) -> u8 {
     val + noeq + of
 }
+
+struct Arity1<T>;
```
franziskuskiefer, line 67: This is not a valid struct. `T` must be used somehow (at least with phantom data).

franziskuskiefer (approved):
thanks, lgtm with that one nit.

tests/naming/src/lib.rs
W95Psp: Oh right! Thanks, was only compiling though Hax (I guess this check happens at MIR or something)

PR #287, opened 2023-10-13
Introduce core library

W95Psp:
This PR:
 - adds `libs` folder with a partial F* `core` model and a few other libraries (see `libs/fstar`);
 - adds a few examples in `examples`;
 - adds a CI for those examples.


.github/workflows/install_and_test.yml
```diff
@@ -28,6 +28,34 @@ jobs:
...
+
+    - name: Checkout Hacl
+      uses: actions/checkout@v3
+      with:
+        repository: 'hacl-star/hacl-star'
+        path: "hacl-star"
+      
+    - name: Extract and typecheck examples
+      run: |
+        export HACL_HOME="$GITHUB_WORKSPACE/hacl-star"
```
franziskuskiefer, line 43: We should throw these things into helpers. But this is fine for now.

examples/Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[workspace]
+members = [
+     "chacha20",
```
franziskuskiefer, line 3: What about kyber-serialize?

examples/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+| Name     | Status         |
+| -------- | -------------- |
+| chacha20 | Lax-typechecks |
+| lob      | Typechecks     |
+| sha256   | Lax-typechecks |
+
+## How-to extract and typecheck an example
+To generate F* modules and **lax-typecheck** them for an example `<EXAMPLE>`:
+1. `cd <EXAMPLE>/proofs/fstar/extraction`
```
franziskuskiefer, line 11: The `proofs` directory doesn't exist.
Maybe we should add a script here to extract everything?

examples/README.md
```diff
@@ -0,0 +1,13 @@
+# Examples
+
+| Name     | Status         |
+| -------- | -------------- |
+| chacha20 | Lax-typechecks |
+| lob      | Typechecks     |
```
franziskuskiefer, line 6: log fails for me with
```text
* Error 37 at Lob_backend.fst(15,47-15,64):
  - Constructor Candid.Types.Internal.TypeInner_Variant not found
```

examples/README.md
```diff
@@ -0,0 +1,13 @@
...
+| sha256   | Lax-typechecks |
```
franziskuskiefer, line 7: This fails with
```text
* Error 72 at Sha256.fst(103,44-103,49):
  - Identifier not found: [slice]
```

examples/README.md
```diff
@@ -0,0 +1,13 @@
+# Examples
+
+| Name     | Status         |
+| -------- | -------------- |
+| chacha20 | Lax-typechecks |
```
franziskuskiefer, line 5: This fails with the same slice error as sha256.

examples/chacha20/src/lib.rs
```diff
@@ -0,0 +1,124 @@
...
+    b = update_array(b, plain);
+    b = chacha20_encrypt_block(st0, ctr, &b);
+    b[0..plain.len()].to_vec()
+}
+
+pub fn chacha20_update(st0: State, m: &[u8]) -> Vec<u8> {
+    let mut blocks_out = Vec::new();
+    let num_blocks = m.len() / 64;
+    let remainder_len = m.len() % 64;
+    // for (i, msg_block) in m.chunks(64).enumerate() {
```
franziskuskiefer, line 102: Let's drop the commented code.

franziskuskiefer (changes requested):
Thanks, looks like some things aren't working yet, otherwise this looks good.

examples/Cargo.toml
```diff
@@ -0,0 +1,6 @@
...
+     "lob",
+     "sha256",
+]
```
franziskuskiefer, line 6: This should get resolver 2.

examples/README.md
franziskuskiefer: Or just say that `make` runs the extraction as well (which is a little confusing, given that it's run from the extraction dir)

examples/Cargo.toml
W95Psp: Yes, we should include it at some point, but for this PR I just want to introduce `core` + a few examples.
Then I'll make a follow-up PR with more examples (and better CI).

franziskuskiefer (approved):

PR #244, opened 2023-09-04
Attributes: improvements

W95Psp:
This PR:
 - redesign the way Hax-specific attributes are produced and formatted in the proc-macro;
 - simplifies and factorizes things in the engine;
 - makes the following attributes available:
    + `requires`,
    + `ensures`,
    + `decreases`, **(no support in the fstar backend yet)**
    + `include`,
    + `exclude`,
    + `lemma_statement`,
    + `modeled_by`, **(no support in the fstar backend yet)**
    + `refine`.

## Restrictions
For now, `requires` clause offer no support for `&mut` inputs (see  #290)

## How attributes works
The crate `hax_proc_macros` provides user-friendly proc-macro attributes, for instance:
```rust
#[ensures(|result| result == x * 2)]
pub fn twice(x: u64) -> u64 {
    x + x
}
```
This is transformed to:
```rust
const _: () = {
    #[_hax::json("{\"Uid\":{\"uid\":\"dc424364103a40269ad2d703121d49fb\"}}")]
    #[_hax::json("{\"ItemStatus\":{\"Included\":{\"late_skip\":true}}}")]
    #[allow(unused)]
    fn ensures(x: u64, result: u64) -> bool {
        result == x * 2
    }
};
#[_hax::json(
    "{\"AssociatedItem\":{\"role\":\"Ensures\",\"item\":{\"uid\":\"dc424364103a40269ad2d703121d49fb\"}}}"
)]
pub fn twice(x: u64) -> u64 {
    x + x
}
```
Note that here, `#[_hax::json(...)]` is no longer a proc-macro attribute: that's a tool attribute (which are never expanded).
The JSON payload on `_hax::json` is a serde serialization of an `enum` of every possible Hax attribute.

## Example
```rust
#[hax::modeled_by(FStar::IO::debug_print_string)]
pub fn f(line: String) {
    println!("{}", line)
}

#[hax::decreases((m, n))]
pub fn ackermann(m: u64, n: u64) -> u64 {
    match (m, n) {
        (0, _) => n + 1,
        (_, 0) => ackermann(m - 1, 1),
        _ => ackermann(m - 1, ackermann(m, n - 1)),
    }
}

#[hax::lemma_statement]
/// $`\forall n \in \mathbb{N}, \textrm{ackermann}(2, n) = 2 (n + 3) - 3`$
pub fn ackermann_property_m1(n: u64) -> bool {
    ackermann(2, n) == 2 * (n + 3) - 3
}

[...]

PR #293, opened 2023-10-20
Rust crates: harmonize versions, fill missing fields

W95Psp:


franziskuskiefer (commented):
Thanks.

Are we going with 0.1 or 0.0.1?

So we have the following crates

- hax-test-harness (doesn't need to be published)
- hax-lib-macros
- hax-lint
- hax-frontend-exporter-options
- hax-adt-into
- hax-frontend-exporter
- hax-diagnostics
- hax-phase-debug-webapp
- hax-subcommands (binaries)
  - cargo-hax
  - hax-export-json-schemas
  - hax-pretty-print-diagnostics
- hax-cli-options-engine
- hax-cli-options
- hax-driver

Did I miss one?
Let's add all of this to a readme to make sure we know what's there and what needs to be published.

franziskuskiefer (changes requested):
Also, there appear to be changes from #291 in here. Don't they require #244 ?

W95Psp:
I think now things are ready, I bumped the version, added a PUBLISHING.md and fixed a few things

franziskuskiefer (approved):

PR #298, opened 2023-10-20
Fixes for &mut index assignations

W95Psp:
In the expression `x[i] = e`, `x[i]` was handled correctly only if the underlying indexing operation was [IndexMut](https://doc.rust-lang.org/core/ops/trait.IndexMut.html#tymethod.index_mut).
However, if `x` is of type `&mut T`, the trait being used is not [IndexMut](https://doc.rust-lang.org/core/ops/trait.IndexMut.html#tymethod.index_mut) anylonger but [Index](https://doc.rust-lang.org/core/ops/trait.Index.html#tymethod.index)!

This PR basically adds support for Index `x[i]` in LHS of assigns. It also adds a test.

### DIFF
--- a/engine/lib/ast_utils.ml
+++ b/engine/lib/ast_utils.ml
@@ -789,6 +789,11 @@ module Make (F : Features.T) = struct
           let* place = of_expr place in
           wrap @@ FieldProjection { place; projector }
       | App { f = { e = GlobalVar f; _ }; args = [ place; index ] }
+        when Global_ident.eq_name Core__ops__index__Index__index f ->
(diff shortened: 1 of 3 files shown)


franziskuskiefer (approved):
🎉

franziskuskiefer:
@W95Psp It looks like the generated F* code isn't quite right. I get things like
```
Rust_primitives.Hax.update_at (f_tail self <: t_CList v_N_QUEUES v_N_THREADS)
```
where it should probably be `self.f_tail`.

PR #300, opened 2023-10-20
PPX for antiquotations of Rust code in OCaml

W95Psp:
This PR aims at allowing the following:
```ocaml
match (e: SomeAST.expr) with
| [%rust_pat? "assert!(PAT_1_someName)"] -> something where `someName` is bound
...
```

`[%rust_pat? "CODE"]`: CODE is an arbitrary Rust expression. Anything named `PAT_N_<NAME>` will be turned into a pattern variable in OCaml.
Basically, this thing will call Hax itself, output hax's internal AST in JSON, parse that, and generate an OCaml pattern that match any internal hax expressions that looks like CODE.

The logic is mostly working already! Will finish that when I'm back, but that should help a lot. Currently we have horrible hand-written patterns like https://github.com/hacspec/hax/blob/cabd4b5efb7144f2ab2c3b1b242ba65ff382d725/engine/lib/phases/phase_reconstruct_for_loops.ml#L52
We could replace that whole thing with `[%rust_pat? "for _PAT_1_i in _PAT_1_range { PAT_1_body }"]` morally.
This will be much easier to read, to write and much more stable (if the translation changes, such horrible pattern had to be revisited)

### DIFF
--- a/engine/lib/ast.ml
+++ b/engine/lib/ast.ml
@@ -43,13 +43,14 @@ and primitive_ident = Deref | Cast | LogicalOp of logical_op
 module Global_ident = struct
   module T = struct
     type t =
-      [ `Concrete of concrete_ident
-      | `Primitive of primitive_ident
-      | `TupleType of int
-      | `TupleCons of int
-      | `TupleField of int * int
-      | `Projector of [ `Concrete of concrete_ident | `TupleField of int * int ]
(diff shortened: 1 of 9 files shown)


W95Psp:
Closing this PR: it is stale, but we might want to rework that in the future.

PR #302, opened 2023-10-20
fix(tests): fix warnings

mzacho:


franziskuskiefer (approved):
Thanks

W95Psp:
Hi, just replying quickly (seeing some emails, but I'm on vacation this week)
This wrong case in `tests/naming/src/lib.rs` is on purpose, I should have documented that and added `#[warn(non_camel_case_types)]`

tests/naming/src/lib.rs
```diff
@@ -70,7 +70,7 @@ trait T1 {}
...
 impl T1 for (Foo, u8) {}
 
-trait T2_for_a {}
-impl T2_for_a for Arity1<(Foo, u8)> {}
-trait T3_e_for_a {}
-impl T3_e_for_a for Foo {}
+trait T2ForA {}
+impl T2ForA for Arity1<(Foo, u8)> {}
+trait T3EForA {}
+impl T3EForA for Foo {}
```
W95Psp, line 76: Please revert and add `#[warn(non_camel_case_types)]`
This tests the naming translation: while impls have no name in Rust, we need to address them in some backends. An impl `impl Trait for Type` is named `Trait_for_Type` when `Type` is simple enough. Those impls test this naming scheme.

W95Psp (changes requested):

tests/naming/src/lib.rs
mzacho: fixed

tests/naming/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![allow(dead_code)]
+#[warn(non_camel_case_types)]
```
W95Psp, line 2: I was thinking about local attributes, but that would clutter the file, but indeed a module attribute is nicer here. But then I think you need to add a bang:  
```suggestion
#![warn(non_camel_case_types)]
```

(otherwise it's only applying to `Foo`)

PR #312, opened 2023-10-31
fix(hax-lib-macros): drop nightly features

W95Psp:
This PR makes the crate `hax-lib-macros` buildable with non-nightly Rustc


franziskuskiefer (approved):
Thanks!

PR #313, opened 2023-10-31
Add hax cfg when running cargo hax

franziskuskiefer:


franziskuskiefer:
This actually overrides all `RUSTFLAGS` ... So maybe we want to be a little more intelligent 🤔

W95Psp:
Oh, I'm seeing that PR only now, I'm doing something very similar in some local branch :smile:

franziskuskiefer:
Closing in favour of #314

PR #315, opened 2023-10-31
Fix unsizes on mutations

### LINKED ISSUE #256: `copy_from_slice` on a array is not accepted
cc @geonnave

W95Psp:
Fixes #256: `copy_from_slice` (and any other functions with &mut inputs actually) on sized things (e.g. fixed size arrays `[T; N]`) 

This is related to thin / fat pointers.

cc @karthikbhargavan @xvzcf @geonnave

### DIFF
--- a/tests/mut-ref-functionalization/src/lib.rs
+++ b/tests/mut-ref-functionalization/src/lib.rs
@@ -4,6 +4,14 @@ struct S {
     b: [u8; 5],
 }
 
+pub fn foo(mut lhs: S, rhs: &S) -> S {
+    for i in 0..1 {
+        lhs.b[i] += rhs.b[i];
+    }
+
+    lhs
+}
+
 impl S {
     fn update(&mut self, x: u8) {
         self.b[0] = x;
--- a/tests/mut-ref-functionalization/src/lib.rs
+++ b/tests/mut-ref-functionalization/src/lib.rs
@@ -6,6 +6,11 @@ fn index_mutation(x: core::ops::Range<usize>, a: &'static [u8]) {
     v[1] = 3;
 }
 
+fn index_mutation_unsize(mut x: [u8; 12]) -> u8 {
+    x[4..5].copy_from_slice(&[1, 2]);
+    42
+}
+
 fn build_vec() -> Vec<u8> {
     vec![1, 2, 3]
 }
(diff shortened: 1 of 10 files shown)


tests/mut-ref-functionalization/src/lib.rs
```diff
@@ -6,6 +6,11 @@ fn index_mutation(x: core::ops::Range<usize>, a: &'static [u8]) {
     v[1] = 3;
 }
 
+fn index_mutation_unsize(mut x: [u8; 12]) -> u8 {
+    x[4..5].copy_from_slice(&[1, 2]);
+    42
+}
+
```
franziskuskiefer, line 13: Can we add a test like this?
```rust
struct S {
    b: [u8; 5],
}

impl S {
    fn update(&mut self, x: u8) {
        self.b[0] = x;
    }
}
```
I think the output here doesn't look quite right.
```ocaml
let self:t_S = { self with f_b = Rust_primitives.Hax.update_at (f_b self <: t_S)
```

franziskuskiefer (changes requested):
Thanks. Two comments.

I noticed that running tests fails because of `js_of_ocaml`. This is unrelated to this PR (probably introduced in #265), but something we need to get fixed.
Manually installing it fixes the issue
```
File "bin/dune", line 27, characters 28-39:
27 |  (libraries zarith_stubs_js js_of_ocaml lib))
                                 ^^^^^^^^^^^
Error: Library "js_of_ocaml" not found.
```

tests/mut-ref-functionalization/src/lib.rs
W95Psp: Ah, I thought this was fixed.. Though I think I know what the problem is, will push soon

W95Psp:
I fixed the field bug, and added the example you suggested @franziskuskiefer :)

franziskuskiefer:
> I fixed the field bug, and added the example you suggested @franziskuskiefer :)

And you force pushed ... So this will have to wait until tomorrow I'm afraid because I have to start the review from scratch

W95Psp:
Sorry... I'm so used to running `git rebase main`, but after doing it that's a pain to revert :(

franziskuskiefer (approved):
Thanks

franziskuskiefer:
CI is failing

PR #317, opened 2023-11-02
fix(engine): add all `js_of_ocaml` dependencies

W95Psp:
Does that fix the bug of missing `js_of_ocaml` while building the engine @franziskuskiefer?


franziskuskiefer (approved):
Yes this seems to install everything needed.

PR #331, opened 2023-11-09
feat(engine): propagate generics in fun calls

W95Psp:
This is particularly important for generic constants: if a generic constant on a function is not constrained enough (in terms of type), and if we extract it as an implicit parameter, then the target language might not be able to infer it.

Thus, we need those const generics to be explicit.

This PR propagates the (until then implicit) information about generics (constants, but also types and lifetimes). 

Now, I need to either:
 - [x] ~~write a phase that transforms each generic const of functions (definitions and call) into plain values;~~
 - [x] just tweak the backends.

EDIT: tweaking the backends is the way to go, we cannot always convert const-generics to normal terms.


franziskuskiefer (approved):
Thanks looks good.

PR #333, opened 2023-11-09
feat(engine/fstar): `let`s instead of typing lambdas

W95Psp:
We used to have closure like `|x: T, y: U| body` extract to `fun (x:
T) (y: U) -> body` in the F* backend.

When such a typed lambda is consumed by a higer-order function, we sometimes face issues with unification in F*.

This commit introduce a trick that allows for better unification in F*: we now generate a let binding per input in lambdas. For example, `|x: T, y: U| body` now extracts to:
```fstar
fun x y ->
    let x: T = x in
    let y: T = y in
    body
```

### DIFF
--- a/engine/backends/fstar/fstar_backend.ml
+++ b/engine/backends/fstar/fstar_backend.ml
@@ -496,10 +496,28 @@ struct
           (F.term @@ F.AST.Name (pglobal_ident e.span constructor))
           [ r ]
     | Closure { params; body } ->
-        let ppat_ascribed pat =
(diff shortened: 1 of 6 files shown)


franziskuskiefer (changes requested):
looks like CI is failing
cancelling the review request for now

franziskuskiefer:
@W95Psp the chacha patch doesn't apply cleanly anymore. You have to update that.

W95Psp:
Yeah, you're right :( gonna do that.

franziskuskiefer (approved):
:shipit:

PR #334, opened 2023-11-09
feat(hax-lib): intro hax-lib

W95Psp:
This commit adds a crate named `hax-lib`.

    This crate contains hax-specific helpers for Rust programs. Those
    helpers are usually no-ops when compiled normally but meaningful when
    compiled under hax.

    For now, essentially, this adds an assert and an assume macro.

### DIFF
--- a/hax-lib/Cargo.toml
+++ b/hax-lib/Cargo.toml
@@ -6,7 +6,8 @@ license.workspace = true
 homepage.workspace = true
 edition.workspace = true
 repository.workspace = true
-readme.workspace = true
+readme = "README.md"
+description = "Hax-specific helpers for Rust programs"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
--- a/hax-lib/Cargo.toml
+++ b/hax-lib/Cargo.toml
@@ -1,0 +1,13 @@
+[package]
+name = "hax-lib"
+version.workspace = true
+authors.workspace = true
+license.workspace = true
+homepage.workspace = true
+edition.workspace = true
+repository.workspace = true
+readme.workspace = true
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
--- a/hax-lib/src/lib.rs
+++ b/hax-lib/src/lib.rs
@@ -1,6 +1,17 @@
 //! Hax-specific helpers for Rust programs. Those helpers are usually
 //! no-ops when compiled normally but meaningful when compiled under
 //! hax.
+//!
+//! # Example:
+//!
+//! ```rust
+//! fn sum(x: Vec<u32>, y: Vec<u32>) -> Vec<u32> {
+//!   hax_lib::assume!(x.len() == y.len());
+//!   hax_lib::assert!(hax_lib::forall(|i: usize| hax_lib::implies(i < x.len(), || x[i] < 4242)));
+//!   hax_lib::debug_assert!(hax_lib::exists(|i: usize| hax_lib::implies(i < x.len(), || x[i] > 123)));
+//!   x.into_iter().zip(y.into_iter()).map(|(x, y)| x + y).collect()
+//! }
+//! ```
 
 #[doc(hidden)]
 #[cfg(hax)]
@@ -103,6 +114,6 @@ pub fn exists<T>(_f: impl Fn(T) -> bool) -> bool {
 }
 
 /// The logical implication `a ==> b`.
-pub fn implies(lhs: bool, rhs: bool) -> bool {
-    !lhs || rhs
+pub fn implies(lhs: bool, rhs: impl Fn() -> bool) -> bool {
+    !lhs || rhs()
 }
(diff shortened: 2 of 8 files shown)


hax-lib/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "hax-lib"
+version.workspace = true
+authors.workspace = true
+license.workspace = true
+homepage.workspace = true
+edition.workspace = true
+repository.workspace = true
+readme.workspace = true
```
franziskuskiefer, line 9: Please add a `description` as well.

franziskuskiefer, line 9: And add a readme that says what this is doing.
In this case the `readme.workspace` will throw warnings because it won't use this one.

tests/generics/src/lib.rs
```diff
@@ -1,41 +1,45 @@
...
 }
 
-fn foo<const LEN: usize>(arr: [usize; LEN]) -> usize {
-    let mut acc = LEN + 9;
-    for i in 0..LEN {
-        acc += arr[i];
-    }
-    acc
-}
+// fn dup<T: Clone>(x: T) -> (T, T) {
```
franziskuskiefer, line 7: What's with all the commented code here?

hax-lib/src/lib.rs
```diff
@@ -0,0 +1,52 @@
...
+
+#[macro_export]
+macro_rules! assert {
+    ($($arg:tt)*) => {
+        $crate::proxy_macro_if_not_hax!(::std::assert, $crate::assert, $($arg)*)
+    };
+}
+
+#[cfg(hax)]
+pub fn assert(_formula: bool) {}
```
franziskuskiefer, line 34: How are these supposed to be used? There's no version for these two function when `cfg(not(hax))`

hax-lib/src/lib.rs
```diff
@@ -0,0 +1,52 @@
+#[cfg(hax)]
```
franziskuskiefer, line 1: Please add documentation on top here that explains how this is being used.

franziskuskiefer (changes requested):
Thanks! Mostly comments, but let's do another round.

tests/generics/src/lib.rs
W95Psp: Hm, that was a quick test I committed without noticing :roll_eyes: 
Thanks for catching that!

W95Psp:
Thanks for the review :smiley: 
I've fixed all the points you mentioned (I think!).
If those changes are OK for you, I'll rebase things, the CI is broken on the PR for the other branch `type-closure-fstar` :(

W95Psp:
CI is going through in the other PR :partying_face:

hax-lib/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+#[macro_export]
+/// Proxy to `std::assert!`. Compiled with `hax`, this is transformed
+/// into a `assert` in the backend.
+macro_rules! assert {
+    ($($arg:tt)*) => {
+        $crate::proxy_macro_if_not_hax!(::std::assert, $crate::assert, $($arg)*)
+    };
+}
+
+#[cfg(hax)]
```
franziskuskiefer, line 41: Couldn't we inline the functions into the macros then instead of having dead functions here? (See suggestion above)
And if we have them here, we should hide them from the documentation if no one should use them.

Same for assume

hax-lib/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+        $crate::proxy_macro_if_not_hax!(::std::debug_assert, no, $($arg)*)
+    };
+}
+
+#[macro_export]
+/// Proxy to `std::assert!`. Compiled with `hax`, this is transformed
+/// into a `assert` in the backend.
+macro_rules! assert {
+    ($($arg:tt)*) => {
+        $crate::proxy_macro_if_not_hax!(::std::assert, $crate::assert, $($arg)*)
```
franziskuskiefer, line 37:
```suggestion
        $crate::proxy_macro_if_not_hax!(::std::assert, |_formula: bool| {} $($arg)*)
```

hax-lib/src/lib.rs
```diff
@@ -0,0 +1,104 @@
+//! Hax-specific helpers for Rust programs. Those helpers are usually
+//! no-ops when compiled normally but meaningful when compiled under
+//! hax.
```
franziskuskiefer, line 3: Let's add examples (and thus tests) for everything.
Either here or at the macros.

```suggestion
//! hax.
//!
//! # Example:
//!
//! ```rust
//! fn sum(x: u32, y: u32) -> u32 {
//!   hax_lib::assert!(x < 4242 && y < 424242);
//!   hax_lib::debug_assert!(x < 4242 && y < 424242);
//!   x + y
//! }
//! ```
```

franziskuskiefer (approved):
Approving, but please address the comments before merging.

hax-lib/src/lib.rs
W95Psp: No, we cannot inline those:
 - if we replace that with a closure, then we cannot translate to the right F* definition;
 - if we inline a `fn` item, then it won't be located in the right place.

I'm going to just hide those from documentation then!

PR #341, opened 2023-11-13
feat: opts for F* backend (`z3rlimit`, `fuel`, `ifuel`)

W95Psp:
This PR adds options for the F* backend to control F*'s `z3rlimit`, `fuel` and `ifuel`.

It also add an `open FStar.Mul` in the header of extracted modules, which is often convenient when playing with extracted modules.

Last but not least: this PR adds a newline at the end of the file 🎉🥳✨ (this always hurts my eyes in diffs)


cli/options/src/lib.rs
```diff
@@ -113,10 +113,23 @@ impl NormalizePaths for PathOrDash {
     }
 }
 
+#[derive(JsonSchema, Parser, Debug, Clone, Serialize, Deserialize)]
+pub struct FStarOptions {
```
franziskuskiefer, line 117: We don't need to add a full test yet, but can you add an example in the documentation at least so people know how to use it?

franziskuskiefer (approved):

PR #344, opened 2023-11-13
feat: `nix`ify examples & CI

W95Psp:
This PR makes the examples extract and typecheck via Nix.
This also gives (Nix) users a one-liner (`nix develop .#fstar-examples`) to extract & typecheck the F* examples with the correct F* and Hacl* versions.

This PR also uses Nix to run the examples.


examples/Cargo.toml
```diff
@@ -5,3 +5,6 @@ members = [
      "sha256",
 ]
 resolver = "2"
+
+[workspace.dependencies]
+hax-lib-macros = { path = "../hax-lib-macros" }
```
franziskuskiefer, line 10:
```suggestion
hax-lib-macros = { path = "../hax-lib-macros" }

```

examples/README.md
```diff
@@ -8,6 +8,20 @@
...
+<details>
+  <summary><b>Requirements</b></summary>
+  
+  First, make sure to have hax installed in PATH. Then:
+  
+  * With Nix, `nix develop .#fstar-examples` setups a shell automatically for you.
+     
+  * Without Nix:
+    1. install F* `v2023.09.03` manually (see https://github.com/FStarLang/FStar/blob/master/INSTALL.md);
+    2. clone [Hacl*]() somewhere;
```
franziskuskiefer, line 20: Missing link?

examples/default.nix
```diff
@@ -0,0 +1,42 @@
...
+    // {
+      inherit cargoArtifacts;
+      pname = "hax-examples";
+      doCheck = false;
+      buildPhaseCargoCommand = ''
+        cd examples
+        export CACHE_DIR=$(mktemp -d)
+        export HINT_DIR=$(mktemp -d)
+        export SHELL=${stdenv.shell}
+        make
```
franziskuskiefer, line 37: You have to clean first, otherwise this isn't doing much intereting.

examples/sha256/Cargo.toml
```diff
@@ -8,4 +8,4 @@ edition = "2021"
 path = "src/sha256.rs"
 
 [dependencies]
-hax-lib-macros = { git = "https://github.com/hacspec/hax.git", branch = "attributes" }
+hax-lib-macros.workspace = true
```
franziskuskiefer, line 11:
```suggestion
hax-lib-macros.workspace = true

```

franziskuskiefer (commented):
Are there still regular tests? Nix can always only be in addition to the regular workflows as they must always work.

examples/default.nix
W95Psp: Well, actually the sources are filtered above, so that the build is independent from generated F* or Coq (or whatever) files.
I added a comment about that, but also a `make clean` (which should be a no-op, but that doesn't hurt)

W95Psp:
Regarding regular tests, previous ones were also using Nix, just in a bad way.
So, we might want to test that with a Opam installed F* in the long tests (those that run only when in the merge queue).
What do you think?

examples/README.md
```diff
@@ -17,8 +17,9 @@
      
   * Without Nix:
     1. install F* `v2023.09.03` manually (see https://github.com/FStarLang/FStar/blob/master/INSTALL.md);
-    2. clone [Hacl*]() somewhere;
-    3. `export HACL_HOME=THE_DIRECTORY_WHERE_YOU_HAVE_HACL_STAR`;
+       1. make sure to have `fstar.exe` in PATH;
+       2. or set the `FSTAR_HOME` environment variable.
+    2. clone [Hacl*](https://github.com/hacl-star/hacl-star/pull) somewhere;
```
franziskuskiefer, line 22:
```suggestion
    2. clone [Hacl*](https://github.com/hacl-star/hacl-star) somewhere;
```

franziskuskiefer (approved):
> Regarding regular tests, previous ones were also using Nix, just in a bad way.
So, we might want to test that with a Opam installed F* in the long tests (those that run only when in the merge queue).
What do you think?

Ok, let's file an issue to add CI for this. Which one's run on the PR and which on the merge queue, I don't mind.

PR #345, opened 2023-11-13
corelib: update F* model

W95Psp:
This PR contains all the edits we (@karthikbhargavan, @xvzcf, and I) did for Kyber.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ result
 .DS_Store
 .depend
 .cache
+proof-libs/fstar/rust_primitives/#*#
 **/proofs/*/extraction/*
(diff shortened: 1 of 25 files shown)


.gitignore
```diff
@@ -10,6 +10,7 @@ result
 .DS_Store
 .depend
 .cache
+proof-libs/fstar/rust_primitives/#*#
```
franziskuskiefer, line 13: Why do we ignore these?

franziskuskiefer (approved):
lgtm with one question

.gitignore
xvzcf: These are "autosave" files generated by Emacs

PR #347, opened 2023-11-14
Smallest possible ProVerif backend

jschneider-bensch:
This includes the reserved words for ProVerif, adds some (for now) pure boilerplate at the beginning and end, and does not much more than is necessary to translate the result of `cargo new --lib mylib` to valid ProVerif with the Rust `fn` translated to an empty `letfun` in ProVerif.

### DIFF
--- a/cli/options/src/lib.rs
+++ b/cli/options/src/lib.rs
@@ -119,9 +119,9 @@ pub enum Backend {
     Fstar,
     /// Use the Coq backend
     Coq,
-    /// Use the EasyCrypt backend
+    /// Use the EasyCrypt backend (warning: work in progress!)
     Easycrypt,
-    /// Use the ProVerif backend
+    /// Use the ProVerif backend (warning: work in progress!)
     ProVerif,
 }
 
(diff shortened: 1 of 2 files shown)


W95Psp (approved):
Thank you! :)
That looks good to me, though, should we merge the backend yet, or wait to have something more usable?
I don't have strong opinion about this (we already have the EC backend in those lines).
Maybe we can add it already, but make it as "in progress" or hide it in the CLI. (we'd just need to change `cli/options/src/lib.rs`)

Wdyt @franziskuskiefer?

franziskuskiefer:
> Maybe we can add it already, but make it as "in progress" or hide it in the CLI. (we'd just need to change `cli/options/src/lib.rs`)
> 
> Wdyt @franziskuskiefer?

I think we should get it in. Not all backends are fully working right now and that's ok. Let's add warnings to EC and PV that they are work in progress. But it's useful to do this iteratively.

karthikbhargavan:
This looks good to me and a useful (and exciting) beginning. Let's merge it and keep updating it.
We may also want to introduce a minimal Rust library for protocol actions (maybe in the next PR).

PR #349, opened 2023-11-14
Ssprove backend base

cmester0:
The SSProve backend, waiting on #348, ~~#357~~ ~~and #362~~


W95Psp:
Looks like you need to `cargo test --test toolchain` and then  `cargo insta review` :)

cmester0:
> Looks like you need to `cargo test --test toolchain` and then `cargo insta review` :)

Seems I am getting different results when running the tests locally (instead of with nix). However I think the issue is that I removed reordering for Coq and SSProve, as they put some functions in an incorrect order, while the default order works.

cmester0:
This should be ready for merging. @W95Psp

engine/lib/phases/phase_state_passing_loop.ml
```diff
@@ -0,0 +1,378 @@
+(* TODO: handle Exn report *)
```
W95Psp, line 1: Mmh, this is basically a copy-paste of `phase_local_mutation` with a few differences, right?
We need to either make `phase_local_mutation` parametric or split it in two.

I recall we discussed that phase at some point.
If I'm not mistaking, you want to allow local mutation, but at the boundaries of functions and loops? i.e. you want to rewrite mutations of outer variables within a loop?

Currently, I think this `phase_state_passing_loop` is still rewriting all local mutation into shadowing bindings though! Or I'm mistaking?

engine/bin/lib.ml
```diff
@@ -43,7 +43,13 @@ let import_thir_items (include_clauses : Types.inclusion_clause list)
...
       items
   in
-  let items = Deps.sort items in
+  (* TODO: should keep order for non-dependent functions by default? This breaks ssprove extraction. Alternatively this should be a phase *)
+  let items =
+    Deps.sort items
+    (* match options.backend.backend with *)
+    (* | Coq | Ssprove -> items *)
+    (* | _ -> Deps.sort items *)
+  in
```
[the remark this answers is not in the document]
cmester0: Hmm, seems updates fixed the ordering issue for the generated SSProve code. Should I file an issue about making this into a phase?

engine/lib/phases/phase_state_passing_loop.ml
cmester0: Yes, I did copy paste this from `phases_local_mutation`. I want to keep mutable variables (to some extent), the only changes I did was keeping the mutability, as I do not see why it was removed by this phase, however I could refactor it into two phases instead, and just keep the first for the SSProve backend.

cmester0: I might want to allow more mutability, as SSProve can actually reason about it. However, for now I just want to keep the mutable assignments throughout the translations, this is still translated to "shadowing" mutable let binding, but this is an assignment to a memory location by name, so it has the same effect.

cmester0:
@W95Psp we should get this in, we can always do follow up PRs for testing and generalization.

W95Psp (approved):
OK, let's merge then, it has been waiting for too long anyway!

PR #351, opened 2023-11-14
Minimal example test for ProVerif backend

jschneider-bensch:
Include a test in the toolchain test harness that checks whether extraction works for the output of  `cargo new --lib proverif-minimal`.

### DIFF
--- a/tests/Cargo.lock
+++ b/tests/Cargo.lock
@@ -266,6 +266,10 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


jschneider-bensch:
Ready for review now that I could test locally that the test is correctly picked up by the harness.

W95Psp (approved):
LGTM, let's merge that now in `proverif-base` (before we merge #355 -- in `main` that time :sweat_smile:)

PR #352, opened 2023-11-14
chacha20 example without diff

franziskuskiefer:
Move the assumes into the Rust code and udpate the example F* code.


examples/Makefile
```diff
@@ -1,5 +1,8 @@
 .PHONY: default
 default:
+	make -C limited-order-book clean
 	make -C limited-order-book
+	make -C chacha20 clean
 	make -C chacha20
+	OTHERFLAGS="--lax" make -C sha256 clean
 	OTHERFLAGS="--lax" make -C sha256
```
W95Psp, line 8: Ah, I've just added a `clean` target in #344.

W95Psp (approved):

examples/Makefile
franziskuskiefer: heh, ok that's cleaner. Undoing this one.

PR #354, opened 2023-11-15
feat(test-harness): read DUNEJOBS (fixes #350)

### LINKED ISSUE #350: Feature Request: Option to restrict number of dune jobs in test harness
My machine suffers a weird and terrible curse, which causes it to go OOM and kill random programs whenever `dune` runs with more than one job for an extended period of time.
The test harness is the latest place where I found this to be an issue (for me). I have set `DUNEJOBS=1` globally, but it seems to not be respected in the harness when building the engine.

W95Psp:


jschneider-bensch (approved):
Thank you! This does the (one) job for me. :slightly_smiling_face:

PR #355, opened 2023-11-15
Proverif base

W95Psp:
@jschneider-bensch I just noticed now #347 was a PR against `proverif-base`, not against `main`!
Let's merge that into `main` anyway :) (as discussed in #347)


cli/options/src/lib.rs
```diff
@@ -132,8 +132,10 @@ pub enum Backend {
     Fstar(FStarOptions),
     /// Use the Coq backend
     Coq,
-    /// Use the EasyCrypt backend
+    /// Use the EasyCrypt backend (warning: work in progress!)
     Easycrypt,
+    /// Use the ProVerif backend (warning: work in progress!)
```
jschneider-bensch, line 137: I'll take care of putting in a warning in the CLI output when `into pro-verif` is invoked.

jschneider-bensch (approved):
Thanks for merging! Looks good from my side :+1:

cli/options/src/lib.rs
W95Psp: Ok, great!

Issue #168, opened 2023-06-29
Add phase: elim non-struct records on variants into struct records


W95Psp:
Some backends don't support the following:
```rust
enum Path::Foo {
  Path::Foo::Variant1 {f1: T1, ..., fN: TN},
  ...
}
```
(note paths on enum names or variant names is not valid Rust, it's just for illustration)

We should have a phase that (optionally) convert those to:
```rust
struct Path::Foo::Variant1::Payload { f1: T1, ..., fN: TN }
enum Path::Foo {
  Path::Foo::Variant1(Path::Foo::Variant1::Payload),
  ...
}
```

To make this transformation to produce an AST without record-variants, we should add a feature, say `record_payload`, turned on by default in `Features.Rust`. Then, the field `is_record` of the record `Ast.variant` should not be a `bool` any longer, but a `F.record_payload option` (so that we can [refute that pattern](https://v2.ocaml.org/releases/4.12/htmlman/gadts.html#p:gadt-refutation-cases)).

I think this should be worked on based on `inherent_impls` branch, to enjoy record-related improvements. Btw, we'll need an `append_type_name` in `concrete_ident.ml{i,}` to construct `Path::Foo::Variant1::Payload` from `Path::Foo::Variant1`.

@cmester0, what do you think about this plan?

cmester0:
This does seem like a good way to do things. It is way better than doing it in the each backend.

franziskuskiefer:
@cmester0 is this still needed? If not, please close it.

franziskuskiefer:
> [@cmester0](https://github.com/cmester0) is this still needed? If not, please close it.

@cmester0 👆🏻

maximebuyse:
This seems like it would benefit only the Rocq backend, so not planned for now.

PR #357, opened 2023-11-15
Adding phase to simplify enum and record matching statements

cmester0:
Implementation of #168 , #180 closed and moved here instead.


engine/lib/ast.ml
```diff
@@ -271,13 +271,16 @@ functor
...
     and pat' =
       | PWild
       | PAscription of { typ : ty; typ_span : span; pat : pat }
       | PConstruct of {
           name : global_ident;
           args : field_pat list;
-          is_record : bool; (* are fields named? *)
+          is_record : F.project_instead_of_match option;
+          (* are fields named? *)
+          (* F.project_instead_of_match *)
```
W95Psp, line 283: cleanup?

engine/lib/phases/phase_project_instead_of_match.ml
```diff
@@ -0,0 +1,227 @@
...
+              captures = List.map ~f:dexpr captures;
+            }
+
+    and darm' (_span : span) (a : A.arm') : B.arm' =
+      let simple_pat, remaining_pats = project_pat a.arm_pat in
+      {
+        arm_pat = simple_pat;
+        body = lets_of_pat_bindings remaining_pats (dexpr a.body);
+      }
+      [@@inline_ands bindings_of dexpr]
```
W95Psp, line 220: Mmh, I'm a bit worried about that one.
If I understand correctly, here something like:
```rust
match x {
  Some(Foo {a: None, b}) => b,
  Some(Foo {a: Some(a), b}) => a + b,
  None => 0,
}
```
would be rewritten into:
```rust
match x {
  Some(some_name) =>
    match some_name {
      Foo {a: some_name, b} => match some_name { None => b }
    },
  Some(some_name) => 
     match some_name {
       Foo {a: some_name, b} => match some_name { Some(a) => a + b},
     },
  None => 0,
}
```
am I right?

But then, the first sub-sub-match is incomplete and the second arm won't be reached.

engine/lib/phases/phase_project_instead_of_match.ml
```diff
@@ -0,0 +1,227 @@
...
+        B.pat' * (B.pat * B.expr) list =
+      match pat with
+      | PWild -> (PWild, [])
+      | PAscription { typ; typ_span; pat } ->
+          let simple_pat, remaining_pats = project_pat pat in
+          ( PAscription { typ = dty span typ; pat = simple_pat; typ_span },
+            remaining_pats )
+      | PConstruct { name; args; is_record = Some _; is_struct } ->
+          let update_args = List.map ~f:(project_field_pat span) args in
+          let new_id = UA.fresh_local_ident_in_expr [] "some_name" in
```
W95Psp, line 65: `fresh_local_ident_in_expr` gives you fresh things only relative to a given list of expressions.
So here, you have no guarantees.
When a pattern has multiple sub-patterns, this is an issue, i.e. `match x {Foo {a: PAT_A, b: PAT_B} | ...}`. Currently `project_pat'` would generate two `some_name` bindings.

I guess one solution would be `project_pat'` to take an index or something.

engine/lib/phases/phase_project_instead_of_match.ml
```diff
@@ -0,0 +1,227 @@
...
+                                subpat = None;
+                              };
+                          span;
+                          typ = TApp { ident = name; args = [] };
+                        };
+                    };
+                  ];
+                is_record = (None : FB.project_instead_of_match option);
+                is_struct;
+              },
```
W95Psp, line 92: Should that actually be a `PConstruct`? Here we rewrite `Foo {a: ..., b: ..., c: ...}` into `Foo {Foo: some_name}`, that's wrong right?

W95Psp: I feel like this should be rather be a `PBinding`

W95Psp (changes requested):
Thanks!
I have some comments, I think some edge cases won't be handled correctly.

Can you add a test for this phase also?

engine/lib/phases/phase_project_instead_of_match.ml
cmester0: Oh yes, I forgot to add the expressions here. I should probably make sure the variable is fresh compared to the expression it is being used in.

engine/lib/phases/phase_project_instead_of_match.ml
cmester0: Currently a similar problem exists for
```rust
#![allow(dead_code)]
pub struct Foo {
    pub a : Option<usize>,
    pub b : usize,
}

pub fn bar(x : Option<Foo>) -> usize {
    match x {
        Some(f) if match f { Foo {a: None, b} => true, _ => false } => {
            let Foo {a: None, b} = f;
            b
        }
        Some(f) if match f { Foo {a: Some(a), b} => true, _ => false } => {
            let Foo {a: Some(a), b} = f;
            a + b
        }
        None => 0,
    }
}
```
which is the correct "dumb" way to do the translation. However I should fix this, such that multiple cases for the same constructor are respected (which is currently not the case).

cmester0: This is indeed an issue, I should have some check about which case I end up in before binding the values.

cmester0: I would expect the translation to look something like:
```rust
match x {
  Some(f) if if let (None, b) = (f.a, f.b) { true } else { false } => {
    let None = f.a;
    let b = f.b;
    b
  }
  None => 0,
  Some(f) if if let (Some(a), b) = (f.a, f.b) { true } else { false } => {
    let Some(a) = f.a;
    let b = f.b;
    a + b
  },
}
```
However we currently do not translate the conditions on match cases?

cmester0:
I am having a hard time actually fixing the issues, without the use of conditional match statements, which are currently not supported. This should probably fall under the same or a very similar phase to removing conditional matching as most of the issues are the same. For now I use very few or none of the features, however disables any matching on structs for the ssprove backend.

cmester0:
This was more difficult than expected, and not too relevant yet, thus we will close this for now.

Issue #358, opened 2023-11-15
Warn about WIP backend when `into pro-verif` is invoked.


jschneider-bensch:

W95Psp:
fixed by #359 :partying_face:

PR #359, opened 2023-11-15
Warn on CLI about experimental backends

jschneider-bensch:
This will emit a warning message when using the EasyCrypt or ProVerif backends, addressing #358.
Hope I found the appropriate place for the warning code!

### DIFF
--- a/cli/driver/src/exporter.rs
+++ b/cli/driver/src/exporter.rs
@@ -1,4 +1,4 @@
-use hax_cli_options::{PathOrDash, ENV_VAR_OPTIONS_FRONTEND};
+use hax_cli_options::{Backend, PathOrDash, ENV_VAR_OPTIONS_FRONTEND};
 use hax_frontend_exporter;
 use hax_frontend_exporter::state::{ExportedSpans, LocalContextS};
 use hax_frontend_exporter::SInto;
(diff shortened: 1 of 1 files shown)


cli/driver/src/exporter.rs
```diff
@@ -390,6 +390,16 @@ impl Callbacks for ExtractionCallbacks {
...
+                    match backend.backend {
+                        hax_cli_options::Backend::Easycrypt
+                        | hax_cli_options::Backend::ProVerif => {
+                            eprintln!(
+                                "⚠️ Warning: Experimental backend \"{}\" is work in progress.",
+                                backend.backend
+                            )
+                        }
+                        _ => (),
+                    };
```
W95Psp, line 402: Small (purely syntactical) nit: you could write that with the [`matches` macro](https://doc.rust-lang.org/std/macro.matches.html)
```rust
use hax_cli_options::Backend;
if matches!(backend.backend, Backend::Easycrypt | Backend::ProVerif) {
  ...
}

W95Psp (approved):
LGTM, just a small nit (but feel free to ignore it, it's quite subjective)

cli/driver/src/exporter.rs
jschneider-bensch: Thanks for pointing that out! That's much more concise. :+1:

W95Psp: :smiley:

PR #361, opened 2023-11-15
Docs (`rustdoc`, `odoc`): build and publish to gh-pages (via CI)

W95Psp:
This PR adds build for `rustdoc` and `odoc` (ocaml's doc tool), and bundles documentation in one `index.html` ToC.
It also adds CI for deploying to GH Pages


engine/doc/dune
```diff
@@ -0,0 +1,3 @@
+(documentation
+  (package hax-engine)
+  (mld_files index))
```
franziskuskiefer, line 3:
```suggestion
  (mld_files index))

```

engine/doc/index.mld
```diff
@@ -0,0 +1,21 @@
...
+    {- utilities and PPXs:
+       {ul {- {!module-Hacspeclib_macro_parser}}
+           {- {!module-Ppx_functor_application}}
+           {- {!module-Ppx_generate_features}}
+           {- {!module-Ppx_inline}}
+        }
+    }
+}
+
+
```
franziskuskiefer, line 21:
```suggestion
```

.github/workflows/gh_pages.yml
```diff
@@ -0,0 +1,24 @@
+name: Deploy documentation to GH Pages
+
+on:
+  pull_request:
+  merge_group:
```
franziskuskiefer, line 5: Let's only do this on main and on dispatch.

.github/workflows/gh_pages.yml
```diff
@@ -0,0 +1,24 @@
...
+    if: ${{ github.event_name == 'workflow_dispatch'  || github.event_name == 'merge_group'  }}
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: DeterminateSystems/nix-installer-action@main
+    - uses: DeterminateSystems/magic-nix-cache-action@main
+    - name: Build docs
+      run: nix build .#hax.docs --out-link ./built-docs
+    - name: Deploy to gh-pages
+      uses: peaceiris/actions-gh-pages@v3
```
franziskuskiefer, line 21: Let's deploy by action. Then we don't need to depend on third-party actions.
See for example https://github.com/cryspen/hacl-packages/blob/main/.github/workflows/gh-pages.yml#L142-L149
(this also needs to change of the page deployment setting to action)

franziskuskiefer (changes requested):
Generally lgtm with a few nits.

.github/workflows/gh_pages.yml
W95Psp: Makes sense, thanks!

W95Psp:
Thanks, I didn't know about those deploy actions to GH Pages, they look nice! :)

franziskuskiefer (approved):
Thanks

PR #362, opened 2023-11-15
Add IIConst and TIConst

cmester0:


cmester0:
We need to double check if we actually need constants for anything. This implementations just solves an issue that happens because of a bug. In `imported_thir.ml`, in the line
```ocaml
  | NamedConst { def_id = id; _ } -> GlobalVar (def_id Value id)
```
the `Value` should be `AssociatedItem Value`.

cmester0:
Should be resolved by #366 , closing this for now.

PR #368, opened 2023-11-17
feat(engine): local id disambiguation

W95Psp:
This PR renames local identifiers if name clash occurs, for instance:
```rust
fn debug(label: &str, value: u32) {
        println!("[{}] a={}", label, value)
    }

    macro_rules! hello {
        ($label:expr, $value:expr, $($e:tt)*) => {
            let a = $value;
            $($e)*
            debug($label, a)
        };
    }

    fn f() {
        hello!("1", 104,
               hello!("2", 205,
                      hello!("3", 306, let a = 123;);
               );
        );
        debug("last", a)
    }
``` 

This issue is not occurring only when using macros; it also occurs because of rustc internal desugaring before THIR.


tests/naming/src/lib.rs
```diff
@@ -105,3 +105,26 @@ trait FooTrait {
 fn constants<T: FooTrait>() -> usize {
     <T as FooTrait>::ASSOCIATED_CONSTANT + INHERENT_CONSTANT
 }
+
+mod ambiguous_names {
+    fn debug(label: &str, value: u32) {
+        println!("[{}] a={}", label, value)
+    }
+
+    macro_rules! hello {
```
franziskuskiefer, line 114: I thought the macro is gone by thir anyway? So does this really test something?

engine/lib/print_rust.ml
```diff
@@ -560,7 +560,7 @@ let rustfmt_annotated' (x : AnnotatedString.t) : AnnotatedString.t =
             (original, (Span.dummy (), s))
           else (
             Stdio.prerr_endline @@ "\n##### RUSTFMT TOKEN ERROR #####";
-            Stdio.prerr_endline @@ "prev=" ^ [%show: AnnotatedString.t] prev;
+            (* Stdio.prerr_endline @@ "prev=" ^ [%show: AnnotatedString.t] prev; *)
```
franziskuskiefer, line 563: Should this be removed?

franziskuskiefer (approved):
Approving, not sure about the test though.

tests/naming/src/lib.rs
W95Psp: Yes macros are expanded before THIR, but here what I want to test is hax's renaming for local variables that have same name but different internal ids. A simple way to do this is to exploit the hygienic aspect of Rust macros :)
I rewrote a bit the example so that it's more understandable

PR #369, opened 2023-11-17
Monomorphize update at

W95Psp:
@karthikbhargavan


franziskuskiefer (approved):
This solves the issue on libcrux. Let's get it in

Issue #370, opened 2023-11-18
Impl error on type in trait.


cmester0:
The following:
```rust
pub trait Foo {
  type bar;
}
```
results in the error
```bash
Fatal error: exception Failure("impl_expr_atom: Todo implsource::param \n\nBinder(\n    <Self as Group>,\n    [],\n)")
```

cmester0:
This seems to be an issue with `predicates` being empty in the `ImplSource::Param` branch of `frontend/exporter/src/traits.rs`.

W95Psp:
Hi, yes I got that error at some point as well :(
This issue is caused by some constraint being of the shape `<Self as Foo>::bar : SomeTrait`: `Self` here acts as a local marker, so that we can refer to the associated type `bar`.
I guess your fix in PR #371 was to get rid of the error?
We need a `Self` keyword in sourceimpl I think. I can push a PR!

PR #371, opened 2023-11-20
allow trait param with empty predicates

cmester0:
Attempt at solving #370

### DIFF
--- a/frontend/exporter/src/traits.rs
+++ b/frontend/exporter/src/traits.rs
@@ -253,8 +253,17 @@ impl<'tcx> IntoImplExpr<'tcx> for rustc_middle::ty::PolyTraitRef<'tcx> {
                         })
                         .map(|path| (predicate, path))
                 }) else {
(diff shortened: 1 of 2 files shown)


frontend/exporter/src/traits.rs
```diff
@@ -253,8 +253,17 @@ impl<'tcx> IntoImplExpr<'tcx> for rustc_middle::ty::PolyTraitRef<'tcx> {
...
+                        if predicates.is_empty() {
+                            ImplExprAtom::LocalBound {
+                                clause_id: rand::random(), // TODO
+                                path: Vec::<search_clause::PathChunk>::new().sinto(s)
+                            }
+                        }
+                        else {
+                            ImplExprAtom::Todo(format!("implsource::param \n\n{:#?}", self))
+                        }
+                        .with_args(impl_exprs(s, &nested))
```
W95Psp, line 266: That removes the error, but a `LocalBound` is not what we need (and the `random` will not help :smiley:)

PR #372, opened 2023-11-21
Add `Self` implsource

### LINKED ISSUE #20: Traits: compute explicit implementation expressions
Consider the following code:
```rust
trait Show {
    fn show(self) -> String;
}
// Let's name this impl. `ShowStr`
impl Show for &str {
    fn show(self) -> String {
        self.to_string()
    }
}
// Let's name this impl. `ShowU8`
impl Show for u8 {
    fn show(self) -> String {
        format!("{}", self)
    }
}
// Let's name this impl. `ShowVec`
impl<T: Show, U: Show> Show for (T, U) {
    fn show(self) -> String {
        format!("({}, {})", self.0.show(), self.1.show())
    }
}
```

The expression `(1, "a").show()` is for now translated as a call to the method `show` of the typeclass/trait `Show` with argument `(1, "a")`.
Some backends might have no typeclasses or traits, and some (like F*) have typeclasses mechanisms whose inference mechanism is not great.

Thus, we need the frontend to derive the implementation expressions (i.e. `ShowVec ShowU8 ShowStr` for `(1, "a").show()`).

-----------------
**Edit:** we've got another constraint.  
In the engine, we often transform some patterns (i.e. mutation, loops) into function calls that are trait methods.
Thus, maybe we should *not* do impl resolution in Rust, but on the OCaml side. Or do both: we should anyway export as much info in the frontend, but we'll need to do some impl resolution in the engine.

----------------
Status: @sonmarcho has some great prototype for that :tada: 

----------------
Status (2023-12-21): this was done by #249 on the frontend side, and will be completed in the engine by #372

### LINKED ISSUE #370: Impl error on type in trait.

W95Psp:
Fixes #370

This PR:
 - adds a `Self` kind for `ImplSource`. This is useful for generic information within a trait.
 - propagate `impl` on function calls.  

Fixes #20

**Note**: this PR failing is blocking for @cmester0's #349 and #348.
Since this is a bit urgent, my plan is to make failure silent for now: any backend that relies on this feature will then possibly generate bad code if we hit the edge case that currently makes this PR fails.

### DIFF
--- a/frontend/exporter/src/rustc_utils.rs
+++ b/frontend/exporter/src/rustc_utils.rs
@@ -23,6 +23,13 @@ impl<'tcx> ty::Predicate<'tcx> {
     }
 }
 
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+pub struct AnnotatedPredicate<'tcx> {
+    pub is_extra_self_predicate: bool,
+    pub predicate: ty::Predicate<'tcx>,
+    pub span: rustc_span::Span,
+}
+
 #[extension_traits::extension(pub trait TyCtxtExtPredOrAbove)]
 impl<'tcx> ty::TyCtxt<'tcx> {
     /// Just like `TyCtxt::predicates_defined_on`, but in the case of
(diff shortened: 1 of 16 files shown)


tests/attributes/src/lib.rs
```diff
@@ -1,41 +1,9 @@
-use hax_lib_macros as hax;
-
-// dummy max value
-const u32_max: u32 = 90000;
-
-#[hax::requires(x > 10 && y > 10 && z > 10 && x + y + z < u32_max)]
```
franziskuskiefer, line 6: Why did you remove this code?

frontend/exporter/src/traits.rs
```diff
@@ -116,7 +117,8 @@ mod search_clause {
             let predicates = tcx
                 .predicates_defined_on_or_above(self.def_id())
                 .into_iter()
-                .map(|(predicate, _)| predicate);
+                // .map(|(predicate, _)| predicate);
```
franziskuskiefer, line 120: remove?

frontend/exporter/src/traits.rs
```diff
@@ -272,7 +283,7 @@ impl<'tcx> IntoImplExpr<'tcx> for rustc_middle::ty::PolyTraitRef<'tcx> {
                 r#trait: self.skip_binder().sinto(s),
             }
             .with_args(impl_exprs(s, &x.nested)),
-            x => ImplExprAtom::Todo(format!("{:#?}\n\n{:#?}", x, self)).into(),
+            x => ImplExprAtom::Todo(format!("ImplExprAtom::Todo {:#?}\n\n{:#?}", x, self)).into(),
```
franziskuskiefer, line 286: Can we add issues to todos?

frontend/exporter/src/rustc_utils.rs
```diff
@@ -23,24 +23,75 @@ impl<'tcx> ty::Predicate<'tcx> {
...
+    pub predicate: ty::Predicate<'tcx>,
+    pub span: rustc_span::Span,
+}
+
 #[extension_traits::extension(pub trait TyCtxtExtPredOrAbove)]
 impl<'tcx> ty::TyCtxt<'tcx> {
     /// Just like `TyCtxt::predicates_defined_on`, but in the case of
     /// a trait or impl item, also includes the predicates defined on
     /// the parent.
+    // fn predicates_defined_on_or_above(
```
franziskuskiefer, line 38: remove?

franziskuskiefer (approved):
lgtm with a few nits

tests/attributes/src/lib.rs
W95Psp: Oops, that is a mistake (I have the bad habit of using this `attributes` test as a playground...)

franziskuskiefer (changes requested):
Cancelling for now, looks like there's a merge conflict and ci failing

franziskuskiefer (approved):
Lets get this in.

PR #373, opened 2023-11-21
ProVerif: Basic struct translation

jschneider-bensch:
This PR is a first attempt at translating Rust `structs` to ProVerif.

For a given Rust `struct` I'm generating a constructor for a fixed type `state` on the ProVerif side. In addition, it's also generating a `reduc` for every field of the `struct` that basically allows projection access to the fields in ProVerif. Note that in Rust, accessing like that is only possible in the enclosing module unless the field is `pub` which I'm not distinguishing right now.
For now, it doesn't use attributes to "mark" which Rust `struct` should be translated as `state` constructors in ProVerif, but just translates everything.

As an example for what this PR does, it takes the following Rust code:
```rust
// Record struct with single field
struct Ainitial {
    x: u8,
}

// Record struct with multiple fields
struct A {
    one: usize,
    two: usize,
}

// Non-record struct
struct B(usize);
```
to this ProVerif develoment:
```text
channel c.
type state.

fun t_A(bitstring, bitstring): state.
reduc forall f_one: bitstring, f_two: bitstring;
    accessor_t_A_f_one(f_one, f_two) = f_one.
reduc forall f_one: bitstring, f_two: bitstring;
    accessor_t_A_f_two(f_one, f_two) = f_two.

fun t_Ainitial(bitstring): state.
reduc forall f_x: bitstring;
    accessor_t_Ainitial_f_x(f_x) = f_x.

fun t_B(bitstring): state.
reduc forall t_B_0: bitstring;
    accessor_t_B__0(t_B_0) = t_B_0.

process
    0
```
I've also included a toolchain test to this effect.


engine/backends/proverif/proverif_backend.ml
```diff
@@ -114,14 +114,74 @@ module Print = struct
       method ty_int _ = string "bitstring"
 
       method! item' item =
+        let fun_and_reduc name variants =
+          let constructor = List.hd variants in
+          match constructor with
+          | None -> empty
+          | Some constructor ->
+              let field_prefix =
+                if constructor.is_record then string ""
```
W95Psp, line 123: Small nit: why not using `empty`?
```suggestion
                if constructor.is_record then empty
```

engine/backends/proverif/proverif_backend.ml
```diff
@@ -114,14 +114,74 @@ module Print = struct
...
+                string "fun" ^^ space ^^ print#concrete_ident name
+                ^^ iblock parens fun_args_types
+                ^^ string ": state."
+              in
+              let reduc_line =
+                string "reduc forall " ^^ iblock Fn.id fun_args_full
+                ^^ string ";"
+              in
+              let build_accessor (ident, ty, attr) =
+                string "accessor_" ^^ print#concrete_ident name ^^ string "_"
```
W95Psp, line 157: Another small, thing: [pprint exposes `underscore`](http://cambium.inria.fr/~fpottier/pprint/doc/pprint/PPrint/index.html#val-underscore), so you can write:
```suggestion
                string "accessor_" ^^ print#concrete_ident name ^^ underscore
```

(same below with `dot`, but that's really not important at all, just a note, you can ignore those things if you like)

W95Psp (approved):
Thanks, that looks good to me! :)
(just a few very minor things)

engine/backends/proverif/proverif_backend.ml
jschneider-bensch: Thanks for pointing these out! I keep forgetting what's already pre-defined in `PPrint`.

PR #383, opened 2023-12-04
disable debug asserts in lib

franziskuskiefer:
When not compiling with debug asserts we don't even want the debug_assert macro


W95Psp (approved):
LGTM!

PR #385, opened 2023-12-06
ProVerif backend: Translation of `fn`s to `letfun`s

jschneider-bensch:
This PR includes basic translation of Rust `fn`s into ProVerif `letfun` macros. The next steps would be separate treatment of those `fn`s that have been annotated as ProVerif processes and hooking up known `letfun`s where possible (At the moment I've introduced dummy constructors up to arity 7 in order to generate syntactically valid, if useless, ProVerif for function calls).


engine/backends/proverif/proverif_backend.ml
```diff
@@ -166,21 +165,90 @@ module Print = struct
...
+              (print#expr_at Expr_App_arg >> group)
+              args
+          in
+          let f = dummy_fn |> group in
+          f ^^ iblock parens args
+
+      method! literal : Generic_printer_base.literal_ctx -> literal fn =
+        fun _ctx -> function
+          | String s -> string "no char literals in ProVerif"
+          | Char c -> string "no char literals in ProVerif"
```
W95Psp, line 244: You should `raise` an error just like above, and you might need to add a `Unsupported` variant to the [enum Kind](https://github.com/hacspec/hax/blob/fea5dbcf4ab371ba6cb18e8c3f7e3742d4982dd1/frontend/diagnostics/src/lib.rs#L131)

engine/backends/proverif/proverif_backend.ml
```diff
@@ -166,21 +165,90 @@ module Print = struct
...
+          | Int { value; negative; _ } ->
+              string "int2bitstring"
+              ^^ iblock parens
+                   (string value |> precede (if negative then minus else empty))
+          | Float { value; kind; negative } ->
+              string "no float literals in ProVerif"
```
W95Psp, line 250: Same, you  should `raise` here

engine/backends/proverif/proverif_backend.ml
```diff
@@ -166,21 +165,90 @@ module Print = struct
...
+          | TParam i -> print#local_ident i
+          | TApp _ -> super#ty ctx ty
+          | _ -> string "bitstring"
+
+      method! expr_app : expr -> expr list -> generic_value list fn =
+        fun f args _generic_args ->
+          let dummy_fn =
+            match List.length args with
+            | n when n < 8 -> string "dummy_fn_" ^^ PPrint.OCaml.int n
+            | _ -> string "not_supported"
```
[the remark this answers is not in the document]
jschneider-bensch: Thanks, will do!

engine/backends/proverif/proverif_backend.ml
jschneider-bensch: I could introduce an `Unsupported` variant to signal that the backend doesn't support the input, but doesn't `ExplicitRejection` already cover that? We're not in a phase anymore here, but the intention seems the same to me.

W95Psp: Oh, you're right, that's a good idea yes!

engine/backends/proverif/proverif_backend.ml
```diff
@@ -166,21 +165,90 @@ module Print = struct
...
+      method! doc_construct_inductive
+          : is_record:bool ->
+            is_struct:bool ->
+            constructor:concrete_ident ->
+            base:document option ->
+            (global_ident * document) list fn =
+        fun ~is_record ~is_struct:_ ~constructor ~base:_ args ->
+          if is_record then
+            string "t_"
+            (* FIXME: How do I get at the ident from the struct definition instead? *)
```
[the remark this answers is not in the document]
jschneider-bensch: I've updated it to use the new API from #392.

W95Psp (approved):
That looks good to me, let's get it in! :)

jschneider-bensch:
Thanks! I had experimented with moving the ProVerif toolchain tests into their own directory on this branch, but I've reverted that again, just now.

PR #386, opened 2023-12-07
Coq proof lib

cmester0:
Add coq library files.


cmester0:
From what I can see `MachineIntergers.v` file is not directly part of CompCert library, so we do need it here.

proof-libs/coq/coq/src/Hacspec_lib.v
```diff
@@ -0,0 +1,146 @@
...
+
+From Coq Require Import ZArith.
+Import List.ListNotations.
+Open Scope Z_scope.
+Open Scope bool_scope.
+
+From Hacspec Require Export MachineIntegers.
+From Hacspec Require Export Hacspec_Lib.
+
+(** Should be moved to Hacspec_Lib.v **)
```
spitters, line 12: What's this comment?

proof-libs/coq/coq/src/Hacspec_lib.v
```diff
@@ -0,0 +1,146 @@
...
+
+Definition array_slice {a: Type} `{Default a} {len : nat} (input: nseq a len) {WS} (start: @int WS) (slice_len: @int WS) : seq a := slice (array_to_seq input) (unsigned start) (unsigned (start .+ slice_len)).
+Notation slice := array_slice.
+Definition seq_new {A: Type} `{Default A} {WS} (len: @int WS) : seq A := seq_new (unsigned len).
+Notation new_seq := seq_new.
+Notation num_exact_chunks := seq_num_exact_chunks.
+Notation get_exact_chunk := seq_get_exact_chunk.
+Definition set_chunk {a: Type} `{Default a} {len} (s: seq a) {WS} (chunk_len: @int WS) (chunk_num: @int WS) (chunk: array_or_seq a len) : seq a := seq_set_chunk s (unsigned chunk_len) (unsigned chunk_num) (as_seq chunk).
+Definition set_exact_chunk {a} `{H : Default a} {len} s {WS} := @set_chunk a H len s WS.
+     Notation get_remainder_chunk := seq_get_remainder_chunk.
```
spitters, line 89: Nit: remove spaces

proof-libs/coq/coq/src/QuickChickLib.v
```diff
@@ -0,0 +1,100 @@
...
+
+#[global] Instance show_public_byte_seq : Show (public_byte_seq) :=
+  Build_Show (public_byte_seq) (fun x =>
+    match x with 
+    | [] => "[]"%string
+    | x :: xs => ("[" ++ fold_left (fun a b => (a ++ " " ++ show b)) xs (show x) ++ "]")%string
+    end).
+Definition g_public_byte_seq : G (public_byte_seq) :=
+  listOf arbitrary.
+  (* @genList int8 gen_int8. *)
```
spitters, line 94: Remove these comments?

spitters:
> From what I can see `MachineIntergers.v` file is not directly part of CompCert library, so we do need it here.

Interesting. So, where does it come from then? If possible, we should try to avoid code duplication.

cmester0:
> > From what I can see `MachineIntergers.v` file is not directly part of CompCert library, so we do need it here.
> 
> Interesting. So, where does it come from then? If possible, we should try to avoid code duplication.

Hmm it might be here https://github.com/AbsInt/CompCert/blob/master/lib/Integers.v , however it seems modifications been made to that file, as modules are used, while we just need a typeclass.

spitters:
Yes, that seems like a likely source.
It's possible that we are just using an old version. 
For the last yard, we've switched to the math-comp integers. It would be good to switch to the new stdlib one that's in preparation when it becomes available.

cmester0:
Yes, but I would like to add some testing, and structure the library a bit, but I guess those can be follow up PRs.

spitters:
Why are you going back to the compcert integers? The jasmin/math-comp ones are needed for the jasmin connection.

Getting this merged would help us make a choice. Moreover, that library comes from fiat-crypto, which we also connect to with hacspec.
https://github.com/coq/coq/pull/17043

spitters:
Ok. Yes, that's a good idea. Didn't we use the MC/jasmin integers for the
SSProve backend?

cmester0:
Closed in favor of #987 and annotated core lib.

PR #390, opened 2023-12-11
feat(fstar): add `interfaces` flag

W95Psp:
Now `cargo hax into fstar` accepts an optional `--interface` flag.
It takes inclusion clauses (similar to `-i`, e.g.  `+module::path::** -module::path::foo::**`).
By default, the backend produces no fsti, but any selected module will have a fsti extraction.

### DIFF
--- a/cli/options/src/lib.rs
+++ b/cli/options/src/lib.rs
@@ -138,7 +138,7 @@ pub struct FStarOptions {
         value_delimiter = ' ',
         allow_hyphen_values(true)
     )]
-    // TODO: InclusionKind is a bit too expressive here (i.e. `!` makes no sense)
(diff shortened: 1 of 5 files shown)


cli/options/src/lib.rs
```diff
@@ -121,9 +121,25 @@ pub struct FStarOptions {
...
+    /// every item is included. Rust path chunks can be either a
+    /// concrete string, or a glob (just like bash globs, but with
+    /// Rust paths).
+    #[arg(
+        long,
+        value_parser = parse_inclusion_clause,
+        value_delimiter = ' ',
+        allow_hyphen_values(true)
+    )]
+    // TODO: InclusionKind is a bit too expressive here (i.e. `!` makes no sense)
```
franziskuskiefer, line 141: Can you file an issue for that and reference here?

franziskuskiefer (approved):

PR #388, opened 2023-12-11
fix(engine): make `fn` types always arrows

W95Psp:
Now functions *always* have a arrow type of arity >=1.

PR #392, opened 2023-12-13
feat(engine): add transoforms to name policies

W95Psp:


franziskuskiefer (approved):
lgtm. Let's wait for review from @jschneider-bensch before merging.

W95Psp:
Uh, I think I included too much stuff in my commit, I need to rebase it

engine/lib/ast_utils.ml
```diff
@@ -226,6 +226,91 @@ module Make (F : Features.T) = struct
...
+            | _ -> { e; span; typ }
+        end
+      in
+      obj#visit_expr ()
+
+    let insert_prop_coercions : (module ERROR) -> expr -> expr =
+      let lift_fun _f = None in
+      let lift_arg _arg ~goal:_ = None in
+      fun err -> insert_coercions err lift_fun lift_arg
+
```
jschneider-bensch, line 313: This is probably unrelated to the name policy, right?

engine/lib/concrete_ident/names.rs
```diff
@@ -112,6 +112,7 @@ fn unsize() {}
 mod hax {
     fn failure() {}
     struct Failure;
+    struct Prop;
```
jschneider-bensch, line 115: Also unrelated?

jschneider-bensch (commented):
Thanks for this! I tested it out with my current branch of the PV backend :+1:  and it works a charm for simplifying the naming of struct and enum constructors and fields.

I noticed that `fn`s with no arguments now received a dummy argument `_ : ()`. Could that be related to the other changes in the PR?

engine/lib/ast_utils.ml
W95Psp: Yes, exactly! :sweat_smile:

W95Psp:
The `()` thing is implemented by #388, so that now every Rust function is always translated as a function of non-zero arity in Hax's AST

jschneider-bensch (approved):
Thanks, this looks good to me!

Issue #402, opened 2023-12-21
F*: interface mode: missed binders


W95Psp:
In F*, binders `val` declarations cannot be patterns.
In the interface mode, we generate `val` declarations out of Rust functions possibly annotated with pre- or post-conditions.

Rust functions accept any pattern on function inputs. If a binding introduced by a pattern happens to be used in a post or a pre, currently, this produces bad F* code.

We should give non-trivial patterns (e.g. not variable patterns) names and rewrite pre and post-condition.

W95Psp:
This is still relevant.
This transformation should be done outside of the backend, as a phase.

franziskuskiefer:
@W95Psp please add a reproducer

W95Psp:
✅

PR #403, opened 2023-12-21
fix(engine/fstar): always name binders on `val`s

W95Psp:
This PR adds binder names as much as possible when producing `val` declarations in `fsti` interfaces F* files.
I say "as much as possible" because of issue #402.


franziskuskiefer (approved):
Looks like this slipped through. Let's get it in.

Issue #405, opened 2024-01-02
Allow loops without side effect


W95Psp:
Currently, extracting a loop whose body performs no side effect results in the following error:
```
error[HAX0001]: (FunctionalizeLoops) something is not implemented yet.
                Loop without mutation?
```

This error comes from the phase `FunctionalizeLoops`, which --for now-- assumes purity.
Under the assumption things are pure, a loop without side effect is a no op; such a loop is useless but should be supported anyway.
Also, some backend might use `FunctionalizeLoops` without assuming purity.

W95Psp:
It seems like this still reproduce, but I recall you did something around this @maximebuyse: do you remember something as well?

cmester0:
We also run into this issue when translating ElectionGuard

franziskuskiefer:
Lets allow this, maybe add a warning.

franziskuskiefer:
What's the state of this? We run into this in libcrux code.

W95Psp:
Sorry Franziskus, I missed that comment.
@maximebuyse I think we wanted to fix this at some point, this should not be too complicated, what do you think?

maximebuyse:
> Sorry Franziskus, I missed that comment. [@maximebuyse](https://github.com/maximebuyse) I think we wanted to fix this at some point, this should not be too complicated, what do you think?

Yes, I can try to find the time for that this week.

franziskuskiefer:
I think the title is also a little misleading as the same error is thrown when there's something panicking in the loop. For example, adding an `assert!(i < 10);` to the example throws the same error.

W95Psp:
Indeed Franziskus, I agree that from a Rust perspective, an assert (that is, a conditional panic) can be seen as a side effect.

However, in hax, panics are not effectful: a panic is translated to a point in the program where we have to provide a proof that this point is unreachable.
Similarly, every function in hax is supposed to be faithful to its signature: if the signature is pure, then the function is assumed to be pure.
So if a loop body calls a function `f` that actually writes to disk, this error will show up.

PR #406, opened 2024-01-02
Delete rust-issues.md

### LINKED ISSUE #236: Update `rust-issues.md`
Most of the things listed there are supported now

W95Psp:
See #236: those issues are fixed, modulo #405.

I think we should delete that file and stick to issues instead.
Maybe we could add an `unsupported-syntax` label and add a link to https://github.com/hacspec/hax/labels/unsupported-syntax in the README or book?
If we stick to a particular naming scheme for such `unsupported-syntax` labeled issues, that would give users a good view of what Hax rejects. We could tag `wont-fix` + close issues for syntaxes we never plan to support (if any!). 

Fixes #236


franziskuskiefer (approved):
Sounds good.
Can you create the label and documentation?

PR #421, opened 2024-01-07
Remove some leftover hacspec-v2 strings

franziskuskiefer:


W95Psp (approved):
Good catch, thanks :)

PR #430, opened 2024-01-13
Secret independence support for F*

karthikbhargavan:
This PR adds a set of proof libraries for F* that incorporates secrecy levels.
We may push the secret independence analysis to Rust in the future, hence the library modifications in this PR create a new set of libraries that are independent.
Keeping the two sets of libraries in sync will be a bit painful, so we should seek to resolve this in future PRs.


W95Psp (approved):
I think this is OK for now, but currently this brings a lot of duplication, but hopefully, we find a way of dealing with that later on (filing #435 meanwhile).

Also, I think you rewrote the lib using a few commits behind main, thus the diff between folders `fstar-secret-integers` and `fstar` include the inverse of a few commits from https://github.com/hacspec/hax/pull/424

PR #439, opened 2024-01-17
Fix question marks

### LINKED ISSUE #399: Fix monadic transform for `?`s
- [ ] Add tests
 - [ ] F*

W95Psp:
This PR fixes #399 and adds some more tests.

### DIFF
--- a/tests/side-effects/src/lib.rs
+++ b/tests/side-effects/src/lib.rs
@@ -1,47 +1,47 @@
 #![allow(dead_code)]
 fn add3(x: u32, y: u32, z: u32) -> u32 {
-    x + y + z
+    x.wrapping_add(y).wrapping_add(z)
 }
 
 fn local_mutation(mut x: u32) -> u32 {
     let mut y = 0;
     if {
-        x = x + 1;
+        x = x.wrapping_add(1);
         x > 3
     } {
-        x = x - 3;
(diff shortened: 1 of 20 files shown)


tests/side-effects/src/lib.rs
```diff
@@ -1,47 +1,47 @@
 #![allow(dead_code)]
 fn add3(x: u32, y: u32, z: u32) -> u32 {
-    x + y + z
+    x.wrapping_add(y).wrapping_add(z)
```
franziskuskiefer, line 3: How are these changes related to the PR?
While you're here, add comments what these tests are doing.

franziskuskiefer (changes requested):
Thanks, generally looks good to me.
But lets start documenting things. Please comment all the code you're touching.

tests/side-effects/src/lib.rs
W95Psp: I change everything from `+` to `wrapping_add` so that I make sure (locally) that things are typechecking in F*, without bothering to add preconditions everywhere so that the addition doesn't overflow!

I want at some point to make our "unit" tests to be typed-checked in the backends at CI, but for now there is no such infrastructure :(

I added comments!

franziskuskiefer (approved):
Thanks!

PR #440, opened 2024-01-18
Fix core ops range tc

W95Psp:
This PR:
 - fixes the F* typeclass inference in Core.Ops.Range (cc @karthikbhargavan);
 - tweak the makefiles a bit.


karthikbhargavan (approved):
These changes are needed for Kyber. I added a couple more changes to fix universes, hoping that they help avoid regressions.

PR #455, opened 2024-01-25
Extend generic printer API with auxiliary information

jschneider-bensch:
This PR allows passing in auxiliary information on creation of a `print` object. This is useful if a generic-printer-based backend requires some information to be collected before doing a certain print pass on the AST, e.g. in the ProVerif backend I want to first scan the AST for function definitions so that I can recognize them during printing at the call-site.

I can imagine other uses as well, e.g. we might not need to insert a full preamble of pre-defined models every time, but only those that are really necessary to model the contents of the AST. Being able to go through the AST once, collecting this information and making it available on later print passes that construct the preamble is a simple solution here as well.

### DIFF
--- a/engine/backends/proverif/proverif_backend.ml
+++ b/engine/backends/proverif/proverif_backend.ml
@@ -128,7 +128,7 @@ module Print = struct
   let is_known_function fname =
     List.exists ~f:(assoc_known_function fname) library_functions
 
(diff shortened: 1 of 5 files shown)


W95Psp (approved):
Nice, that looks good to me, thank you

PR #459, opened 2024-01-25
Crypto Primitives Abstraction Layer

jschneider-bensch:
This PR introduces an abstraction layer for cryptographic primitives in the `hax_lib_protocol` crate and rewrites the Noise example to use only that.
The abstraction layer at the moment is just a thin wrapper around `libcrux` but in the future could be extended to other backends.

Two things before this is ready for review:
- [ ] Docs for the abstraction layer
- [ ] ProVerif extraction of the Noise example seems to go into an infinite loop


spitters:
Just curious, will it fit with the crates in https://github.com/RustCrypto ?
They are interested in hacspec integration:
https://github.com/RustCrypto/meta/issues/9

hax-lib-protocol/src/cal.rs
```diff
@@ -0,0 +1,194 @@
...
+use crate::ProtocolError;
+
+/// An abstract Diffie-Hellman scalar.
+#[derive(Clone)]
+pub struct DHScalar(Vec<u8>);
+
+impl DHScalar {
+    /// Wrap bytes into a Diffie-Hellman scalar. Does *not* perform
+    /// input validation.
+    pub fn from_bytes(bytes: &[u8]) -> Self {
```
franziskuskiefer, line 13: Is there a reason not to use `From`? This is fine too. Just wondering.

hax-lib-protocol/Cargo.toml
```diff
@@ -11,3 +11,4 @@ readme.workspace = true
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+libcrux = "0.0.2-pre.2"
```
franziskuskiefer, line 14: Can we get a readme here for the directory explaining a little what it is doing?

hax-lib-protocol/src/cal.rs
```diff
@@ -0,0 +1,194 @@
...
+        AEADIV(libcrux::aead::Iv::new(bytes).unwrap())
+    }
+}
+
+/// An abstract AEAD authentication tag.
+pub struct AEADTag(libcrux::aead::Tag);
+impl AEADTag {
+    pub fn from_bytes(bytes: &[u8]) -> Self {
+        /// Attempt deserialization of an AEAD tag from `bytes`. Panics if
```
franziskuskiefer, line 105: Should this be a doc comment on the function?

hax-lib-protocol/src/cal.rs
```diff
@@ -0,0 +1,194 @@
+//! This module defines a cryptographic abstraction layer for use in
```
franziskuskiefer, line 1: Given that we want an abstraction, we should have traits to make it easier to implement this again for other backends.
But that's not urgent. Let's open an issue for this.

hax-lib-protocol/src/lib.rs
```diff
@@ -1,14 +1,11 @@
-//! This crate provides types and traits for implementing a protocol state
-//! machine.
-//!
-//! A protocol party is conceived of as having a set of possible states, one of
-//! which is the initial state. Transitioning to a different state is possible
-//! either through receiving and processing a message or through writing a
-//! message.
+pub mod cal;
```
franziskuskiefer, line 1: What does `cal` mean?

Can you add new crate docs?

franziskuskiefer (changes requested):
Thanks, looks good. Just a few nits and questions.

engine/backends/proverif/proverif_backend.ml
```diff
@@ -30,6 +31,7 @@ module SubtypeToInputLanguage
           (* and type slice = Features.Off.slice *)
           (* and type raw_pointer = Features.Off.raw_pointer *)
             with type early_exit = Features.On.early_exit
+             and type slice = Features.On.slice
```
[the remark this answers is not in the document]
jschneider-bensch: The rewritten noise example contains slices, so it would be rejected during extraction if I didn't allow slices here.

hax-lib-protocol/src/cal.rs
jschneider-bensch: No reason, I didn't think of implementing `From`, when adapting the `hacspec_lib` API to this.

hax-lib-protocol/src/lib.rs
jschneider-bensch: `cal` could be short for cryptographic abstraction layer, but yeah it pretty opaque! How about `abstract_crypto` instead?

franziskuskiefer: Right now it's not very abstract 🙂 I'd probably just call it `crypto`. But you can as well keep `cal` and just put a doc comment on it to explain what it means.

hax-lib-protocol/src/cal.rs
jschneider-bensch: I created #471 to track this.

hax-lib-protocol/src/lib.rs
jschneider-bensch: That's true. :D
I like your suggestion.

franziskuskiefer (approved):
Lets get this in.

PR #469, opened 2024-01-29
fix(cli): use `stderr` for detection mode

W95Psp:
Rustc already (sometimes) uses `stdout` for sending some information, when using `--deps`, rustc sends extra things on `stdout` thus we cannot parse it anylonger.

Fixing #468


franziskuskiefer (approved):
Thanks, lgtm.

PR #470, opened 2024-01-29
feat(engine/fstar): `--interfaces`: add option to extract only `fsti`

### LINKED ISSUE #397: CLI: F*: `interfaces` flag is a bit too expressive
`hax into fstar --interfaces` and `hax into --include_namespaces` accept the same kind of value (they share a same type and parser).

With the latter, the user can either:
 - exclude a namespace (prefix `-`);
 - include an item and all its local dependencies (prefix `+`);
 - include just an item without including its dependencies (prefix `+!`). 

The former accepts the same grammar, but `+!` makes no sense: either we extract an interface for a module, or we don't, but there is no third option.

Here, either the Rust type `InclusionClause` should be made polymorphic to allow or not `+!`, or we should have a different type.

### LINKED ISSUE #465: F* cli: `--interface`: add an option to extract only an `fsti` without a `fst`
When something of the shape `+!Foo` is passed to `--interfaces`, only a `Foo.fsti` should be extracted (ignoring the `Foo.fst`)

W95Psp:
Fixes #397 and fixes #465

### DIFF
--- a/cli/options/src/lib.rs
+++ b/cli/options/src/lib.rs
@@ -127,18 +127,18 @@ pub struct FStarOptions {
     /// Modules for which Hax should extract interfaces (`*.fsti`
     /// files). By default we extract no interface. This flag expects
     /// a space-separated list of inclusion clauses. An inclusion
-    /// clause is a Rust path prefixed with `+` or `-`. `-` excludes
-    /// any matched item, `+` includes any matched item. By default,
-    /// every item is included. Rust path chunks can be either a
-    /// concrete string, or a glob (just like bash globs, but with
-    /// Rust paths).
+    /// clause is a Rust path prefixed with `+`, `+!` or `-`. `-`
+    /// excludes any matched item, `+` includes any matched item, `+!`
+    /// extracts only an interface and drops its implementation. By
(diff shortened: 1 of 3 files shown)


franziskuskiefer (changes requested):
So I can't get both anymore? That's a little weird.
Now I have to explicitly add all the modules I want to have interfaces for. Can't we make this a little nicer?

franziskuskiefer (approved):
Ok, lets get this in. It works.
We may want to improve on this later. Right now we have to add `+!` twice (for `-i` and `--interfaces`), to get only an interfaces, which is a little unintuitive. If the `-i` flag is only about what to include, a `+` should be enough.

W95Psp:
You mean that if you want only the interface for Foo (without impl), then you don't want to extract dependencies of Foo's items, right?

So maybe we should have, for `-i` "modifies":
 - `+` includes items with dependencies and with bodies;
 - `+!` include items without dependencies but with bodies;
 - `+!!` (or something else...) include items without dependencies or bodies.

But then, we need to have a representation for items without bodies, define what that means for types, for traits, etc. And also we need to define what it means in the various backend, to have empty bodies, which is not easy.

PR #473, opened 2024-01-29
feat(backend/fstar): patterns: allow shallow POr

### LINKED ISSUE #462: Support (shallow) disjunctive patterns in F*
Disjunctive patterns aren't supported in the F* backend. They are supported since #306.

There is a test in https://github.com/hacspec/hax/blob/main/tests/pattern-or/src/lib.rs

W95Psp:
Fixes #462


franziskuskiefer (approved):
Thanks! 🥳

Issue #448, opened 2024-01-22
[PV/Noise] Provide PV definitions for core library functions / types


jschneider-bensch:
- [ ] `Alloc__slice__Impl__to_vec`
- [ ] `Core__ops__deref__Deref__deref`
- [ ] `Core__ops__index__Index__index`
- [ ] `Rust_primitives__unsize`
- [ ] `Core__num__Impl_9__to_le_bytes`
- [ ] `Alloc__slice__Impl__into_vec`
- [ ] `Alloc__vec__Impl_1__truncate`
- [ ] `Alloc__vec__Impl_2__extend_from_slice`
- [ ] `Alloc__slice__Impl__concat`
- [ ] `Core__option__Impl__is_some`
- [ ] `Core__clone__Clone__clone`
- [ ] `Core__cmp__PartialEq__eq`
- [ ] `Core__cmp__PartialEq__ne`
- [ ] `Core__cmp__PartialOrd__lt`
- [ ] `Core__ops__arith__Add__add`
- [ ] `Core__ops__arith__Sub__sub`
- [ ] `Core__option__Option__None`
- [ ] `Core__option__Option__Some`
- [ ] `Core__result__Impl__map_err`
- [ ] `Core__option__Option__Some`
- [ ] `Core__option__Option__None`
- [ ] `Core__ops__range__Range`
- [ ] `Core__option__Option`
- [ ] `Alloc__vec__Vec`

Issue #447, opened 2024-01-22
[PV/Noise] Provide PV definitions for crypto dependencies


jschneider-bensch:
- [ ]      `Hax_lib_protocol__crypto__hash`
 - [ ]      `Hax_lib_protocol__crypto__hmac`
 - [ ]      `Hax_lib_protocol__crypto__aead_decrypt`
 - [ ]      `Hax_lib_protocol__crypto__aead_encrypt`
 - [ ]      `Hax_lib_protocol__crypto__dh_scalar_multiply`
 - [ ]      `Hax_lib_protocol__crypto__dh_scalar_multiply_base`
 - [ ]      `Hax_lib_protocol__crypto__Impl__from_bytes` (`hax_lib_protocol::cal::impl__DHScalar__from_bytes`)
 - [ ]      `Hax_lib_protocol__crypto__Impl_1__from_bytes` (`hax_lib_protocol::cal::impl__AEADTag__from_bytes`)
 - [ ]      `Hax_lib_protocol__crypto__Impl_4__from_bytes` (`hax_lib_protocol::cal::impl__DHElement__from_bytes`)
 - [ ]      `Hax_lib_protocol__crypto__Impl_5__from_bytes` (`hax_lib_protocol::cal::impl__AEADKey__from_bytes`)
 - [ ]      `Hax_lib_protocol__crypto__Impl_6__from_bytes` (`hax_lib_protocol::cal::impl__AEADIV__from_bytes`)

 - [ ]      `Hax_lib_protocol__crypto__HashAlgorithm__Sha256`
 - [ ]      `Hax_lib_protocol__crypto__DHGroup__X25519`
 - [ ]      `Hax_lib_protocol__crypto__AEADAlgorithm__Chacha20Poly1305`
 - [ ]      `Hax_lib_protocol__crypto__HMACAlgorithm__Sha256`

 - [ ]      `Hax_lib_protocol__crypto__DHScalar`

franziskuskiefer:
The hacspec_lib functions should be replaced by core functions, once that dependency is gone. (Also see comment on #452.)

PR #477, opened 2024-01-30
Infrastructure for known functions, types and constructors

### LINKED ISSUE #449: [ProVerif] Insert definitions of crypto dependencies + core library
We want to be able to recognize known (i.e. from `core` or another pre-defined library) functions and types in the AST and replace them with a pre-defined translation.

jschneider-bensch:
This PR adds more infrastructure to recognize known types, constructors and constructor patterns.
It also provides placeholder stubs for inserting pre-defined translations of these in a follow up.

Fixes #449.


engine/names/src/lib.rs
```diff
@@ -1,11 +1,43 @@
...
+    let _ = slice.len();
+    let _ = slice.to_vec();
+    let _ = [slice, slice].concat();
+    let mut v = vec![0];
+    v.extend_from_slice(slice);
+    v.truncate(1);
+
+    let _ = hash(HashAlgorithm::Sha256, &bytes);
+    let _ = cipher_text.clone();
+    // End crypto abstractions example
```
W95Psp, line 39: You can put this code in a separate `crypto_abstractions.rs` (and add a `mod crypto_abstractions;` to that file) if you like, it would be nicer than begin-end comments IMO :)

And you could `use hax_lib_protocol::crypto::*` only there.

engine/backends/proverif/proverif_backend.ml
```diff
@@ -115,35 +115,207 @@ module Print = struct
   let library_functions :
       (Concrete_ident_generated.name * (AST.expr list -> document)) list =
     [
-      (Core__ops__try_trait__Try__branch, fun args -> empty);
-      (* just an example *)
+      (* Core dependencies *)
```
W95Psp, line 118: What do you plan on doing with those?
Do you plan to rename each of those manually to something else?

Since you are building the library on ProVerif's side as well, can't you just name functions with a scheme we can directly produce, without the indirection of rewriting all those symbols?

engine/backends/proverif/proverif_backend.ml
```diff
@@ -115,35 +115,207 @@ module Print = struct
...
+
   let is_known_function fname =
-    List.exists ~f:(assoc_known_function fname) library_functions
+    List.exists ~f:(assoc_known_name fname) library_functions
+
+  let is_known_constructor cname =
+    List.exists ~f:(assoc_known_name cname) library_constructors
+
+  let is_known_type tname =
+    List.exists ~f:(assoc_known_name tname) library_types
```
W95Psp, line 285: I know those `translate_known_*` functions are called only after calling their associated `is_known_*` functions, but I don't like to much those `find_exn`...
Maybe instead of doing `match _ with | ... where is_known_* -> code`, we should `match _ with | ... -> (match List.find ... with | Some ... -> code | _ -> None)`?

jschneider-bensch: Thanks for this suggestion, I'll try that.
Perhaps I can also refactor this more anyway, since it's all basically the same logic.

engine/backends/proverif/proverif_backend.ml
W95Psp: Well, I mean that instead of manually saying `hax_lib_protocol::cal::hmac` is actually `foobar` in the proverif side, why not calling that `cal_hmac` in proverif, and just translate all the item `hax_lib_protocol::cal::foo` as `cal_foo`? or even, why not use a fully qualified name there? i.e. `hax_lib_protocol__cal__hmac`

jschneider-bensch: Ah, I think I understand now, thanks! I'm wondering if that would only work if we just want translate one for one to matching names and not to something more complicated on the ProVerif side.
An example may be when we want to drop some of the arguments of a function when translating to ProVerif, then we would need special logic for that translation in particular. Sorry, this is a bit abstract right now. Do you think I could address it properly when I work on #447 and #448, where I have to fill in the translations?

engine/backends/proverif/proverif_backend.ml
jschneider-bensch: I've simplified this a bit and removed these `find_exn`s.

W95Psp: Cool, thanks!

jschneider-bensch:
I realized I had introduced a regression on translating `Result` constructor by messing up the order of match arms. I fixed it by changing the order now, but should revisit when I work on #448, so it uses the same translation mechanism as the other `core` names.

W95Psp (approved):

PR #486, opened 2024-01-31
[ProVerif] Translate crate-internal names

### LINKED ISSUE #450: [ProVerif] Coherent translation of intra-crate function calls

jschneider-bensch:
This PR provides coherent translation of crate-internal names, e.g. where before you had `some_function` at the definition site and `crate::some::mod::function` at the call site in another module, you now get `crate__some__mod__some_function` everywhere. Similar for types and their constructors.

Fixes #450.

### DIFF
--- a/engine/backends/proverif/proverif_backend.ml
+++ b/engine/backends/proverif/proverif_backend.ml
@@ -297,7 +297,7 @@ module Print = struct
           | { e = GlobalVar name; _ } -> (
               match name with
               | `Projector (`Concrete i) | `Concrete i ->
-                  super#concrete_ident' ~under_current_ns:true i |> group
+                  print#concrete_ident' ~under_current_ns:false i |> group
               | _ -> super#expr_at Expr_App_f f |> group)
         in
 
@@ -339,6 +339,8 @@ module Print = struct
             when Global_ident.eq_name Core__ops__try_trait__Try__branch n ->
               super#expr' ctx expr.e
           | _ -> super#expr' ctx e
+
+      method concrete_ident = print#concrete_ident' ~under_current_ns:false
 
       method! item' item =
         let fun_and_reduc base_name constructor =
@@ -399,8 +401,8 @@ module Print = struct
             in
             string "letfun" ^^ space
             ^^ align
-                 (print#concrete_ident name ^^ params_string ^^ space ^^ equals
-                ^^ hardline
+                 (print#concrete_ident' ~under_current_ns:false name
+                 ^^ params_string ^^ space ^^ equals ^^ hardline
                  ^^ print#expr_at Item_Fn_body body
                  ^^ dot)
         (* `struct` definitions are transformed into simple constructors and `reduc`s for accessing fields. *)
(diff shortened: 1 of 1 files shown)


engine/backends/proverif/proverif_backend.ml
```diff
@@ -285,6 +285,24 @@ module Print = struct
...
+            (comma ^^ break 1)
+            (print#expr_at Expr_App_arg >> group)
+            args
+        in
+        let f =
+          match f with
+          | { e = GlobalVar name; _ } -> (
+              match name with
+              | `Projector (`Concrete i) | `Concrete i ->
+                  print#concrete_ident' ~under_current_ns:false i |> group
```
W95Psp, line 300: Do you need this `print#concrete_ident' ~under_current_ns:false` (insead of just using `print#concrete_ident ...`), since you overrided `concrete_ident`?

engine/backends/proverif/proverif_backend.ml
```diff
@@ -381,8 +401,8 @@ module Print = struct
             in
             string "letfun" ^^ space
             ^^ align
-                 (print#concrete_ident name ^^ params_string ^^ space ^^ equals
-                ^^ hardline
+                 (print#concrete_ident' ~under_current_ns:false name
```
W95Psp, line 404: Same as above

engine/backends/proverif/proverif_backend.ml
```diff
@@ -419,6 +439,16 @@ module Print = struct
...
           ^^ (print#expr_at Expr_Let_body body |> group)
 
+      method concrete_ident' ~(under_current_ns : bool) : concrete_ident fn =
+        fun id ->
+          if under_current_ns then print#name_of_concrete_ident id
+          else
+            let crate, path = print#namespace_of_concrete_ident id in
+            let full_path = crate :: path in
+            separate_map (underscore ^^ underscore) utf8string full_path
+            ^^ underscore ^^ underscore
```
W95Psp, line 449: we might want to move this renaming to the name policy thing at some point.
Otherwise, name collisions are possible
But for now that's completely fine, let's see that later

W95Psp (approved):

engine/backends/proverif/proverif_backend.ml
jschneider-bensch: You're right, that's not necessary.

engine/backends/proverif/proverif_backend.ml
jschneider-bensch: For sure, if we had like a `path_separator` that you could specify in the name policy, then this could be a bit simpler. :+1: 
Do you mean collisions are possible in general with this ad hoc method of defining a path renaming scheme, or with this specific one? I thought I could avoid collisions here by including the full path always.

W95Psp: Yes, exactly, I think we should do that at some point. I'm basically doing something similar, but only in certain cases (e.g. I collapse certain kind of namespaces, not all in all circumstances)

What I mean is that Rust identifiers `foo::a__b` and `foo::a::b` will collide with the current renaming scheme.
It's unlikely, but that'd be nice to avoid. In the name policy thing, I escape all separators to avoid that, but that's a bit cumbersome, so yeah, for now I think we don't care, and then we'll move that logic in the name policy module at some point :)

jschneider-bensch: Oops, I forgot about unescaped separators! Thanks for pointing this out.

PR #487, opened 2024-02-01
Frontend: fix `Self`

### LINKED ISSUE #482: Frontend: Fix Self with sub obligations

W95Psp:
This PR allows to the `Self` impl expression to carry a subpath.
For example with a where clause `where Self::AssocType: SomeTrait`, calling a method `foo` that belongs to `SomeTrait` will use the implementation located at path "Self/AssocItem("AssocType")/Parent("SomeTrait")".

Fixes #482.

### DIFF
--- a/engine/lib/import_thir.ml
+++ b/engine/lib/import_thir.ml
@@ -917,6 +917,19 @@ end) : EXPR = struct
 
   and c_impl_expr_atom (span : Thir.span) (ie : Thir.impl_expr_atom) : impl_expr
       =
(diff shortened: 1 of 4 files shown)


W95Psp:
@sonmarcho that's the one commit you need!

franziskuskiefer (commented):
Can we get a test for whatever this is doing?
It also needs a description in the PR.

franziskuskiefer (approved):
🙏🏻

PR #489, opened 2024-02-01
feat(nix): make derivations lighter, use `cachix`

W95Psp:
This PR:
 - adds a GitHub action that pushes binaries to Cachix;
 - reformulates a bit the README on Nix install;
 - add instructions on how to use binary cache when using Nix;
 - the OCaml compiler leaves absolute paths to itself inside the binaries it produces... thus this PR also removes such references (otherwise Nix tracks them and the dependency closure contains 500Mb of OCaml compiler).

### DIFF
--- a/.github/workflows/install_and_test.yml
+++ b/.github/workflows/install_and_test.yml
@@ -40,8 +40,10 @@ jobs:
       uses: actions/checkout@v3
       with:
         repository: 'hacspec/specs'
+        path: specs
         
     - name: Extract specifications
+      working-directory: specs
       run: |
         paths=$(tomlq -r '.workspace.members | .[]' Cargo.toml)
         for cratePath in $paths; do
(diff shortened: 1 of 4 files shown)


.github/workflows/install_and_test.yml
```diff
@@ -66,3 +66,12 @@ jobs:
...
           hacspec-weierstrass-coq
           hacspec-weierstrass-fstar
+
+    - name: Push to Cachix
+      if: ${{ github.event_name == 'workflow_dispatch'  || github.event_name == 'merge_group'  }}
+      run: |
+        nix profile install nixpkgs#cachix nixpkgs#jq
+        nix build --json \
+          | jq -r '.[].outputs | to_entries[].value' \
+          | cachix push hax
```
franziskuskiefer, line 76: So this doesn't need any sort of authentication?

franziskuskiefer (approved):

.github/workflows/install_and_test.yml
W95Psp: It does, but I added a secret on the repo's config!

PR #518, opened 2024-02-19
feat(proof-libs/fstar): bit vec: use funext

W95Psp:
This commit changes the representation of bit vectors from partial maps to partial maps with functional extensibility.

This allows us to assert the equality of two bit-vectors when they are bit-wise equal.


karthikbhargavan (approved):
Seems to work nicely.

PR #519, opened 2024-02-20
feat(ci): test Kyber extraction and proofs

W95Psp:
This commit adds to CI the extraction to F* of Kyber and TC.
This mostly is a copy-paste from https://github.com/cryspen/libcrux/blob/main/.github/workflows/hax.yml.

Note this job is a bit long, thus it is ran only on when merging in the merge queue.

Status: this is working, but currently Kyber TC is broken, so we need https://github.com/cryspen/libcrux/pull/202 to be merged first

### DIFF
--- a/.github/workflows/kyber.yml
+++ b/.github/workflows/kyber.yml
@@ -1,0 +1,101 @@
+name: Extract and TC Kyber
+
+on:
+  pull_request:
+  merge_group:
+  workflow_dispatch:
+  push:
+    branches: [main]
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  hax:
+    if: ${{ github.event_name == 'workflow_dispatch' || github.event_name == 'merge_group' }}
+    runs-on: "ubuntu-latest"
+
+    steps:
+      - name: ⤵ Clone Libcrux repository
+        uses: actions/checkout@v4
+        with:
+          repository: cryspen/libcrux
+
+      - uses: actions/checkout@v4
+        with:
+          path: hax
+
+      - uses: DeterminateSystems/nix-installer-action@main
+      - uses: DeterminateSystems/magic-nix-cache-action@main
+
+      - name: ⤵ Install hax
+        run: |
+          tree /home/runner/work/
+          nix profile install ./hax
+
+      - name: ⤵ Install FStar
+        run: nix profile install github:FStarLang/FStar/v2024.01.13
+
+      - name: ⤵ Clone HACL-star repository
+        uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/kyber.yml
```diff
@@ -0,0 +1,101 @@
...
+  workflow_dispatch:
+  push:
+    branches: [main]
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  hax:
+    # if: ${{ github.event_name == 'workflow_dispatch' || github.event_name == 'merge_group' }}
```
franziskuskiefer, line 15: Do you want to run it only on the merge queue? 🙂 
Also, it only needs 10 minutes and the `Install & test / nix-action` job is slower. So we could run it directly. But I'm fine keeping it on merge queue only.

.github/workflows/kyber.yml
```diff
@@ -0,0 +1,101 @@
...
+      - uses: actions/checkout@v4
+        with:
+          path: hax
+
+      - uses: DeterminateSystems/nix-installer-action@main
+      - uses: DeterminateSystems/magic-nix-cache-action@main
+
+      - name: ⤵ Install hax
+        run: |
+          tree /home/runner/work/
```
franziskuskiefer, line 33: We probably don't need the debugging output.

```suggestion
```

.github/workflows/kyber.yml
```diff
@@ -0,0 +1,101 @@
...
+      - name: ⤵ Install FStar
+        run: nix profile install github:FStarLang/FStar/v2024.01.13
+
+      - name: ⤵ Clone HACL-star repository
+        uses: actions/checkout@v4
+        with:
+          repository: hacl-star/hacl-star
+          path: hacl-star
+
+      - name: 🏃 Extract the Kyber reference code
```
franziskuskiefer, line 45: We should factor out these steps at some point to have fewer logic in here and avoid having to touch things in two places every time something changes/breaks. Can you file an issue for unifying things here?

franziskuskiefer (approved):
Looks like some debug leftovers are still in here, otherwise this looks good to me.

.github/workflows/kyber.yml
W95Psp: Oh, I forgot to uncomment that indeed... Thanks!
I think it's fast only because hax was cached (in the log that takes less than 1 minute), but otherwise, that will take around 9 minutes of build, so something like 20 minutes in total

.github/workflows/kyber.yml
W95Psp: :+1:  https://github.com/hacspec/hax/issues/520

PR #524, opened 2024-02-22
Fix #523: associated type bounds names 

### LINKED ISSUE #523: Duplicated name caused by Sized trait
```
pub trait Lang: Sized {
    type Var;
    fn s(self, _: i32) -> (Self, Self::Var);
}
```
extracts to
```
module Rust
open Core
open FStar.Mul

class t_Lang (v_Self: Type) = {
  [@@@ FStar.Tactics.Typeclasses.no_method]_super_529871881:Core.Marker.t_Sized v_Self;
  f_Var:Type;
  f_Var:Core.Marker.t_Sized _;
  f_s:v_Self -> i32 -> (v_Self & _)
}
```
which contains twice the field `f_Var` and fails with
```
* Error 47 at Rust.fst(9,2-9,7):
  - Duplicate top-level names [Rust.__proj__Mkt_Lang__item__f_Var]
  - Previously declared at Rust.fst(8,2-8,7)
```

W95Psp:
Fixes #523


frontend/exporter/src/traits.rs
```diff
@@ -376,14 +376,30 @@ impl<'tcx> IntoImplExpr<'tcx> for rustc_middle::ty::PolyTraitRef<'tcx> {
     }
 }
 
-pub fn clause_id_of_predicate(predicate: rustc_middle::ty::Predicate) -> u64 {
+#[tracing::instrument(level = "trace", skip(s))]
+pub fn clause_id_of_predicate<'tcx, S: UnderOwnerState<'tcx>>(
```
franziskuskiefer, line 380: So you are attaching a hash here to make the ids unique? Please add documentation what is happening and why.

frontend/exporter/Cargo.toml
```diff
@@ -23,3 +23,4 @@ tracing.workspace = true
 paste = "1.0.11"
 extension-traits = "1.0.1"
 lazy_static = "1.4.0"
+deterministic-hash = "1.0.1"
```
franziskuskiefer, line 26: What hash function is this? Can't we use something better? This looks like a pretty odd dependency that hasn't been updated in over 2 years.

franziskuskiefer (changes requested):
Thanks

frontend/exporter/Cargo.toml
W95Psp: It is just a wrapper around other hashers: it's simply "normalizing" some types (i.e. `usize` is mapped to `u64` regardless of the platform) before calling the inner hasher, so that the hash is deterministic over various plateforms.

franziskuskiefer: Looking at what this is doing we should rather copy the few lines that are needed here. I want to avoid getting into a dependency hell here and this doesn't seem worth it.

W95Psp: Makes sense, that's done in e82e74a 🫡

franziskuskiefer (approved):
Thanks 🙏🏻

PR #531, opened 2024-02-26
Introduce `opaque_type` attribute

### LINKED ISSUE #525: Add an `opaque` attribute
We should provide a `opaque` attribute to make type opaque.

W95Psp:
This PR:
 - introduces `opaque_type` attribute that can mark enums and structs so that their (F* for now) extraction doesn't reveal their definition.
 - introduce an option in the test harness so that we can write tests that specify backend options;
 - add a test.

Fixes #525

### DIFF
--- a/engine/backends/fstar/fstar_backend.ml
+++ b/engine/backends/fstar/fstar_backend.ml
@@ -846,6 +846,31 @@ struct
                             |> List.map ~f:to_pattern) ),
                    pty e.span ty );
                ] )
+    | Type { name; generics; _ }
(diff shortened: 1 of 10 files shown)


karthikbhargavan (commented):
Can we rename this to hax_opaque or something? Or is it already guarded by some hax prefix and so it is not necessary?

W95Psp:
That's indeed already in some `hax`-prefixed module, so things should be fine :)

karthikbhargavan (approved):

PR #538, opened 2024-02-28
feat(proofs-lib/fstar): intro Core.Mem, Core.Fmt, impl__usize__leadin…

W95Psp:
This PR adds a few missing definitions for lax checking riot-rs


proof-libs/fstar/core/Core.Mem.fsti
```diff
@@ -0,0 +1,5 @@
+module Core.Mem
+open Rust_primitives
+
+// TODO: remove default type
```
franziskuskiefer, line 4: Can you add an issue for this and link it here?

franziskuskiefer (approved):
Thanks

PR #548, opened 2024-03-05
fix(engine): translate associated types to `Self` (e.g. `Self::T`)

### LINKED ISSUE #536: Associated types are extracted to underscores which cannot be resolved
```
pub trait Lang {
    type Var;
    fn f(self, _: Self::Var) -> Self::Var;
}
```
extracts to 
```
module Rust.Lang
open Core
open FStar.Mul

class t_Lang (v_Self: Type) = {
  f_Var:Type;
  f_Var_9841369763059924533:Core.Marker.t_Sized _;
  f_f:v_Self -> _ -> _
}
```
which fails with
```
* Error 66 at /.../extraction/Rust.Lang.fst(8,48-8,49):
  - Failed to resolve implicit argument ?3
    of type Type
    introduced for
        user-provided implicit term at
        /.../extraction/Rust.Lang.fst(8,48-8,49)
```

W95Psp:
This commit fixes #536.  The AST of hax has all the information for projected or associated things, but the F* backend doesn't use them much for now. This commit adds support for exactly `Self::WhateverAssociatedType`, but not for more (i.e. parent associated type).

### DIFF
--- a/engine/backends/fstar/fstar_backend.ml
+++ b/engine/backends/fstar/fstar_backend.ml
@@ -280,6 +280,9 @@ struct
     | TArray { typ; length } ->
         F.mk_e_app (F.term_of_lid [ "t_Array" ]) [ pty span typ; pexpr length ]
     | TParam i -> F.term @@ F.AST.Var (F.lid_of_id @@ plocal_ident i)
+    | TAssociatedType { impl = Self; item } ->
+        F.term
+        @@ F.AST.Var (F.lid [ U.Concrete_ident_view.to_definition_name item ])
     | TAssociatedType { impl; item } -> (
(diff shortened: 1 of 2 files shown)


franziskuskiefer (approved):
Sounds like this is not fixing the full issue. Can you either open a follow-up issue or not close it and add a comment to what's working now and what remains open?

W95Psp:
Well, yes, it's solving only one case of associated type.
Looking around at what is still buggy, I'm surprised to see I actually implemented some support for other such things in the F* backend, I forgot about those!

I opened https://github.com/hacspec/hax/issues/549 for more complex associated types, but I'm sure there are more such situations

PR #552, opened 2024-03-05
feat(engine/phases): drop-references: accept `?`

W95Psp:
The drop references phases constrained the input language to be question mark free, but it did not make any sense


franziskuskiefer (approved):
Can you add a test for this?

W95Psp:
I actually cannot: this is a type-level only change. This phase used to have a too strong "precondition" for no reason. This PR allows backends to apply this phase a bit earlier, basically.

PR #553, opened 2024-03-05
Charon traits merge

W95Psp:


W95Psp:
On my side, I'm OK with the changes introduced in this PR.
It seems huge but it's not big, the `mir.rs` put aside.

@franziskuskiefer, I'm postponing tests on MIR to https://github.com/hacspec/hax/issues/481: I have no time for that, and this PR has been waiting for too long, Son needs it.

However, I think we need to have https://github.com/hacspec/hax/issues/481 as soon as possible: let's plan that next Thursday at CC.

franziskuskiefer (approved):

PR #555, opened 2024-03-06
feat(f*): `array_of_list`: inline the length of the array

W95Psp:
This commit changes `array_of_list l` into `array_of_list N l` with `N` a literal. We know `N` at extraction time, so it's free to add, and it's helping type inference, especially in presence of typeclasses.


franziskuskiefer (approved):
More information ℹ️

PR #556, opened 2024-03-06
fix(frontend): extract string literals as strings, not as u8 slices

W95Psp:


frontend/exporter/src/constant_utils.rs
```diff
@@ -68,6 +68,18 @@ impl From<ConstantFieldExpr> for FieldExpr {
     }
 }
 
+impl ConstantLiteral {
+    /// Rustc always represents string constants as `&[u8]`, but this
+    /// is not nice to consume. This associated function interpret
+    /// bytes as an unicode string, and as a byte string otherwise.
+    fn byte_str(bytes: Vec<u8>, style: StrStyle) -> Self {
+        match String::from_utf8(bytes.clone()) {
```
franziskuskiefer, line 76: What about u16?
Strings are super hard, that's why they're usually not strings but bytes. This is probably fine for most cases. We should just make sure we know this is happening in case we run into issues with strings.

franziskuskiefer (approved):

frontend/exporter/src/constant_utils.rs
W95Psp: Yeah, but in this case, I think it's only because of some constraints in rustc's internals: `b"whatever"` literals are not typed strings, nor are scalar. Basically in Rustc's type-level constants representation, everything is a bunch of `u8`s, even ADTs. (everything is represented with trees of u8, basically)
So in this exact place, I think it's just an encoding for a `str` literal, which should really be a unicode thing.

What do you mean for u16?

franziskuskiefer: That should've been utf16. Anyway, this change is fine. Just wanted to make sure we remember this is done in case something brakes 😄

PR #557, opened 2024-03-06
F*: insert `solve ()` in impls parent bound fields

W95Psp:
This PR:
 - propagate parent bounds impl expr and trait information from the frontend to the AST;
 - in the F* backend: use this information to populate parent bounds fields;
 - introduces a new phase `drop_sized_trait`, that drops the sized trait (which is useless in our backends but extremely verbose).

As a side effect, this also causes some item reordering: now the dependency analysis takes care of parent bounds.
That causes some reordering in the extraction of Kyber, I will push a PR.

### DIFF
--- a/engine/backends/coq/ssprove/ssprove_backend.ml
+++ b/engine/backends/coq/ssprove/ssprove_backend.ml
@@ -1834,12 +1834,12 @@ struct
                                       body,
                                     ret_typ ));
                              ]
-                         | IIType ty ->
+                         | IIType {typ; _} ->
                              [
                                SSP.AST.LetDef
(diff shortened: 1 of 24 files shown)


franziskuskiefer (approved):
Thanks.
As always, please add comments to the parts that you're touching to slowly populate the code with comments and make it more approachable.

What about backends that may need the sized trait? Should we allow the option to keep it? Maybe just file an issue to keep track of this, or put it into a comment in the code.

W95Psp:
That's the nice thing: this `drop_sized_trait` thing is a phase, so each backend might apply it or not! :grinning:

PR #550, opened 2024-03-05
[PV] Enable Bertie handshake extraction

jschneider-bensch:
This PR contains many (sorry!) improvements to the ProVerif backend which in sum enable a typechecking extraction of the full Bertie TLS handshake, given a very small additional library of definitions.

While the model does typecheck, it is still unlikely to be useful for analysis, since it does not yet incorporate e.g. the fixes from #559.
 
Fixes #543 
Fixes #542

PR #561, opened 2024-03-11
Extraction macros for ProVerif

### LINKED ISSUE #560: [ProVerif] Annotation macros for constructors, handwritten models

jschneider-bensch:
This PR addresses #560, by introducing macros to annotate functions as either needing handwritten models or that they should be extracted as constructors of their return types.

~It's a draft at the moment, since it builds on changes in #550.~
Fixes #560


hax-lib-macros/src/lib.rs
```diff
@@ -520,3 +520,21 @@ pub fn protocol_messages(_attr: pm::TokenStream, item: pm::TokenStream) -> pm::T
...
+/// A marker indicating a `fn` as a ProVerif process initialization.
+#[proc_macro_error]
+#[proc_macro_attribute]
+pub fn pv_constructor(_attr: pm::TokenStream, item: pm::TokenStream) -> pm::TokenStream {
+    let item: ItemFn = parse_macro_input!(item);
+    let attr = AttrPayload::PVConstructor;
+    quote! {#attr #item}.into()
+}
+
+/// A marker indicating a `fn` as a ProVerif process initialization.
```
W95Psp, line 533: Are the doc comments up to date here?

hax-lib-macros/types/src/lib.rs
```diff
@@ -54,6 +54,8 @@ pub enum AssociationRole {
     ProcessWrite,
     ProcessInit,
     ProtocolMessages,
+    PVConstructor,
+    PVHandwritten,
```
W95Psp, line 58: Do you actually need constructors both in AssociationRole and in AttrPayload?
`AssociationRole` defines the nature of the relation of two linked items, e.g. a function `f` is linked to `f_precondition` (via a UUID) and the nature of their relation is "Requires" (I need to write a comment for that :sweat_smile:)

hax-lib-macros/src/lib.rs
jschneider-bensch: Good catch, they're not!

hax-lib-macros/types/src/lib.rs
jschneider-bensch: I see, yes, then `AssociationRole` is not necessary at all, since there are no associated items.

W95Psp (approved):
Lgtm :)

Issue #562, opened 2024-03-11
[ProVerif] Translate early return version of question marks


jschneider-bensch:
The translation of question marks into early returns (cf. #559) means the handling of question marks by the backend has to be adapted.

jschneider-bensch:
Addressed in #564 .

PR #564, opened 2024-03-11
Adapt translation to early return question mark

jschneider-bensch:
This PR addresses #562.

~It depends on changes in #550 and #559 so it is a draft until those are in.~


engine/backends/proverif/proverif_backend.ml
```diff
@@ -278,6 +311,27 @@ module Make (Options : OPTS) : MAKE = struct
...
+                  (*   | TApp { ident = `TupleType _ } *)
+                  (*   (\* Tuple types should be translated without conversion from bitstring *\) *)
+                  (*     -> *)
+                  (*       iblock parens inner_field_doc *)
+                  (*   | _ -> *)
+                  (*       iblock parens *)
+                  (*         (inner_field_type_doc ^^ string "_to_bitstring" *)
+                  (*         ^^ iblock parens inner_field_doc) *)
+                  (* in *)
+                  (* string "Some" ^^ inner_block *)
```
W95Psp, line 334: Should we just remove this?

engine/backends/proverif/proverif_backend.ml
```diff
@@ -278,6 +311,27 @@ module Make (Options : OPTS) : MAKE = struct
                           ^^ iblock parens inner_field_doc)
                   in
                   string "Some" ^^ inner_block
+              | PConstruct { name; args }
+              (* Some(expr) -> Some(<expr_typ_to_bitstring>(expr))*)
```
W95Psp, line 315: Can you rephrase the comment `handles "Some(expr) -> Some..."` or something? I actually thought this was commented code :sweat_smile:

W95Psp (approved):
A few nits, but otherwise that looks good to me :)

engine/backends/proverif/proverif_backend.ml
jschneider-bensch: Oops, totally! I wrote a more sensible comment explaining what happens and why.

PR #568, opened 2024-03-12
fix(engine/phase/cf-into-monads): use `call_Constructor`, not `call`

W95Psp:
Otherwise, we get a function call instead of a an expression that construct an inductive.


W95Psp:
aaah, I merged by mistake in the branch of this PR :'( that's not too a big deal though

franziskuskiefer (approved):

PR #571, opened 2024-03-13
feat: translate enum descriminant and their casts

W95Psp:
For every enum `E` which can be casted to a primitive type, this PR derives a `E_cast_to_repr: E -> R` with `R` the primitive type specified on `E` with `#[repr(R)]`, or the rust default if nothing was specified.


test-harness/src/snapshots/toolchain__reordering into-fstar.snap
```diff
@@ -34,6 +36,11 @@ type t_Foo =
   | Foo_A : t_Foo
   | Foo_B : t_Foo
 
+let t_Foo_cast_to_repr (x: t_Foo) : isize =
+  match x with
+  | Foo_A  -> isz 0
```
franziskuskiefer, line 41: This looks like C and D aren't handled. Is that on purpose?

franziskuskiefer: Shouldn't this be the translation of this?
```rust
enum Foo {
    A = 1,
    B = 5,
    C(),
    D {},
}
```
Then there's a C and D as well.

W95Psp: No, that's the extraction of something else, the snapshot you want to look at is test-harness/src/snapshots/toolchain__enum-repr into-fstar.snap

franziskuskiefer: I suggest stop re-using the same, meaningless name. Using descriptive names is as important in tests as it is in the rest of the code.

W95Psp: Yes, that's a good idea indeed. I'm gonna push a commit

W95Psp: done

franziskuskiefer (approved):

Issue #44, opened 2023-04-14
Add facilities for grouping items as recursive bundles 


W95Psp:
Currently there is no easy way of grouping recursive items from a backend.
However, the feature is already implemented in module `engine/lib/dependencies.ml`.

Related to that, there is https://github.com/hacspec/hax/issues/396.
But this is completely orthogonal: we should still have a simple way for backends to view a set of items as a set of recursive bundles.

Thus, this issue is about exposing a function `item list -> item list list` using what already exists in `engine/lib/dependencies.ml`.

W95Psp:
Fixed by #935

PR #572, opened 2024-03-13
Basic rec fstar

W95Psp:
This PR makes the F* backend output a `let rec` for simple recursive functions, not mutual ones.

Supporting mutual recursive functions is trickier because of namespaces, see #44.


karthikbhargavan (approved):
Looks good.

PR #573, opened 2024-03-13
fix(engine): monomorphize update_at only for vecs, slices and arrays

W95Psp:
The engine used to "monomorphize" `update_at` into specialized versions: in F*, the TC resolution used to slow down a lot typechecking.
However, this monomirphization is wrong in general, so this PR gates this transformation with additional checks on the types.


karthikbhargavan (approved):
Maybe add a short description and an example with a newtype?
The code changes look good.

PR #574, opened 2024-03-13
feat(proof-libs/fstar): add missing implementations for Bertie

W95Psp:
This PR adds some models for Rust core/std/alloc/Rand libraries to `proof-libs/fstar`
These need to be better structured in separate folders in a future PR.


karthikbhargavan (approved):
Needs a description ("Filling in missing F* libraries for Bertie")

Putting `let!` in Rust_primitives is a bit weird. Can you better document/justify this?

proof-libs/fstar/core/Rand_core.fsti
```diff
@@ -0,0 +1,12 @@
+module Rand_core
```
franziskuskiefer, line 1: This is not part of the Rust core library (or std or alloc). So it probably shouldn't be in here.

std and alloc should go into their own folders too. But that may be ok for now.

franziskuskiefer (changes requested):

proof-libs/fstar/core/Rand_core.fsti
karthikbhargavan: Merging this for now, and adding an issue to clean up this folder structure.

karthikbhargavan (approved):

PR #575, opened 2024-03-15
Ssprove backend lib

cmester0:


cmester0:
There are still some issues, and the library is not fully implemented, however this should be enough for now, we can always add more as we go.

PR #577, opened 2024-03-21
Add match guard to AST

cmester0:
I have added the match guard to the AST. Currently no backend supports these, so this makes the rejection of this feature explicit. Furthermore we can now add a phase for removing match conditionals. This should also allow us to rewrite record pattern matching into guarded match statements (should help with #180 / #168).


cmester0:
We can add a phase that unrolls (guarded) matches, e.g.
```rust
    match foo {
        Foo { a, .. } if a > 10 => 1,
        Foo { b, .. } if b => 2,
        _ => 3
    };
```
becomes
```rust
let _ =
  if let Foo { a, .. } = foo && a > 10 {
    1
  } else if let Foo { b, .. } = foo && b {
    2
  } else if let _ = foo {
    3
  } else {
    panic!()
  };
```
Which should allow us to do matching on records by moving the projection to the right side

```rust
let _ =
  if let a = foo.a && a > 10 {
    1
  } else if let b = foo.b && b {
    2
  } else if let _ = foo {
    3
  } else {
    panic!()
  };
```

Issue #563, opened 2024-03-11
[ProVerif] Add more convenient way to include handwritten models


jschneider-bensch:
While the ProVerif backend at the moment allows marking items for manual modelling, it does not offer a convenient way to integrate a set of handwritten models after extraction.

To address this, the `pv_handwritten` attribute should be extended to (optionally) include perhaps a filename where a handwritten model for the given item is specified. This can then be inserted in the extraction in a way that allows automatic replacement of the filename indicator with the handwritten model.

jschneider-bensch:
Closing this in favor of #620.

PR #587, opened 2024-04-09
Inline backend macro

W95Psp:
This PR adds a set of proc macros (`fstar!`, `coq!`, `proverif!`) that allow for embedding F*/Coq/ProVerif/... expressions directly in Rust.
Technically, this PR adds a macro to `hax-lib-macros` and adds a `Quote` node in the AST in the engine.

[Documentation for the `fstar!`, `coq!` etc macros is here](https://github.com/hacspec/hax/pull/587/files#diff-923ccf3019aa37fbc741dad9a293c66d297fdd1ff2c80a17eb280d832b009023R544-R576)

This PR also merges the barrett and compress examples from https://github.com/hacspec/hax/tree/franziskus/toronto-2024, leveraging the `fstar!` macro to remove the need for patches.

PR #595, opened 2024-04-10
[ProVerif] Insert markers for including handwritten model instead of stub definitions

jschneider-bensch:
As one part of addressing #563, this PR replaces the stub definitions for designated-as-handwritten translations by commented-out inclusion markers of the form `(* {{ proofs/proverif/handwritten/<path_to_item_module>.pvl }} *)` that indicate where the handwritten model can be found.

The idea is to connect this to further tooling (e.g. in the hax driver?) which would replace these markers with the contents of the handwritten model, most probably replacing the first occurrence of a given handwritten model marker with it's file contents and removing the remaining ones, if any. This would ensure that handwritten models are inserted as early as possible and should avoid ProVerif missing a definition because it's defined later in the file.

As a caveat to this simple approach, removing the handwritten stubs entirely means the extracted model will not typecheck anymore if there are handwritten portions still missing. To address this, we could generate the stub *and* the inclusion marker and make the marker replacement tooling a bit smarter so it's able to also remove the stub definitions.

### DIFF
--- a/engine/backends/proverif/proverif_backend.ml
+++ b/engine/backends/proverif/proverif_backend.ml
@@ -224,6 +224,14 @@ module Make (Options : OPTS) : MAKE = struct
         (** Print a ProVerif letfun call. *)
 
         (* Helpers *)
+
+        method module_of_concrete_ident : concrete_ident fn =
+          fun id ->
+            let crate, path = print#namespace_of_concrete_ident id in
+            let full_path = crate :: path in
+            separate_map (underscore ^^ underscore) utf8string full_path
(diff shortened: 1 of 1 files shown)


franziskuskiefer:
@W95Psp shall this be close? If there are takeaways from this, maybe put them on an issue.

W95Psp:
Yes, let's close it. @jschneider-bensch we need to make the "inlining" macros (from #587) work in the PV backend though!

W95Psp:
https://github.com/hacspec/hax/issues/620 tracks that.

PR #597, opened 2024-04-11
fix(engine): allow default generics

### LINKED ISSUE #596: Missing const with default value support
We just hit this:

```
error[HAX0001]: (AST import) something is not implemented yet.
                c_generic_param:Const with a default
  --> src/riot-rs-runqueue/src/runqueue.rs:25:68
   |
25 | pub struct RunQueue<const N_QUEUES: usize, const N_THREADS: usize, const N_CORES: usize = 1> {
   |                                                                    ^^^^^^^^^^^^^^^^^^^^^^^^
```

W95Psp:
Default generics are just a nicety provided by Rust, but at the end, Rustc always inline the default when not provided. We don't want to propagate this nicety in the backends, since it's useful only to ease writing new code.

Thus, this PR just drops the default information if any.

Fixes #596

### DIFF
--- a/engine/backends/fstar/fstar_backend.ml
+++ b/engine/backends/fstar/fstar_backend.ml
@@ -578,10 +578,8 @@ struct
       let ident = plocal_ident p.ident in
       match p.kind with
       | GPLifetime _ -> Error.assertion_failure span "pgeneric_param:LIFETIME"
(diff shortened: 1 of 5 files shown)


karthikbhargavan (approved):
Changes look good to me.

PR #612, opened 2024-04-18
feat: include flag: allow for dropping bodies

W95Psp:
This PR adds the `+:` modifier to inclusion clauses. `-i '... +:some::glob::pattern'` will extract signature-only for the items matched by `some::glob::pattern`.

I added some doc in the book, see https://github.com/hacspec/book/pull/15

### DIFF
--- a/cli/options/src/lib.rs
+++ b/cli/options/src/lib.rs
@@ -145,8 +145,11 @@ pub struct FStarOptions {
     ifuel: u32,
     /// Modules for which Hax should extract interfaces (`*.fsti`
     /// files) in supplement to implementations (`*.fst` files). By
-    /// default we extract no interface, only implementations. This
-    /// flag expects a space-separated list of inclusion clauses. An
+    /// default we extract no interface, only implementations. If a
+    /// item is signature only (see the `+:` prefix of the
+    /// `--include_namespaces` flag of the `into` subcommand), then
+    /// its namespace is extracted with an interface. This flag
+    /// expects a space-separated list of inclusion clauses. An
     /// inclusion clause is a Rust path prefixed with `+`, `+!` or
     /// `-`. `-` means implementation only, `+!` means interface only
     /// and `+` means implementation and interface. Rust path chunks
@@ -198,6 +201,7 @@ enum DepsKind {
 enum InclusionKind {
     /// `+query` include the items selected by `query`
     Included(DepsKind),
+    SignatureOnly,
     Excluded,
 }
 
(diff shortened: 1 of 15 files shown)


franziskuskiefer (changes requested):
What exactly is the difference between `+:` and the existing `+~` and `+!`?

Also, this doesn't seem to do the right thing. We need more tests here.
Running this on the sha3 crate on the libcrux dev branch I'm getting and `.fst` as well and the following error
```
error[HAX0008]: (reject_RawOrMutPointer) ExplicitRejection { reason: "a node of kind [Raw_pointer] have been found in the AST" }
  --> libcrux-sha3/src/x4/internal.rs:27:1
```

W95Psp:
The `-i` clause selects items based on patterns. A pattern selects a number of items.
Patterns are processed from then left to the right, and a prefix modifier says what to do with the selected items.  
 - `+`: include selected items and all their dependencies (transitively);
 - `+~`: include selected items and direct dependencies (no transitivity);
 - `+!`: include selected items without its dependencies;
 - `+:`: include only the type signature of the selected items (dropping bodies).

Concerning the `sha3` crate, that's expected: `internal.rs` contains type definitions that contain unsupported features (raw pointers). A type definition cannot be extracted if any of its variants or fields contains unsupported code, we should just exclude that `internal.rs` file, or, at least, the faulty definition.

cli/options/src/lib.rs
```diff
@@ -238,14 +243,37 @@ fn parse_inclusion_clause(
...
+    /// item. The modifiers are:
+    ///
+    ///  - `+`: includes the selected items with their dependencies,
+    /// transitively (e.g. if function `f` calls `g` which in turn
+    /// calls `h`, then `+k::f` includes `f`, `g` and `h`)
+
+    /// {n} - `+~`: includes the selected items with their direct
+    /// dependencies only (following the previous example, `+~k::f`
+    /// would select `f` and `g`, but not `h`)
+
```
franziskuskiefer, line 270:
```suggestion
    ///
```

cli/options/src/lib.rs
```diff
@@ -238,14 +243,37 @@ fn parse_inclusion_clause(
...
+    /// {n} - `+!`: includes the selected items, without their
+    /// dependencies (`+!k::f` would only select `f`)
+
```
franziskuskiefer, line 273:
```suggestion
    ///
```

cli/options/src/lib.rs
```diff
@@ -238,14 +243,37 @@ fn parse_inclusion_clause(
...
+    /// This is a space-separated list of patterns prefixed with a
+    /// modifier, read from the left to the right.
+    ///
+    /// A pattern is a Rust path (say `mycrate::mymod::myfn`) where
+    /// globs are allowed: `*` matches any name
+    /// (e.g. `mycrate::mymod::myfn` is matched by
+    /// `mycrate::*::myfn`), while `**` matches any subpath, empty
+    /// included (e.g. `mycrate::mymod::myfn` is matched by
+    /// `**::myfn`).
+
```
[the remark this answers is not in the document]
franziskuskiefer: I'm not sure I understand. You get the line breaks in the help text when you add an empty comment line. And I don't think it makes it harder to read here.
Anyway, not a huge deal. It just looks a little weird here.

franziskuskiefer (approved):
Thanks, looks like this fixes things.

Let's file follow-ups for all the identified issues.

W95Psp:
:+1: filed #617 and #618

PR #615, opened 2024-04-22
refactor(engine): ast: make `quote` its own type

W95Psp:


franziskuskiefer (changes requested):
Please add an issue that this is fixing. Why are we doing this? And the PR needs a proper description in this regard as well.

cli/options/src/lib.rs
```diff
@@ -259,10 +259,11 @@ pub struct TranslationOptions {
     /// prefixed by modifiers are processed from left to right,
     /// excluding or including items. Each pattern selects a number of
     /// item. The modifiers are:
-    ///
-    ///  - `+`: includes the selected items with their dependencies,
-    /// transitively (e.g. if function `f` calls `g` which in turn
-    /// calls `h`, then `+k::f` includes `f`, `g` and `h`)
+
+    /// {n}{n} - `+`: includes the selected items with their
```
franziskuskiefer, line 263: I still don't understand why you don't just use regular comments that produce the spaces for you.

engine/bin/lib.ml
```diff
@@ -126,7 +126,15 @@ let run (options : Types.engine_options) : Types.output =
 let main (options : Types.engine_options) =
   Printexc.record_backtrace true;
   let result =
-    try Ok (run options) with e -> Error (e, Printexc.get_raw_backtrace ())
+    try Ok (run options) with
+    | Hax_engine.Diagnostics.SpanFreeError.Exn exn ->
+        Error
+          ( Failure
+              ("Uncatched hax exception (please report, this should not \
```
franziskuskiefer, line 133:
```suggestion
              ("Uncaught hax exception (please report, this should not \
```

engine/backends/fstar/fstar_backend.ml
```diff
@@ -573,16 +573,19 @@ struct
...
-          ~f:(function
-            | `Verbatim code -> code
-            | `Expr e -> pexpr e |> term_to_string
-            | `Pat p -> ppat p |> pat_to_string)
-          contents
-        |> String.concat |> F.term_of_string
+    | Quote quote -> pquote e.span quote |> F.term_of_string
     | _ -> .
 
+  and pquote span { contents; _ } =
```
franziskuskiefer, line 579: Please add some comments to the code while you touch it.

examples/barrett/src/lib.rs
```diff
@@ -38,7 +38,7 @@ pub fn barrett_reduce(value: FieldElement) -> FieldElement {
     // assert!(((quotient as i64) * (FIELD_MODULUS as i64)) < 9223372036854775807);
     let sub = quotient * FIELD_MODULUS;
 
-    hax::fstar!(r"Math.Lemmas.cancel_mul_mod (v $quotient) 3329");
+    hax::fstar_expr!(r"Math.Lemmas.cancel_mul_mod (v $quotient) 3329");
```
franziskuskiefer, line 41: Why was this renamed? `fstar` looks nicer.

tests/attributes/src/lib.rs
```diff
@@ -124,13 +124,53 @@ mod newtype_pattern {
     }
 }
 
+#[hax::fstar_before(r#"let before_${inlined_code} = "example before""#)]
```
franziskuskiefer, line 127: We should namespace this instead of having longer macro names.
How about something like `hax::fstar::before`?

franziskuskiefer (changes requested):
lgtm with some nits

tests/attributes/src/lib.rs
W95Psp: Yes, I was thinking about that as well, but abandoned the idea because we cannot export anything else than proc macros inside a proc-macro crate :(
But now you bring that up again, I could just re-export those long names in `hax-lib`, wdyt?

examples/barrett/src/lib.rs
W95Psp: That was to disambiguate between e.g. `fstar` and `fstar_before`, but yeah, that makes no much sense after all, I can revert to just `fstar`!

tests/attributes/src/lib.rs
franziskuskiefer: Ah right... re-exporting would certainly be good to ensure that only one import is needed. Feature guarded so that it doesn't need to be consumed.

engine/bin/lib.ml
W95Psp: Oops, my English is broken :sweat_smile:, thanks!

cli/options/src/lib.rs
W95Psp: That's because I want `clap` (or `rustdoc` at some other places) to look nice! In this situation, one newline is ignored by `clap` and the usual double space trick for markdown is not parsed. Here I could use one empty line, but then the spacing of the dashes are not aligned (clap seems to trim lines in a special way when they are prefixed with `{n}`)

tests/attributes/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-hax-lib-macros = { path = "../../hax-lib-macros" }
+# hax-lib-macros = { path = "../../hax-lib-macros" }
```
franziskuskiefer, line 7: Remove?

franziskuskiefer (approved):

PR #623, opened 2024-04-24
feat(setup/rebuild): use OPAMs prefix if possible

W95Psp:
This PR makes the `rebuild.sh` use OPAM prefix if possible, and also adds a warning if using `setup.sh` in a drity checkout of hax, hinting the use to use the rebuild script.


karthikbhargavan (approved):
Works on my setup. This will avoid some confusing errors, thanks!

PR #627, opened 2024-04-24
Retreive more trait refs

Nadrieril:
This PR retrieves trait refs in a few more cases. This is the only thing still needed for charon to build with hax `main`.

This caused issues in the past (see https://github.com/hacspec/hax/commit/a05c21aea91903cd80dfcb67ce194a63648fa670), some of which are solved by the second commit. Let's see if that's all of them.


W95Psp:
Nice, thanks, looking at that ASAP! :smiley:

frontend/exporter/src/types/copied.rs
```diff
@@ -1617,12 +1619,17 @@ pub enum Ty {
         rustc_middle::ty::TyKind::Adt(adt_def, substs) => {
             let def_id = adt_def.did().sinto(state);
             let generic_args: Vec<GenericArg> = substs.sinto(state);
-            Ty::Adt { def_id, generic_args }
+            let param_env = state.base().tcx.param_env(state.owner_id());
+            let trait_refs = solve_item_traits(state, param_env, adt_def.did(), substs, None);
```
W95Psp, line 1623: Just as a not here, I filed https://github.com/hacspec/hax/issues/637
That's what we discussed the other day, but that will be for another PR.

W95Psp (changes requested):
Just one thing, otherwise that looks good to me!

W95Psp (approved):
Thanks, let's get it in! :)

PR #629, opened 2024-04-24
Add support for math integers

### LINKED ISSUE #607: Nat/Int library
We want:
 - a type `Int` on which we can do arithmetic from Rust;
 - this type should be a newtype around i128 or bigint;
 - the translation should map to F*'s `int` (or coq `int`), and all operations on `Int` should be mapped to native ones in F*
 - we need to be able to lift machine integers to ints and lift down

W95Psp:
This PR adds support for math integers. Will fix #607.

This PR:
 - introduces a `Int` type  (for math integers) that implements common operations;
 - translate `Int` as primitive `int` in F*, and same thing for methods on `Int`;
 - introduces a Specialize phase that "monomorphize" operations (e.g. std's add becomes a specific function);
 - provides a `int` macro for math int literals;
 - adds tests for math ints and their operations.

### DIFF
--- a/engine/lib/phases/phase_specialize.ml
+++ b/engine/lib/phases/phase_specialize.ml
@@ -1,0 +1,165 @@
+open! Prelude
+
+module Make (F : Features.T) =
+  Phase_utils.MakeMonomorphicPhase
+    (F)
+    (struct
+      let phase_id = Diagnostics.Phase.Specialize
+
+      module A = Ast.Make (F)
+      module FB = F
+      module B = Ast.Make (F)
+      module U = Ast_utils.Make (F)
+      module Visitors = Ast_visitors.Make (F)
+      open A
+
+      open struct
+        open Concrete_ident_generated
+
+        type pattern = {
+          fn : name;
+          fn_replace : name;
+          args : (expr -> bool) list;
+          ret : ty -> bool;
+        }
+
+        type ('a, 'b) predicate = 'a -> 'b option
+
+        let mk (args : ('a, 'b) predicate list) (ret : ('c, 'd) predicate)
+            (fn : name) (fn_replace : name) : pattern =
+          let args = List.map ~f:(fun p x -> p x |> Option.is_some) args in
+          let ret t = ret t |> Option.is_some in
+          { fn; fn_replace; args; ret }
+
+        open struct
+          let etyp (e : expr) : ty = e.typ
+          let tref = function TRef { typ; _ } -> Some typ | _ -> None
+
+          let tapp0 = function
+            | TApp { ident; args = [] } -> Some ident
+            | _ -> None
+
+          let ( >>& ) (f1 : ('a, 'b) predicate) (f2 : ('b, 'c) predicate) :
+              ('a, 'c) predicate =
+           fun x -> Option.bind (f1 x) ~f:f2
+
+          let eq : 'a 'b. eq:('a -> 'b -> bool) -> 'a -> ('b, 'b) predicate =
+           fun ~eq x x' -> if eq x x' then Some x' else None
+
+          let eq_global_ident :
+              name -> (Ast.Global_ident.t, Ast.Global_ident.t) predicate =
+            eq ~eq:Ast.Global_ident.eq_name
+
+          let erase : 'a. ('a, unit) predicate = fun _ -> Some ()
+
+          let is_int : (ty, unit) predicate =
+            tapp0 >>& eq_global_ident Hax_lib__int__Int >>& erase
+
+          let any _ = Some ()
+          let int_any = mk [ etyp >> is_int ] any
+          let int_int_any = mk [ etyp >> is_int; etyp >> is_int ] any
+          let any_int = mk [ any ] is_int
+          let rint_any = mk [ etyp >> (tref >>& is_int) ] any
+
+          let rint_rint_any =
+            mk [ etyp >> (tref >>& is_int); etyp >> (tref >>& is_int) ] any
+
+          let any_rint = mk [ any ] (tref >>& is_int)
+
+          (* let is_type0 (ty_name : name) : (expr -> bool) option = *)
+          (*   Some *)
+          (*     (typ *)
+          (*     >> [%matches? *)
+          (*          TApp { ident; args = [] } when Ast.Global_ident.eq_name *)
+          (*                                           ty_name ident]) *)
+
+          (* let is_ref_type0 (ty_name : name) : (expr -> bool) option = *)
+          (*   Some *)
+          (*     (typ *)
+          (*     >> [%matches? *)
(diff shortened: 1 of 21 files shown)


hax-lib/Cargo.toml
```diff
@@ -12,8 +12,6 @@ description = "Hax-specific helpers for Rust programs"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-hax-lib-macros = { path = "../hax-lib-macros", optional = true }
-
-[features]
```
franziskuskiefer, line 17: Where did the features go?

hax-lib/Cargo.toml
```diff
@@ -12,8 +12,6 @@ description = "Hax-specific helpers for Rust programs"
...
-default = ["macros"]
-macros = ["dep:hax-lib-macros"]
+hax-lib-macros = { path = "../hax-lib-macros" }
```
franziskuskiefer, line 15: This needs a version.

hax-lib/Cargo.toml
```diff
@@ -12,8 +12,6 @@ description = "Hax-specific helpers for Rust programs"
...
+num-bigint = { version = "0.4.3", default-features = false, features = [] }
```
franziskuskiefer, line 16: Using both, `default-features` and `features` seems unnecessary.

hax-lib-macros/src/lib.rs
```diff
@@ -541,6 +541,35 @@ pub fn pv_handwritten(_attr: pm::TokenStream, item: pm::TokenStream) -> pm::Toke
     quote! {#attr #item}.into()
 }
 
+/// Create a mathematical integer. This macro expects a integer
+/// literal that consists in an optional minus sign followed by one or
+/// more digits.
+#[proc_macro_error]
+#[proc_macro]
+pub fn int(payload: pm::TokenStream) -> pm::TokenStream {
```
franziskuskiefer, line 549: I don't see a `hax` config. Where is the code compiled out for regular builds?

hax-lib/src/int/bigint.rs
```diff
@@ -0,0 +1,45 @@
...
+//! `BYTES`.
+//! Its interface provides bridges to `num_bigint::BigInt`.
+
+#[crate::exclude]
+mod no_extraction {
+    const BYTES: usize = 128;
+    #[derive(Copy, Clone)]
+    pub struct BigInt {
+        sign: num_bigint::Sign,
+        data: [u8; BYTES],
```
franziskuskiefer, line 12: 1024 bits seems a little low. This should be at least 1024 bytes.

hax-lib/src/int/bigint.rs
```diff
@@ -0,0 +1,45 @@
...
+    }
```
franziskuskiefer, line 13:
```suggestion
    }

```

hax-lib/src/int/bigint.rs
```diff
@@ -0,0 +1,45 @@
...
+    impl BigInt {
+        pub fn new(i: &num_bigint::BigInt) -> Self {
```
franziskuskiefer, line 15: All public functions need doc comments.

hax-lib/src/int/bigint.rs
```diff
@@ -0,0 +1,45 @@
...
+            let (sign, data) = i.to_bytes_be();
+            let data = data.try_into().expect(
+                "`copiable_bigint::BigInt::new`: too big, please consider increasing `BYTES`",
+            );
+            BigInt { sign, data }
+        }
+        pub fn get(self) -> num_bigint::BigInt {
```
franziskuskiefer, line 22:
```suggestion

        pub fn get(self) -> num_bigint::BigInt {
```

hax-lib/src/int/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+
+impl Int {
+    /// Raises `2` at the power `self`
+    pub fn pow2(self) -> Self {
+        let exponent = self.get().to_u32().expect("Exponent doesn't fit in a u32");
+        Self::new(num_bigint::BigInt::from(2u8).pow(exponent))
+    }
+
+    /// Constructs a `Int` out of a string literal. This function
+    /// assumes its argument consists only of digits, with optionally
```
franziskuskiefer, line 64: What type of digits? Hex?

hax-lib/src/int/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+    type Output = Self;
+
+    fn div(self, other: Self) -> Self::Output {
+        Self::new(self.get() / other.get())
+    }
+}
+
+impl Int {
+    /// Raises `2` at the power `self`
+    pub fn pow2(self) -> Self {
```
franziskuskiefer, line 58: Do we need this function anywhere?

hax-lib/src/lib.rs
```diff
@@ -15,6 +15,8 @@
 
 #![no_std]
 
+pub mod int;
```
franziskuskiefer, line 18: How is this compiled out when not in `cfg(hax)`?

franziskuskiefer (changes requested):
Thanks, a couple notes.

hax-lib/src/int/mod.rs
W95Psp: Yes, we use that a lot for Kyber and its spec

hax-lib/Cargo.toml
W95Psp: yeah, sorry, was fighting with CI, fixing that

hax-lib/src/lib.rs
W95Psp: Postponing that to another PR, #639

hax-lib-macros/src/lib.rs
W95Psp: Postponing that to another PR, #639

franziskuskiefer (approved):

engine/lib/phases/phase_specialize.ml
```diff
@@ -0,0 +1,165 @@
...
+
+          (* let is_int = is_type0 Hax_lib__int__Int *)
+          (* let is_ref_int = is_ref_type0 Hax_lib__int__Int *)
+
+          (* let mk_ref_int n f f' = *)
+          (*   { *)
+          (*     fn = f; *)
+          (*     fn_replace = f'; *)
+          (*     args = List.init n ~f:(fun _ -> is_ref_int); *)
+          (*   } *)
```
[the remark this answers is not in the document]
W95Psp: Oh, thanks, I re-wrote the logic of this pattern thing, forgot to clean up!
I cleaned up and added some documentation there.

karthikbhargavan (approved):

PR #641, opened 2024-04-29
add F* core arch type

franziskuskiefer:
We need some core types. Am I doing this right?

### DIFF
--- a/proof-libs/fstar/core/Core.Core_arch.fsti
+++ b/proof-libs/fstar/core/Core.Core_arch.fsti
@@ -1,0 +1,3 @@
+module Core.Core_arch
+
+include Core.Core_arch.X86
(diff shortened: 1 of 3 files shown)


proof-libs/fstar/core/Core.Core_arch.fsti
```diff
@@ -0,0 +1,3 @@
+module Core.Core_arch
+
+include Core.Core_arch.X86
```
W95Psp, line 3: I don't think we need this `include` here: generated Rust will never have `Core.Core_arch.t___m256i`, always `Core.Core_arch.X86.t___m256i`.
```suggestion
```

proof-libs/fstar/core/Core.fst
```diff
@@ -4,3 +4,4 @@ include Rust_primitives
 include Core.Num
 include Core.Iter
 include Core.Ops
+include Core.Core_arch
```
W95Psp, line 7: Same here, no include is required.
It's nice to `include` things only for definitions we deal with in a special way (e.g. the operator `+.` should always be in scope for every module)

```suggestion
```

PR #651, opened 2024-05-06
feat(engine): allow cast of closures as fn pointers

### LINKED ISSUE #500: Cast of closure as Fn pointers
```rust
fn dummy() {
  let f: fn(&u32) -> &u32 = |x| x;
}
```
is rejected with:
```
error[HAX0001]: (AST import) something is not implemented yet.
                Pointer, with [cast] being (Types.ClosureFnPointer Types.Normal)
  --> traits/src/lib.rs:56:27
   |
56 |     let f: fn(&T) -> &T = |x| x;
```

This is trivial to support: in hax' AST, there is no difference between closures and function pointers, we should just ignore the cast

W95Psp:
Fixes #500

### DIFF
--- a/engine/lib/import_thir.ml
+++ b/engine/lib/import_thir.ml
@@ -865,7 +865,7 @@ end) : EXPR = struct
 
   and c_pointer e typ span cast source =
     match cast with
-    | ReifyFnPointer ->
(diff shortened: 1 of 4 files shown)


franziskuskiefer (approved):
Thanks

PR #654, opened 2024-05-07
fix(ci): remove `specs` tests

W95Psp:
The https://github.com/hacspec/specs repository contains specifications written in hacspec, mainly at the time of hacspec v1.

Since hacspec v1 is deprecated and archived, its library is deprecated as well.

The hacspec lib (https://github.com/hacspec/hacspec/tree/master/lib) doesn't typecheck any longer with `num-bigint` version `0.4.5`.

The hacspec lib dependents on `num-bigint` version `0.4`, not specifying the minor version. The `specs` repository has no `Cargo.lock`, and thus, from a clean `git clone`, `cargo build` pins `num-bigint` to version `0.4.5`, breaking the hacspec library.

This broke https://github.com/hacspec/hax/pull/653 recently, and we had the same kind of errors a couple times before in the past.

Two options were considered:
 - remove the tests related to the specs in the CI of hax;
 - commit a `Cargo.lock`.

The choose the first one: we need to update those specifications to hax anyway: this will add motivation to
https://github.com/hacspec/specs/issues/13.

### DIFF
--- a/.github/workflows/install_and_test.yml
+++ b/.github/workflows/install_and_test.yml
@@ -41,33 +41,6 @@ jobs:
       with:
         repository: 'hacspec/specs'
         path: specs
-        
-    - name: Extract specifications
-      working-directory: specs
-      run: |
-        paths=$(tomlq -r '.workspace.members | .[]' Cargo.toml)
-        for cratePath in $paths; do
-          crate=$(tomlq -r '.package.name' "$cratePath/Cargo.toml")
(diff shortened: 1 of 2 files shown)


franziskuskiefer (approved):

karthikbhargavan (approved):
This is needed to make progress, but we should add an issue with "help wanted" for people who want to upgrade this to use hax instead.

Issue #495, opened 2024-02-05
Frontend crash: escaping bound vars


W95Psp:
hax crashes with

```
thread 'rustc' panicked at '`ProjectionPredicate(AliasTy { substs: [P, (&<I as std::iter::Iterator>::Item,)], def_id: DefId(2:2931 ~ core[02ea]::ops::function::FnOnce::Output) }, Term::Ty(bool))` has escaping bound vars, so it cannot be wrapped in a dummy binder.', /rustc/d59363ad0b6391b7fc5bbb02c9ccf9300eef3753/compiler/rustc_middle/src/ty/mod.rs:1195:26
```

on the following code
```rust
fn iterating(list: Vec<u8>) {
    let _indices: Vec<_> = (0..5).filter(|i| list.iter().any(|n| n == i)).collect();
}
```

_Originally posted by @franziskuskiefer in https://github.com/hacspec/hax/issues/461#issuecomment-1923340633_

W95Psp:
I investigated a bit: I think that comes from hax treating `Clause` as the base brick for manipulating clauses, but `Clause`s might lack bound vars.

There are two options:
 - we may want to always deal with `GenericBounds` instead of `Clause`s
 - we may want to build Clauses directly from `GenericBounds`, computing ids directly from `GenericBounds`

Nadrieril:
Isn't this what we discussed last week, and the solution was proper handling of `Binder`?

W95Psp:
well, indeed it is related, but that's a bit different

But yeah basically we need to go full `PolyChooseYourType` :)

Should be basically about refactoring some stuff, I'm investigating what exactly should be refactored

PR #655, opened 2024-05-07
Bump rustc version

Nadrieril:
Some interesting clause-related changes.

Regarding `Clause` hashing, I think we should either never hash the binder or prevent hashing without a binder; the current middle-ground is error-prone. Consistently ignoring the binder seems fine.

### DIFF
--- a/frontend/exporter/src/types/copied.rs
+++ b/frontend/exporter/src/types/copied.rs
@@ -1,5 +1,6 @@
 use crate::prelude::*;
 use crate::rustc_middle::query::Key;
+use rustc_middle::ty;
 
 /// Reflects [`rustc_hir::definitions::DisambiguatedDefPathData`]
 #[derive(
@@ -2784,7 +2785,6 @@ pub struct Impl<Body: IsBody> {
                 .predicates
                 .iter()
                 .copied()
-                .filter_map(|(pred, span)| Some((pred.as_clause()?, span)))
                 .filter_map(|(clause, span)| super_clause_to_clause_and_impl_expr(s, owner_id, clause, span))
                 .collect::<Vec<_>>()
         } else {
@@ -3272,6 +3272,7 @@ pub enum ClauseKind {
     ConstArgHasType(ConstantExpr, Ty),
     WellFormed(GenericArg),
     ConstEvaluatable(ConstantExpr),
+    TypeWellFormedFromEnv(Ty),
 }
 
 /// Reflects [`rustc_middle::ty::ClauseKind`] and adds a hash-consed clause identifier.
@@ -3329,12 +3330,14 @@ pub struct Binder<T> {
 
 /// Reflects [`rustc_middle::ty::GenericPredicates`]
 #[derive(AdtInto)]
-#[args(<'tcx, S: UnderOwnerState<'tcx>>, from: rustc_middle::ty::GenericPredicates<'tcx>, state: S as tcx)]
+#[args(<'tcx, S: UnderOwnerState<'tcx>>, from: rustc_middle::ty::GenericPredicates<'tcx>, state: S as s)]
 #[derive(
     Clone, Debug, Serialize, Deserialize, JsonSchema, Hash, PartialEq, Eq, PartialOrd, Ord,
 )]
 pub struct GenericPredicates {
     pub parent: Option<DefId>,
+    // FIXME: Switch from `Predicate` to `Clause` (will require correct handling of binders).
+    #[value(self.predicates.iter().map(|(clause, span)| (clause.as_predicate().sinto(s), span.sinto(s))).collect())]
     pub predicates: Vec<(Predicate, Span)>,
 }
 
@@ -3418,7 +3421,6 @@ pub enum PredicateKind {
     Subtype(SubtypePredicate),
     Coerce(CoercePredicate),
     ConstEquate(ConstantExpr, ConstantExpr),
-    TypeWellFormedFromEnv(Ty),
     Ambiguous,
     AliasRelate(Term, Term, AliasRelationDirection),
 }
@@ -3452,22 +3454,25 @@ fn region_bounds_at_current_owner<'tcx, S: UnderOwnerState<'tcx>>(s: &S) -> Gene
         }
     };
 
-    let predicates: Vec<_> = if use_item_bounds {
+    let clauses: Vec<ty::Clause<'tcx>> = if use_item_bounds {
         tcx.item_bounds(s.owner_id())
             .subst_identity()
             .iter()
-            .map(|x| (x.as_predicate(), rustc_span::DUMMY_SP))
             .collect()
     } else {
         tcx.predicates_defined_on(s.owner_id())
             .predicates
-            .into_iter()
-            .cloned()
+            .iter()
+            .map(|(x, _span)| x)
+            .copied()
             .collect()
     };
-    predicates
+    clauses
         .iter()
-        .map(|(pred, _span)| tcx.erase_late_bound_regions(pred.clone().kind()).sinto(s))
+        .map(|clause| {
+            tcx.erase_late_bound_regions(clause.as_predicate().kind())
+                .sinto(s)
+        })
         .collect()
 }
 
--- a/frontend/exporter/src/types/copied.rs
+++ b/frontend/exporter/src/types/copied.rs
@@ -1960,8 +1960,19 @@ pub enum PointerCast {
 pub enum BorrowKind {
     Shared,
     Shallow,
-    Unique,
-    Mut { allow_two_phase_borrow: bool },
+    Mut { kind: MutBorrowKind },
+}
+
+/// Reflects [`rustc_middle::mir::MutBorrowKind`]
+#[derive(AdtInto)]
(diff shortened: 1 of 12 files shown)


W95Psp:
Cool, thanks, will review on Monday!
There is a to_predicate that happens in the hashing, which calls Binder::dummy, that checks if there are any unbound variable, and this can cause panics :(
So that's why I think we should always hash with binders, but I agree that the current situation is pretty bad! But that's not directly related to that PR :)

Nadrieril:
With this PR we no longer call to_predicate before hashing Clauses

W95Psp:
Ah, I had no time to look at the code, that's great then! That's simplifies things
Might help with #495 then

Nadrieril:
There are a few other places where we take shortcuts with binders, we'll have to go through these too

frontend/exporter/src/types/copied.rs
```diff
@@ -3302,12 +3330,14 @@ pub struct Binder<T> {
...
 #[derive(AdtInto)]
-#[args(<'tcx, S: UnderOwnerState<'tcx>>, from: rustc_middle::ty::GenericPredicates<'tcx>, state: S as tcx)]
+#[args(<'tcx, S: UnderOwnerState<'tcx>>, from: rustc_middle::ty::GenericPredicates<'tcx>, state: S as s)]
 #[derive(
     Clone, Debug, Serialize, Deserialize, JsonSchema, Hash, PartialEq, Eq, PartialOrd, Ord,
 )]
 pub struct GenericPredicates {
     pub parent: Option<DefId>,
+    // FIXME: Switch from `Predicate` to `Clause` (will require correct handling of binders).
+    #[value(self.predicates.iter().map(|(clause, span)| (clause.as_predicate().sinto(s), span.sinto(s))).collect())]
```
W95Psp, line 3340: I'm not sure I understand exactly the renaming done in Rustc.
`rustc_middle::ty::Predicate` and `rustc_middle::ty::Clause` are wrapping exactly the same pieces of information, right? from what I understand it's just a nominal difference, and the two types offer two different sets of methods
But that split seems less relevant in hax, since we provide exactly the same data for those two

W95Psp: But then, I don't understand your FIXME comment: both rustc_middle::ty::Predicate and rustc_middle::ty::Clause contains binders of PredicateKind, there should be no difference in term of binders handling

frontend/exporter/src/types/copied.rs
```diff
@@ -3269,16 +3284,29 @@ pub struct Clause {
...
     }
 }
 
+impl<'tcx, S: UnderOwnerState<'tcx>> SInto<S, Clause> for rustc_middle::ty::ClauseKind<'tcx> {
+    fn sinto(&self, s: &S) -> Clause {
+        // FIXME: this is dangerous.
+        rustc_middle::ty::Binder::dummy(self.clone()).sinto(s)
+    }
+}
+
```
W95Psp, line 3309: I think this is OK for this PR, but then we should simplify those things, the predicate indirection is useless now (I'm not sure wether this was useful at any point now :sweat_smile:) and requires wrapping with dummy binders, which indeed is quite dangerous.

W95Psp (approved):
Thank you! :grinning:

Seems OK to me, one thing we should settle on is a vocabulary between clauses and predicates: those things now contain exactly the same pieces of information (if I understand correctly)
We should probably map both Predicate and Clause (from rustc) to Clause (from hax), wdyt?

frontend/exporter/src/types/copied.rs
Nadrieril: To be clear: this PR does not introduce any new `Binder::dummy`, these were moved from somewhere else. I agree we should take the time to remove them at some point.

frontend/exporter/src/types/copied.rs
Nadrieril: I mean, we could just use `Predicate` everywhere, but at the cost of a lot of `=> unreachable!()`.

W95Psp: No, what I mean is that there is both `rustc_middle::ty::Predicate` and `rustc_middle::ty::Clause`, which are both newtypes around:

 [Interned](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_data_structures/intern/struct.Interned.html)<'tcx, [WithCachedTypeInfo](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.WithCachedTypeInfo.html)<[Binder](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/sty/struct.Binder.html)<'tcx, [PredicateKind](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/predicate/type.PredicateKind.html)<'tcx>>>>)

So, this distinction between `Clause` and `Predicate` is purely nominal.

Then, `ClauseKind`, called `Clause` in hax, is indeed a subcase of `PredicateKind`, which itself is a field of `Predicate`.

My point here is that the `clause` in `self.predicates.iter().map(|(clause, span)| (clause.as_predicate(). ....` is of type `rust_middle::ty::Clause`, which is basically exactly a `hax::Predicate`, that is a PredicateKind wrapped into a Binder. So for me, that's not a FIXME, we should just derive a `sinto` for `Clause` that calls `as_predicate` and `sinto` on the result.

frontend/exporter/src/types/copied.rs
W95Psp: Well, I meant the "PolyClauseKind" actually, not the `Predicate` (which is a bit a "PolyPredicateKind" ^^)

frontend/exporter/src/types/copied.rs
Nadrieril: It's not just nominal, a `Clause` is guaranteed to correspond to a predicate with kind `PredicateKind::Clause`, which is useful because that's fewer cases to handle. Would you prefer we lose this information and use `Predicate` everywhere?

frontend/exporter/src/types/copied.rs
Nadrieril: I'm not sure what you mean, but we should definitely be more careful with binders

frontend/exporter/src/types/copied.rs
W95Psp: Wait, there is *no* formal guarantee about the content of a Clause, there isn't even a smart constructor that would reject a `Clause` that contains something else than a `PredicateKind::Clause`.
The only thing we get is the documentation saying "a subset of predicates which can be assumed by the trait solver"

Do you then mean that we should reflect this informal invariant and define a type `type Clause = Binder<UnboundedClause>`, renaming the current `hax::Clause` into `UnboundedClause`?

W95Psp: (well, the API is also quite toward that invariant, since `kind` returns indeed a `Binder<ClauseKind>` -- so having this `Clause` defined as a `Binder<UnboundeClause>` would make sense)

Nadrieril: What do you mean? The only way to construct a `Clause` is with `Predicate::as_clause` isn't it?

Nadrieril: The field of `struct Clause(..)` isn't public

Nadrieril: > Do you then mean that we should reflect this informal invariant and define a type type Clause = Binder<UnboundedClause>, renaming the current hax::Clause into UnboundedClause?

Yep!

W95Psp: Yes, you are right, the invariant was not super clear to me, but that's ok indeed.

Alright, then I guess we should just always produce `Binder<ClauseKind>` clauses, never `ClauseKind` by itself; so we should probably move from `Clause` being defined as `ClauseKind + clause id` to `Clause = {kind: Binder<ClauseKind>, id: ClauseId}`. I'm gonna file an issue for that now

Nadrieril: The problem with that is that we don't have access to the binder when recursively converting `PredicateKind`

Nadrieril: The thing I had in mind was:
```rust
enum PredicateKind {
  Clause(UnboundClause),
  ...
}
```
with an API to get from `Predicate` to `Clause` by mapping the binder over. Maybe there's a simpler option

W95Psp: https://github.com/hacspec/hax/issues/661

Nadrieril: let's discuss there

W95Psp: Resolving this conversation, #661 is a future work

frontend/exporter/src/types/copied.rs
W95Psp: Yes, let's see that later anyway

W95Psp:
Let's merge whenever you want @Nadrieril (do you have permissions to press merge though?)

Nadrieril:
I don't have permissions to merge, you can go ahead

PR #656, opened 2024-05-08
prepare hax-lib for publishing

franziskuskiefer:
This sets some more values for hax-lib crates and updates publishing.md.
When this is in let's publish a first round.

### DIFF
--- a/hax-lib-macros/types/README.md
+++ b/hax-lib-macros/types/README.md
@@ -1,3 +1,3 @@
 # hax internal types
 
-A create with hax internal types.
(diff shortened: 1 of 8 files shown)


hax-lib-macros/types/README.md
```diff
@@ -0,0 +1,3 @@
+# hax internal types
+
+A create with hax internal types.
```
W95Psp, line 3: I would rephrase like that, wdyt?
```suggestion
A crate that defines the types of the various payloads of the attributes produced by the crate `hax-lib-macros` and consumed internally by the engine of hax.
```

W95Psp (approved):
Thanks, looks good to me, just one thing

Issue #661, opened 2024-05-13
exporter: refactor `Clause`s


W95Psp:
While defining `Clause`s I was a bit aggressive towards binders, and entirely got rid of them. This is a bad design, as pointed out by @Nadrieril in our conversation in #655.

Currently, a `Clause` is a `ClauseKind` + an identifier.

We should instead have a `Clause` being a struct with first a `Binder<ClauseKind>` and then a `ClauseId`, which accounts for binders.

Nadrieril:
This may cause issues because we don't have access to the binder when recursively converting `PredicateKind`. The thing I had in mind was:

```rust
enum PredicateKind {
  Clause(UnboundClause),
  ...
}
```

with an API to get from Predicate to Clause by mapping the binder over like rustc does. Maybe there's a simpler option.

W95Psp:
Yes, that's a good point.

To recap a bit, the issue is that if we go with a `Clause` object that contains a `Binder<ClauseKind>` and a clause identifier, then no clause identifier will be accessible from a `Predicate`.

If we confirm we should always work with `Binder<ClauseKind>`s and `Binder<PredicateKind>`s and never work with `ClauseKind`s and `PredicateKind`s, then I think we should really have the identifier above the `Binder<...>`.

Currently `Predicate` is defined as `Binder<PredicateKind>`, and we are discussing about adding `Clause` defined as `Binder<PredicateKind>`.

What about defining a `Predicate` (resp. `Clause`) as structs with both a field `kind` of type `PredicateKind` (resp. `ClauseKind`) and a field `clause_id` of type `ClauseId` (using a newtype as you suggest in #626)? (we could also define only one struct for those two with a generic, but that would be more confusing than anything else)

W95Psp:
Regarding our conversation in your PR @Nadrieril, I'm not crazy: there was a renaming from Clause to ClauseKind in Rustc... Clauses used not to have binders at all!

In commit [fca56a8], the type `Clause` has been renamed as `ClauseKind`, then commit [21226e] makes `Clause` a `Binder<ClauseKind>`, essentially (with an indirection via `PredicateKind` and an invariant where the
predicate kind has to be a `ty::PredicateKind::Clause(ClauseKind)`).

[fca56a8]: https://github.com/rust-lang/rust/commit/fca56a8d2c6f803aeec51167dd4dd22529f8bdd7
[21226e]: https://github.com/rust-lang/rust/commit/21226eefb2e2465b9730c37e7d08655865610d90

PR #662, opened 2024-05-13
Refactor of clauses

W95Psp:
This PR implements the changes discussed in #661.


frontend/exporter/src/types/new.rs
```diff
@@ -70,3 +70,63 @@ impl ItemAttributes {
...
+/// constructing by hashing it.
+pub struct PredicateId(u64);
+
+/// `IntoPredicateId` provides a method `clause_id` that dervies a clause
+/// identifier by hashing `self`.
+pub trait IntoPredicateId<'tcx, S: UnderOwnerState<'tcx>> {
+    fn clause_id(&self, s: &S) -> PredicateId;
+}
+
+const _: () = {
```
Nadrieril, line 88: A module would be more idimatic than an anonymous constant here

Nadrieril:
This is quite subtle because we need to make sure to either always hash the clause with its binder or never do that. Could you split this in a few commits that each do a single thing (e.g. define `PredicateId` and its module first, and modify behavior separately)

W95Psp:
Yes, I'm gonna rebase this nicely, no worries!

Nadrieril:
No problem, let me know when you have a version I can look at

Nadrieril (approved):

PR #671, opened 2024-05-15
CI: test Charon

pnmadelaine:


Nadrieril:
Of course this is broken because of rustc version mismatch x)

W95Psp (approved):
Thanks PNM :smiley: Feels great to see how concise the action is thanks to Nix :smile: 

@Nadrieril, yeah, that makes sense, no problem, that will be green later

PR #672, opened 2024-05-15
Skip inner obligations of `ImplSource::Builtin`

Nadrieril:
As they can lead to infinite recursion on e.g. this:
```rust
fn builtin_bound<M: Send>(_msg: M) {}
fn use_builtin_bound() {
    builtin_bound(());
}
```

I tried to add a test in your test suite but there was no bug on `hax` itself; it seems this was only triggered in charon.


W95Psp:
Looks good to me, thanks!

W95Psp (approved):

PR #673, opened 2024-05-15
Delete .github/workflows/add_to_project.yml

W95Psp:
We don't use that board any longer, so let's disable this job (as we discussed @franziskuskiefer)


franziskuskiefer (approved):
Thanks

PR #675, opened 2024-05-15
Frontend: silence an error about escaping bounded variables

### LINKED ISSUE #495: Frontend crash: escaping bound vars

W95Psp:
This PR:
  - fixes #495 by turning an error into a warning with some debugging context;
  - fixes another similar bug when a bound is universally quantified by a lifetime (See the tests introduced by the first commit. When comparing two predicates, we were not systematically erasing regions, causing a bound `for<'a> Foo<&'a T>` not being selected when looking for a `Foo<&T>`.).

### DIFF
--- a/frontend/exporter/src/traits.rs
+++ b/frontend/exporter/src/traits.rs
@@ -299,15 +299,6 @@ pub trait IntoImplExpr<'tcx> {
     ) -> ImplExpr;
 }
 
-impl<'tcx> IntoImplExpr<'tcx> for rustc_middle::ty::TraitRef<'tcx> {
-    fn impl_expr<S: UnderOwnerState<'tcx>>(
-        &self,
-        s: &S,
-        param_env: rustc_middle::ty::ParamEnv<'tcx>,
-    ) -> ImplExpr {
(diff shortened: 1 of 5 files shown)


franziskuskiefer (approved):
thanks

PR #682, opened 2024-05-21
feat: add a logo

W95Psp:
On Friday I experimented a bit with Inkscape to make a logo for hax, any thoughts? :smile: 

.

<p align="center">

</p>

.

[View readme](https://github.com/hacspec/hax/blob/logo/README.md)

Note: I updated the logo so that we see a turnstile, and I changed a bit the colors


W95Psp:
Cleaned up and rebased this PR, now the logo should be properly aligned!

franziskuskiefer (approved):
We need to get the X fixed. But let's merge this version.

PR #685, opened 2024-05-22
refactor(exporter): always use `State::param_env`

W95Psp:
This PR makes the exporter always use `s.param_env()` instead of `s.base().tcx.param_env(s.owner_id())`.

Status: any merge in hax is blocked by https://github.com/hacspec/hax/pull/679, I'm fixing that PR


franziskuskiefer (approved):
Thanks

PR #691, opened 2024-05-29
Update README.md

spitters:


W95Psp (approved):
Thanks, some duplication somehow showed up in there!

PR #694, opened 2024-06-03
Update to OCaml 5

W95Psp:
This PR updates the engine to OCaml 5

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Note:
 
 1. Make sure to have the following installed on your system:
 
-- [`opam`](https://opam.ocaml.org/) (`opam switch create 4.14.1`)
+- [`opam`](https://opam.ocaml.org/) (`opam switch create 5.1.1`)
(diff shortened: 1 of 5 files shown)


franziskuskiefer:
Do we really want to do this when all our other tools don't? I'm of course all for updating, but I want to avoid incompatibilities.

W95Psp:
Mmh, I was not thinking about that indeed.
But the engine of hax is not used as an OCaml library from any other projects, and it doesn't consume any other projects as an OCaml library.
So, since the projects compose only at the binary level, I think there won't be any incompatibilities.
Does that make sense or I'm missing a source of incompatibility?

franziskuskiefer (approved):
Ok, let's try it and see what happens.

Where's that Rust change coming from? Looks like formatting?

W95Psp:
Yes, the Rust change is just about formatting. I think that's something rustfmt used not to format or something like that.

PR #698, opened 2024-06-03
Generate visitors automatically

W95Psp:
This PR:
 - drops the manually implemented visitors, which are hard to maintain and error-prone;
 - adds a OCaml utility binary `generate_visitors` that derive those visitors reading the ast.ml module;
 - adds a dune rule to generate the module `ast_visitors` using `generate_visitors`
 
Note: the CI is failing for Charon, but that's likely because of rustc updates


engine/utils/generate_visitors/README.md
```diff
@@ -0,0 +1,14 @@
+# `generate_visitors`
+
+This binary reads the AST module of hax and creates **standalone**
```
franziskuskiefer, line 3: This certainly needs to go into the top level dev.md.

franziskuskiefer (changes requested):
I'm not sure I understand this PR.

You generate the visitor code but then also commit it? What is the intended workflow here?

W95Psp:
No, I commit small visitors for the types used by the visitor generator :sweat_smile: so yes, that's particularly confusing.
The visitor generator is designed for the types in the module Ast, it doesn't solve the general problem of generating visitors, which is more complicated. But we don't need a general solution, whence that small OCaml utility.

That utility defines a small AST for representing types, and I needed visitors for that small AST, which I generated using the tool itself, but since it's that AST is a bit different, I had to do manual edits. That's why I commit that `visitors.ml` file.
The visitors of the engine itself are now gone and fully generated.

engine/utils/generate_visitors/README.md
W95Psp: I added some more high level documentation in DEV.md!

engine/DEV.md
```diff
@@ -42,6 +42,12 @@ To show the file nicely formated, use: `dune describe pp lib/types.ml` (or `dune
 
 You can also use `dune utop` and then `#show_type Hax_engine.Types.SOME_TYPE` and `#show_constructor Hax_engine.Types.SOME_CONSTRUCTOR`.
 
+### Visitors
+The module `Ast_visitors` provides a `Make` functor that takes a feature set and outputs a module that defines the `map`, `mapreduce` and `reduce` classes.
+
+Those visitors are generated by `./utils/generate_visitors`. 
+To see the implementation of the `Ast_visitors` module, run `dune describe pp lib/ast_visitors.ml`.
```
franziskuskiefer, line 49: Running this I get
```
File "utils/generate_visitors/errors.ml", line 27, characters 29-35:
27 |     | UnsupportedCoreType of string
                                  ^^^^^^
Error: Unbound value string_of_yojson
```

W95Psp: Thanks for the feedback, that's odd. Might be a library version or something, but anyway, I just pushed 3887a9f, can you check if that's fixing it?

W95Psp: Also, do you get this error when building regularly?

franziskuskiefer: Yes, this fails for the regular build as well.

engine/utils/generate_visitors/codegen_visitor.ml
```diff
@@ -0,0 +1,250 @@
+(** Give a list of {!Types.Datatype.t}, this file generates an ocaml
+module of visitors. *)
+
+open Base
```
franziskuskiefer, line 4: All these files need comments.

franziskuskiefer (approved):
Please add comments to all the new code before merging.

engine/DEV.md
W95Psp: Ok! But the patch makes things work, right?

engine/utils/generate_visitors/codegen_visitor.ml
W95Psp: I added loads of comments

PR #703, opened 2024-06-06
feat: isolate `DefId`: faster build times

W95Psp:
The OCaml build depends on the binary `hax-engine-names-extract`, which, at build-time, depends on the hax frontend.
`hax-engine-names-extract` runs the frontend at build time on `hax-engine-names`, which itself used to depend on the crate `hax-engine-frontend`. Running hax however sets specific cfg flags, yielding recompilations.

This PR cut the dependency between `hax-engine-names` and `hax-engine-frontend`. `hax-engine-names` needs only three type definitions from `hax-engine-frontend`: this PR isolates those three types in a module, and  makes `hax-engine-names` directly use that module, inlined.

### DIFF
--- a/engine/names/extract/build.rs
+++ b/engine/names/extract/build.rs
@@ -1,7 +1,15 @@
-use hax_frontend_exporter::{DefId, DefPathItem, DisambiguatedDefPathItem};
 use serde_json;
 use serde_json::Value;
 use std::process::{Command, Stdio};
+
+/// Instead of depending on `hax_frontend_exporter` (that links to
+/// rustc and exposes a huge number of type definitions and their
+/// impls), we just inline a small module here that contains the three
+/// type definition we need. See the module for complementary
+/// informations.
+#[path = "../../../frontend/exporter/src/types/def_id.rs"]
+mod hax_frontend_exporter_def_id;
+use hax_frontend_exporter_def_id::*;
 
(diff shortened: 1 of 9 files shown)


engine/names/extract/build.rs
```diff
@@ -1,8 +1,16 @@
...
 use serde_json;
 use serde_json::Value;
 use std::process::{Command, Stdio};
 
+/// Instead of depending on `hax_frontend_exporter` (that links to
+/// rustc and exposes a huge number of type definitions and their
+/// impls), we just inline a small module here that contains the three
+/// type definition we need. See the module for complementary
+/// informations.
+#[path = "../../../frontend/exporter/src/types/def_id.rs"]
```
franziskuskiefer, line 10: I'm not actually sure if this is publishable.

franziskuskiefer (commented):
💯 for fewer dependencies, esp. this one. But I'm not sure if it actually works.
Do we need to publish this one? I don't see it anywhere in publishing.md. But I'm not sure if you can publish a crate with a `path` attribute.

engine/names/extract/build.rs
W95Psp: Ah, that's a legitimate question!
But this one is not supposed to be published, it's solely for the ocaml build

PR #704, opened 2024-06-06
fix: rename action for extracting ml-kem

W95Psp:
This PR just renames the action that extracts ML-Kem from `hax` to `extract-mlkem`. Having this workflow named `hax` was very confusing (`git blame` blames myself sadly 😅).

Note: when merging this PR, we'll need to change the required jobs for merging in `main`

### DIFF
--- a/.github/workflows/mlkem.yml
+++ b/.github/workflows/mlkem.yml
@@ -1,4 +1,4 @@
-name: Extract and TC Kyber
+name: Extract and TC ML-Kem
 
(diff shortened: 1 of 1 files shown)


franziskuskiefer (approved):
👍🏻  while you change required CI jobs you may also want to switch to the new rules as we need to do that at some point anyway.

W95Psp:
What do you mean by new rules?

PR #709, opened 2024-06-10
fix(exporter): disable impl expr resolution under type aliases

### LINKED ISSUE #707: Trait selection returns `Unimplemented` on type aliases
The frontend fails with `Cannot handle error `Unimplemented` selecting ...` on:

```rust
struct Foo<T: Clone>(T);
type Synonym<T> = Foo<T>;
```

The bug is that Rust doesn't enforce bounds on generic parameters in type aliases.
So writing `type Synonym<T> = Foo<T>;` is legal even if `T` has no bound `Clone`, but hax tries to export an impl source anyway! Obviously, without `T: Clone`, trait selection returns `Unimplemented`.

The solution is not to export impl expressions on type synonyms

Note the following works fine:

```rust
struct Foo<T: Clone>(T);
type Synonym<T: Clone> = Foo<T>;
```

### LINKED ISSUE #708: Type aliases missing bounds in backends
Related to https://github.com/hacspec/hax/pull/709 and https://github.com/hacspec/hax/issues/707.

Rust doesn't enforce bounds on type synonyms, thus #709 kills trait resolution for trait aliases.
But that means type synonyms will be extracted wrongly in the backends due to that missing trait information.

That's another hint that relying on Rust type synonyms is bad. Those are really to be seen as some kinds of macros / early sugar.

W95Psp:
This commit disables trait resolution in the context of a type alias definition.

Rust doesn't enforce bounds on generic parameters in type aliases, so it is legal to write:

```rust
struct Foo<T: Clone>(T);
type Synonym<T> = Foo<T>;
```

Thus, if we compute the impl expression for the bound `T: Clone` that appears on the type expression `Foo<T>`, Rustc's trait selection fails: in the context of `Synonym`, `T` indeed doesn't implement `Clone`, _a priori_.

This is raising an issue about backends: translating such type synonyms with "missing" bounds will fail. See
https://github.com/hacspec/hax/issues/708. Thus, this PR also introduces an option so that type synonyms are not extracted by default.

Fixes #707 and fixes #708

### DIFF
--- a/cli/options/src/lib.rs
+++ b/cli/options/src/lib.rs
@@ -313,6 +313,16 @@ pub struct BackendOptions {
     #[arg(short, long = "debug-engine")]
     pub debug_engine: Option<DebugEngineMode>,
 
+    /// Extract type aliases. This is disabled by default, since
+    /// extracted terms depends on expanded types rather than on type
+    /// aliases. Turning this option on is discouraged: Rust type
+    /// synonyms can ommit generic bounds, which are ususally
+    /// necessary in the hax backends, leading to typechecking
+    /// errors. For more details see
+    /// https://github.com/hacspec/hax/issues/708.
+    #[arg(long)]
+    pub extract_type_aliases: bool,
+
     #[command(flatten)]
     pub translation_options: TranslationOptions,
 }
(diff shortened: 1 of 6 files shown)


Nadrieril:
Could you add the required bounds to the type synonym instead? This is the only correct solution I can imagine.

W95Psp:
That something we could do, indeed, but that's not trivial. I'm afraid this will cause bugs in the frontend :/

A quite "brutal" way of doing that would be: each time `select_trait_candidate` fails on a trait reference, we add this trait reference to the where clause of the type synonym. But that sounds quite hacky :(

Nadrieril:
Yeah :(

franziskuskiefer (approved):

PR #710, opened 2024-06-10
feat(hax): logging: enable tracing in release, add trait-related logs

W95Psp:
This PR makes hax built with release mode ship with all logs available.

It also adds some logs on trait functions.


franziskuskiefer (approved):
🪵

PR #713, opened 2024-06-12
refactor: feature gate all implementations with `full`

W95Psp:
This PR adds a (default) feature `full` for the
`hax-frontend-exporter` crate. With this feature disabled, the crate turns into a mostly type-only crate. The crate depends on rustc and drive `sinto` only with `full` enabled.

This PR also simplifies some the structure of the workspace:
 - remove the `path` trick introduced by #703 in favor of a dependency to `hax-frontend-exporter` where feature `full` is off 
 - remove the crate `hax-frontend-exporter-options`, now it's possible to inline it in `hax-frontend-exporter`

This PR is particularly useful for https://github.com/hacspec/hax/pull/714


frontend/exporter/Cargo.toml
```diff
@@ -13,13 +13,18 @@ description = "Provides mirrors of the algebraic data types used in the Rust com
...
+
+tracing = { workspace = true, optional = true }
+itertools = { workspace = true, optional = true }
+extension-traits = { version = "1.0.1", optional = true }
+lazy_static = { version = "1.4.0", optional = true }
+cfg-if = "1.0.0"
+
+[features]
+default = ["full"]
+full = ["dep:tracing", "dep:itertools", "dep:extension-traits", "dep:lazy_static"]
```
franziskuskiefer, line 30: What exactly is this feature? Full isn't a great name. Something a little more descriptive would be nice. Adding some comments here to describe what it enables would be good too.

W95Psp:
Will be subsumed by https://github.com/hacspec/hax/pull/743

PR #714, opened 2024-06-12
feat(cli): add a `list-names` verb, accept "raw" paths

W95Psp:
This commit:
 - adds a `cargo hax list-names` commands that outputs the list of all external names used by a crate. Note that outputs "raw" paths.
 - make the `--include-namespace` option accept "raw" paths.


W95Psp:
Closing this, we have to implement it a nicer way, PR #743 should help.
Tracking issue: https://github.com/hacspec/hax/issues/767

PR #718, opened 2024-06-13
Fstar val instances

W95Psp:
This PR makes instances interfaces opaque, so that extracting the interfaces for a crate extracts zero code (but constants!).

This is still in draft because that's not what we always want.


W95Psp:
Tracking issue: https://github.com/hacspec/hax/issues/768.
Closing this one for now, we need to implement that in a nicer way.

PR #721, opened 2024-06-14
Update rustc

Nadrieril:
This pushes the rustc pin two months forward.

Something weird to note: in the first commit, a bunch of std functions gain a new invisible const generics argument. A later commit reverts this. But this happens again for nightlies in December. I stopped mid-November to give me time to investigate what's happening there.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,7 +65,7 @@ syn = { version = "1.0.107", features = [
 ] }
 tracing = { version = "0.1", features = [
      "max_level_trace",
-     "release_max_level_warn",
+     "release_max_level_trace",
 ] }
 tracing-subscriber = { version = "0.3", features = [
      "env-filter",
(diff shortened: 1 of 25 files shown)


Nadrieril:
I investigated, the weirdness has to do with the `const_trait` attribute. E.g. https://github.com/rust-lang/rust/pull/118661 adds this attribute to `PartialEq`, and the consequence in hax/charon is that the trait seems to have an extra `const host: bool` generic argument. This attribute is still here today so we'll have to handle it somehow.

Nadrieril:
The corresponding charon PR is ready (https://github.com/AeneasVerif/charon/pull/254). This can be merged.

W95Psp (approved):
Thank you and sorry for the long response time. That looks good to me, let's merge that ASAP!

Nadrieril:
:tada: thank you!

PR #722, opened 2024-06-17
Don't use strings to represent paths

Nadrieril:


W95Psp (approved):
Thanks, definitely makes sense! I don't even recall why it's like that, but I see no good reason 🤔 
Let's merge it!

PR #727, opened 2024-06-20
fix(engine): `Concrete_ident_generated`: `name` -> `t`, derive more

W95Psp:


jschneider-bensch (approved):
Thanks, it looks good to me! This is very useful, e.g. in #728.

PR #728, opened 2024-06-20
[ProVerif] Renaming / AST transformation pseudophase 

jschneider-bensch:
This PR introduces a pseudo-phase in the ProVerif backend that uses a visitor to perform renamings and last minute AST transformations like simplifying a `Result<T,E>` into a `T`.


franziskuskiefer:
Let's close this and do this as part of the move to the generic printer in #745

PR #729, opened 2024-06-20
Move book from `hacspec/book` to `hacspec/hax`

### LINKED ISSUE #725: Move `hacspec/book` into the main repo
Currently the hax book is living in another repo (hacspec/book), which is annoying:
 - we can't update the book in a hax PR while adding a feature
 - it's hard to keep the book in sync with hax itself.

Action items:
 - [ ] move the book
 - [ ] add a CI job to copy the book over to hacspec/book whenevr there is a change so that the book get published as a GH website

W95Psp:
Fixes #725


book/.github/workflows/gh-pages.yml
franziskuskiefer, line 1: This probably shouldn't be here?

.github/workflows/gh_pages.yml
franziskuskiefer, line 1: Can you add a build and push to the `book` repo?

franziskuskiefer (changes requested):
Is the theme directory actually needed? I don't think we use a special theme.

.github/workflows/gh_pages.yml
W95Psp: That's what the step "regenerate and push the book" from the `build-and-deploy-external-gh-pages` job does! It's building the book and then pushing it as a static website to `hacspec/book`

franziskuskiefer: Ah, maybe you shouldn't do two things at once or at least describe what you're doing.

franziskuskiefer (approved):

Issue #686, opened 2024-05-22
`aes` crate: the frontend loops indefinitely


W95Psp:
To reproduce, clone https://github.com/RustCrypto/block-ciphers at commit ae1892c8, and `cargo hax json` on crate `aes`.

This behavior is the same for libsignal, which uses the crate `aes`.

@franziskuskiefer and I looked at it a bit yesterday (with `RUST_LOG=trace cargo hax json`, hax being compiled in debug mode), and it seems like some `Sized` trait obligations are resolved a **HUGE** number of times before killing the memory of the laptop it's running on.
So probably there is somehow an infinite recursion or loop, but that definitively have to do with trait resolution.

sonmarcho:
I think this is related:
https://github.com/AeneasVerif/charon/issues/191

W95Psp:
Strangely enough, PR https://github.com/hacspec/hax/pull/730 doesn't fix this issue but fixes the looping behavior in `cipher`, from `https://github.com/RustCrypto/traits`. So there are various bugs there.

W95Psp:
This is still a problem, a759a1a87e11a3100e605f0312831596a0543dfb doesn't help

W95Psp:
Still there after #970

W95Psp:
@Nadrieril mostly fixed the bug with its PR #996!
So it was just extremely slow because it was re-doing things again and again.

This pseudo-loop was hiding another bug; I have a fix, will push a PR soon. But the aes crate now extract in under 40 seconds! Thanks @Nadrieril!

W95Psp:
Let's close this. The extraction now works: the extraction itself takes around 40s, but the haxmeta/JSON business takes around 4 minutes on my laptop...

Let's see how we can improve that:
 - https://github.com/hacspec/hax/issues/1001

Issue #683, opened 2024-05-22
Frontend crash while substituting


W95Psp:
To reproduce, run `RUST_BACKTRACE=1 cargo hax json` on the `cipher` crate of the repo https://github.com/RustCrypto/traits at commit [b2e31d8df29](https://github.com/RustCrypto/traits/commit/b2e31d8df293026390b6315c10844eed8b54a185).

Backtrace:
```
error: internal compiler error: compiler/rustc_middle/src/ty/subst.rs:901:9: type parameter `T/#1` (T/1) out of range when substituting, substs=[<Self as crypto_common::BlockSizeUser>::BlockSize]

thread 'rustc' panicked at 'Box<dyn Any>', /rustc/5ea66686467d3ec5f8c81570e7f0f16ad8dd8cc3/compiler/rustc_errors/src/lib.rs:1650:9
stack backtrace:
   0: std::panicking::begin_panic::<rustc_errors::ExplicitBug>
   1: <rustc_errors::HandlerInner>::bug::<alloc::string::String>
   2: <rustc_errors::Handler>::bug::<alloc::string::String>
   3: rustc_middle::util::bug::opt_span_bug_fmt::<rustc_span::span_encoding::Span>::{closure#0}
   4: rustc_middle::ty::context::tls::with_opt::<rustc_middle::util::bug::opt_span_bug_fmt<rustc_span::span_encoding::Span>::{closure#0}, !>::{closure#0}
   5: rustc_middle::ty::context::tls::with_context_opt::<rustc_middle::ty::context::tls::with_opt<rustc_middle::util::bug::opt_span_bug_fmt<rustc_span::span_encoding::Span>::{closure#0}, !>::{closure#0}, !>
   6: rustc_middle::util::bug::bug_fmt
   7: <rustc_middle::ty::subst::SubstFolder>::type_param_out_of_range
   8: <rustc_middle::ty::Ty as rustc_type_ir::fold::TypeSuperFoldable<rustc_middle::ty::context::TyCtxt>>::try_super_fold_with::<rustc_middle::ty::subst::SubstFolder>
   9: rustc_middle::ty::subst::<impl rustc_type_ir::fold::TypeFoldable<rustc_middle::ty::context::TyCtxt> for &rustc_middle::ty::list::List<rustc_middle::ty::subst::GenericArg>>::try_fold_with
[...]

PR #730, opened 2024-06-20
fix(frontend): make `path_to` breadth-first

### LINKED ISSUE #692: Hax hangs on a recursive trait with associated type
```rust
pub trait Trait1 {
    type T: Trait1;
}

pub trait Trait2: Trait1 {
    type U;
}
```

Running `hax json` on this code causes it to hang forever. Removing `type U;` fixes the issue.

W95Psp:
Fixes #692

Status: seems like this is introducing a bug, I have to debug.
Some implementations cannot be found after the change

### DIFF
--- a/frontend/exporter/src/traits.rs
+++ b/frontend/exporter/src/traits.rs
@@ -203,7 +203,7 @@ pub mod rustc {
 
         #[tracing::instrument(level = "trace", skip(s, param_env))]
         pub fn path_to<'tcx, S: UnderOwnerState<'tcx>>(
-            starting_points: Vec<AnnotatedPredicate<'tcx>>,
+            starting_points: &[AnnotatedPredicate<'tcx>],
             s: &S,
             target: PolyTraitRef<'tcx>,
             param_env: rustc_middle::ty::ParamEnv<'tcx>,
@@ -223,16 +223,14 @@ pub mod rustc {
             use std::collections::VecDeque;
             let mut candidates: VecDeque<Candidate<'tcx>> = starting_points
                 .into_iter()
-                .map(|pred| {
+                .filter_map(|pred| {
                     let clause = pred.predicate.as_trait_clause();
                     clause.map(|clause| Candidate {
                         path: vec![],
                         pred: clause,
-                        origin: pred.clone(),
+                        origin: *pred,
                     })
                 })
-                .filter(|candidate| candidate.is_some())
-                .map(|candidate| candidate.unwrap())
                 .collect();
 
             let target_pred = target.upcast(tcx);
@@ -245,7 +243,7 @@ pub mod rustc {
                     if seen.contains(&candidate.pred) {
                         continue;
                     }
-                    seen.insert(candidate.pred.clone());
+                    seen.insert(candidate.pred);
                 }
 
                 // if the candidate equals the target, let's return its path
@@ -257,13 +255,13 @@ pub mod rustc {
                 for (index, parent_pred) in candidate.pred.parents_trait_predicates(s) {
                     let mut path = candidate.path.clone();
                     path.push(PathChunk::Parent {
-                        predicate: parent_pred.clone(),
+                        predicate: parent_pred,
                         index,
                     });
                     candidates.push_back(Candidate {
-                        pred: parent_pred.clone(),
+                        pred: parent_pred,
                         path,
-                        origin: candidate.origin.clone(),
+                        origin: candidate.origin,
                     });
                 }
                 for (item, binder) in candidate.pred.associated_items_trait_predicates(s) {
@@ -271,13 +269,13 @@ pub mod rustc {
                         let mut path = candidate.path.clone();
                         path.push(PathChunk::AssocItem {
                             item,
-                            predicate: parent_pred.clone(),
+                            predicate: parent_pred,
                             index,
                         });
                         candidates.push_back(Candidate {
-                            pred: parent_pred.clone(),
+                            pred: parent_pred,
                             path,
-                            origin: candidate.origin.clone(),
+                            origin: candidate.origin,
                         });
                     }
                 }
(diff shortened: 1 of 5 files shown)


maximebuyse:
> A first fix to this implementation (or to a bug revealed by it) is in #797
> 
> There is another problem remaining (reaching https://github.com/hacspec/hax/blob/main/frontend/exporter/src/traits.rs#L349)

This bug introduced by the initial version of this PR is now fixed. It was just that switching from a recursive depth-first search to a breadth-first search with a queue meant the parents and associated traits had to be taken from the candidate path and not `self`. In other words we were only checking paths of length 1 with `path_to`.

W95Psp (commented):
Nice, thank you @maximebuyse! That looks good to me, and @franziskuskiefer is currently trying to break it by throwing it on bigger crates :sweat_smile: :crossed_fingers:

franziskuskiefer:
The frontend panics now in most cases where it looped before.

```
thread 'rustc' panicked at /rustc/6b0f4b5ec3aa707ecaa78230722117324a4ce23c/compiler/rustc_type_ir/src/binder.rs:713:9:
type parameter `T/#1` (T/#1/1) out of range when instantiating, args=[Alias(Projection, AliasTy { args: [Self/#0], def_id: DefId(5:71 ~ crypto_common[e66d]::BlockSizeUser::BlockSize) })]
```

W95Psp:
This is #683, which is not related. IMO we should merge that PR, that's fixing the looping issue

franziskuskiefer:
> that's fixing the looping issue

How do we know that it is fixing the issue?

maximebuyse:
> > that's fixing the looping issue
> 
> How do we know that it is fixing the issue?

You're right, it doesn't work on #692 . The looping behavior was gone because the breadth-first search had a bug.

W95Psp:
Oh, you mean that it's still looping? :/

W95Psp:
I pushed `make-instantiating-bug-nonfatal` which is making the instantiating bug non-fatal. The looping behavior is gone on https://github.com/RustCrypto/traits/tree/master/cipher at least.

frontend/exporter/src/traits.rs
```diff
@@ -200,62 +199,90 @@ pub mod rustc {
...
-                s: &S,
-                target: PolyTraitRef<'tcx>,
-                param_env: rustc_middle::ty::ParamEnv<'tcx>,
-            ) -> Option<Path<'tcx>> {
-                let tcx = s.base().tcx;
-                if predicate_equality(self.upcast(tcx), target.upcast(tcx), param_env, s) {
-                    return Some(vec![]);
+        #[tracing::instrument(level = "trace", skip(s, param_env))]
+        pub fn path_to<'tcx, S: UnderOwnerState<'tcx>>(
+            starting_points: Vec<AnnotatedPredicate<'tcx>>,
```
W95Psp, line 206: Maybe you can take a `&[AnnotatedPredicate<'tcx>]` instead of a vec here?
Then you would not need to clone on call site.

frontend/exporter/src/traits.rs
```diff
@@ -200,62 +199,90 @@ pub mod rustc {
...
+                .map(|pred| {
+                    let clause = pred.predicate.as_trait_clause();
+                    clause.map(|clause| Candidate {
+                        path: vec![],
+                        pred: clause,
+                        origin: pred.clone(),
+                    })
+                })
+                .filter(|candidate| candidate.is_some())
+                .map(|candidate| candidate.unwrap())
```
W95Psp, line 235: Use one [`filter_map`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter_map) instead of map+filter+map here, we'll get rid of the unwrap :) in general we don't want `unwrap`s at all. Btw, you can use `option.s_unwrap(the_state_of_type_State)` if you really ever need to unwrap: that unwraps with a spanned message.

frontend/exporter/src/traits.rs
```diff
@@ -200,62 +199,90 @@ pub mod rustc {
...
+                for (index, parent_pred) in candidate.pred.parents_trait_predicates(s) {
+                    let mut path = candidate.path.clone();
+                    path.push(PathChunk::Parent {
+                        predicate: parent_pred.clone(),
+                        index,
+                    });
+                    candidates.push_back(Candidate {
+                        pred: parent_pred.clone(),
+                        path,
+                        origin: candidate.origin.clone(),
```
W95Psp, line 266: I think we can make `AnnotatedPredicate` copy. Can you add `Copy` to the derives of `rustc_utils::AnnotatedPredicate`, and then remove `clone`s?

W95Psp (commented):
LGTM, with a few nits!

maximebuyse:
> LGTM, with a few nits!

Thanks for the tips, I made the corrections.

W95Psp:
Nice, thanks! Have you tried the patch on both https://github.com/hacspec/hax/issues/686 and https://github.com/hacspec/hax/issues/692? (I updated `make-instantiating-bug-nonfatal`)
If #692 goes through with `make-instantiating-bug-nonfatal`, let's document that and close it in favor of https://github.com/hacspec/hax/issues/747.

Also, you can resolve conversations on the comments I added, otherwise the PR will be blocked from merging :)

maximebuyse:
> Nice, thanks! Have you tried the patch on both #686 and #692? (I updated `make-instantiating-bug-nonfatal`) If #692 goes through with `make-instantiating-bug-nonfatal`, let's document that and close it in favor of #747.
> 
> Also, you can resolve conversations on the comments I added, otherwise the PR will be blocked from merging :)

This seems to solve #692, I added a test with this example.
But for #686, no change (the memory consumption keeps increasing until I have to kill it). So it might be another problem. 
I didn't try with `make-instantiating-bug-non-fatal` but I didn't get instantiating issues apparently.

franziskuskiefer (approved):
Alright, then let's get this in and keep looking for the issue in #686.

W95Psp:
Mmh, ok. Added a comment there: https://github.com/hacspec/hax/issues/686#issuecomment-2255337646.

PR #733, opened 2024-06-24
Update rustc

Nadrieril:
This updates rustc until Jan 1st 2024.


test-harness/src/snapshots/toolchain__side-effects into-fstar.snap
```diff
@@ -110,12 +112,12 @@ let early_returns (x: u32) : u32 =
...
   let y:u32 = 0ul in
-  let x:u32 = Core.Num.impl__u32__wrapping_add x 1ul in
+  let x:u32 = Core.Num.impl__u32__wrapping_add true x 1ul in
   if x >. 3ul
   then
-    let x:u32 = Core.Num.impl__u32__wrapping_sub x 3ul in
+    let x:u32 = Core.Num.impl__u32__wrapping_sub true x 3ul in
     let y:u32 = x /! 2ul in
-    let y:u32 = Core.Num.impl__u32__wrapping_add y 2ul in
+    let y:u32 = Core.Num.impl__u32__wrapping_add true y 2ul in
```
Nadrieril, line 120: Here you can see examples of the new `host` parameter being added to some functions.

W95Psp: :sob:

W95Psp: Since they are removing that in a later nightly, wouldn't be it worth updating rustc to the latest nightly in once, so that we don't spend time working on something that will disappear later?

W95Psp: But in the same time, from https://github.com/rust-lang/rust/pull/126552, it seems they want to re-introduce it at some point, so maybe we should suffer that pain now

Nadrieril: Well they plan to add them back but in a different form

Nadrieril: Essentially, if they're not too annoying I'd rather we shrug and deal with them. If they're too annoying I can _try_ moving 6 months at once but that'll be a lot of work.

Nadrieril: Actually, from that PR it looks like it affected only `PartialEq` and `Add`, which seems manageable.

W95Psp: To decide whether it's worth updating 6 months of rustc in once, the question is: what form will the new tracking of consts/effects will take? if it is still a const generic, the problem will be the same.

Also, another question is: can we, directly within the frontend, add an option to get rid of that `host` const generics both on method definitions and calls? I guess we should "just" lookup traits definition and check whether it was marked `#[const_trait]`? or it's more complicated?

Issue #696, opened 2024-06-03
Static binaries for the engine


franziskuskiefer:
I (@W95Psp) investiguated that:
 - the JS build is quite slow, e.g. on libcrux node overflows, bun doesn't help
 - I tried a few hours to compile the engine statically, with no success
 - we can try is build under alpine with docker, but this will not help for MacOS

Other solutions:
 - arx (with `nix bundle`)
 - docker
 - find wether there is a big bottleneck when extracted in JS  
 - ...

W95Psp:
Here is a small report on trying to produce distributable binaries for the hax-engine.

## OCaml static binaries
After reading https://ocamlpro.com/blog/2021_09_02_generating_static_and_portable_executables_with_ocaml/, I tried this approach but got some issues on linux.
Was not able to try anything on mac.

## Js_of_ocaml
Ended up being either too slow or crashing because of stack overflow. There is no tail call optimization on node, I tried with `bun` wich has some, but got over performance issues.

I also had a issue about caml performances working on strings: parsing JSON was extremely slow, so I had to use JS's own `JSON.parse` and pipe that into the engine.

This solution works for running hax on small crates but performance is really really limiting.

## `nix bundle`
The Nix build of the engine was shipping Rustc and was huge, https://github.com/hacspec/hax/pull/724 solves that.
With that PR, `nix bundle --bundler github:ralismark/nix-appimage ".#hax-engine.bin"` is building a small-ish portable AppImage linux binary. I think that's exactly what we need -- for linux!

## Just ship dynamic binaries
On Mac, the only dynamic library we rely on is `gmp`, we can probably stop relying on zarith and thus get rid of that dependency.

Action items:
 - [ ] use `nix bundle` in the CI to release binaries
 - [ ] try to kill the dependency to `gmp` or copy the dylib over in case of MacOS

franziskuskiefer:
With #734 merged we should be good to go here?

W95Psp:
Yes, indeed! We should make a release and finish the small bit that downloads the engine from GH!

W95Psp:
Shall I just increment the version in `Cargo.toml`, push on crates.io, and push a git ref to the repo?

franziskuskiefer:
Yeah, just make it a `pre.2`.

PR #734, opened 2024-06-24
refactor(engine/fstar-ast): get rid of `zarith` and GMP

W95Psp:
This helps for #696: this PR gets rid of the dependency to the native lib `gmp`.
That's virtually the only native lib used by the engine, so we can (almost) consider binaries to be static from now on, that will do the trick at least for MacOS

Technically, this PR:
 - remove a bunch of useless code from the F* parser & AST
 - kills the dependency to zarith, replacing zarith's integers with... strings. That is fine since those integers are mainly used to represent constants in the AST


franziskuskiefer (approved):
Nice! Deleting code!

PR #748, opened 2024-07-04
Add an option to downgrade errors to warnings

Nadrieril:
In Charon we want to extract a crate even when there are errors. To achieve that, we use `catch_unwind` when translating any item. Except sometimes this isn't enough, because hax's use of `struct_fatal` and `struct_err` informs rustc that there was an error, which causes Charon to fail. This PR adds an option to avoid that.


Nadrieril:
I found another solution: I can just ignore rustc's errors in Charon. So unless you have a use for this too, this is no longer useful.

PR #751, opened 2024-07-08
Engine: propagate trait generics arguments

### LINKED ISSUE #719: Some class parameters are incorrectly passed
In some extracted F* files in `libcrux-ml-kem`, arguments for classes are passed as explicit parameters. For example, `v_K` variable here is supposed to be ghost parameter https://github.com/cryspen/libcrux/blob/dev/libcrux-ml-kem/proofs/fstar/extraction/Libcrux_ml_kem.Ind_cca.fst#L89

It's fixed manually in `dev_ml_kem_lax` branch https://github.com/cryspen/libcrux/blob/dev_ml_kem_lax/libcrux-ml-kem/proofs/fstar/extraction/Libcrux_ml_kem.Ind_cca.fst#L89

W95Psp:
~This PR requires #698.~

This PR propagates trait generics arguments throughout the engine.

On the F* output:
 - now the calls to methods will always be `method #self_type #trait_generic_arg1 ... #trait_generic_argN #Typeclasses.solve #method_generic_type1 ... #method_generic_typeM method_generic_const1 ... method_generic_constI meth_arg1 ... meth_argJ`
 - traits and inductive (enums/structs) should have all their argument explicit

Todo:
 - [x] use the information in the F* backend
 - [x] propagate the information in the proof libs

When undrafted, this PR should fix #719

### DIFF
--- a/engine/backends/fstar/fstar-surface-ast/FStar_Parser_Const.ml
+++ b/engine/backends/fstar/fstar-surface-ast/FStar_Parser_Const.ml
@@ -499,6 +499,8 @@ let (mk_class_lid : FStar_Ident.lid) =
   fstar_tactics_lid' ["Typeclasses"; "mk_class"]
 let (tcresolve_lid : FStar_Ident.lid) =
   fstar_tactics_lid' ["Typeclasses"; "tcresolve"]
+let (solve_lid : FStar_Ident.lid) =
+  fstar_tactics_lid' ["Typeclasses"; "solve"]
 let (tcclass_lid : FStar_Ident.lid) =
   fstar_tactics_lid' ["Typeclasses"; "tcclass"]
 let (tcinstance_lid : FStar_Ident.lid) =
--- a/engine/backends/fstar/fstar_ast.ml
+++ b/engine/backends/fstar/fstar_ast.ml
@@ -35,6 +35,7 @@ module Attrs = struct
 end
 
 let tcresolve = term @@ AST.Var FStar_Parser_Const.tcresolve_lid
+let solve = term @@ AST.Var FStar_Parser_Const.solve_lid
 
 let pat_var_tcresolve (var : string option) =
   let tcresolve = Some (AST.Meta tcresolve) in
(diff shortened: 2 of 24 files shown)


W95Psp:
I think this should work, maybe there are a few things in the library we need to fix, but we will discover those by running hax on various things.

@mamonet can you try this branch on ml-kem?
If you run in anything suspicious related to implcits or something similar, tell me directly!

franziskuskiefer:
@W95Psp this doesn't build with the same issue as in the visitor PR
```
# File "utils/generate_visitors/errors.ml", line 27, characters 29-35:
# 27 |     | UnsupportedCoreType of string
#                                   ^^^^^^
```

W95Psp:
Ah, yes, thanks @franziskuskiefer, will rebase this PR with `generate-visitors` to propagate that change

franziskuskiefer (approved):
lgtm when it works for @mamonet

mamonet:
This PR seems to be solving the issue. I went through the instances that I'm aware of, and all of them verified successfully. However, I got one undesired change here https://github.com/cryspen/libcrux/blob/dev_ml_kem_lax/libcrux-ml-kem/proofs/fstar/extraction/Libcrux_ml_kem.Polynomial.fsti#L35-L37
This create an error ```Expected type Type
  but t_PolynomialRingElement v_Vector
  has type i1: Vector.Traits.t_Operations v_Vector -> Type0``` in references of `t_PolynomialRingElement`
Here is the original type https://github.com/cryspen/libcrux/blob/main/libcrux-ml-kem/proofs/fstar/extraction/Libcrux_ml_kem.Polynomial.fsti#L35-L37

W95Psp:
Fantastic!
Thanks for your feedback! I see what it is, give me 10-15 minutes I will push a fix!

W95Psp:
@mamonet, I pushed a commit, can you tell me if that's fixing that regression? :smiley:

mamonet:
Of course, just 2 minutes to report back

mamonet:
The fix gets `t_PolynomialRingElement` back to what it is, I think this PR is a good shape now.

mamonet (approved):

W95Psp:
Ah, that's great news!
I'm gonna rebase the PR to main, and then we can merge it! :partying_face:

PR #743, opened 2024-06-30
Refactor of the frontend

W95Psp:
This PR is a "big" refactor, I guess that's my way of killing time on this scary election Sunday in France...

Sorry for the bulky monolithic PR, that's a big refactor thing with a lot of very related changes, that would have been 5 or 10 PRs, that would have been a pain to push and merge.

 - Architecture:
    + `cargo-hax` now:
       1. calls `cargo build`, which calls multiple times our custom rustc driver. Instead of running the engine itself or spitting out JSON, the rustc driver now creates a `haxmeta` file per crate in the `target` directory (along with `rmeta` rustc files).  
           A `haxmeta` file is a full  ~CBOR~ [bincode](https://crates.io/crates/bincode/) export of the AST of a crate by hax's exporter. It contains all the data we need.  
           It also communicates on `stderr` what haxmeta files have been produced.  
           This `haxmeta` file mechanism respects much more what rustc does itself, allowing us to benefit from the cache of cargo and rustc. Also, we need that mechanism for things like #371: with `haxmeta` files, we will be able to do cross-crate reasoning.
       3. reading the `stderr` of `cargo build`, we now what `haxmeta` files were produced.
       4. we process those `haxmeta` files, and call the engine
    + The engine is now communicating with `cargo-hax` with messages (on stdin/stdout). Before, the communication was done with one JSON input payload at the beginning and one JSON payload at the end. Now, the engine can communicate interactively with `cargo-hax`. This is useful for:
       * dynamic error reporting
       * rustfmt requests
       * diagnostics pretty printing
       * in the future, a nicer debugging tool
    + I removed the crates `hax-cli-options`, `hax-cli-options-engine`, and `hax-diagnostics`.
       * This stratification stems from us thinking we needed some crates to be nightly and some to be stable, that's not the case
       * I merged everything into a `hax-types` crate, which contains all the types required for communicating between `cargo-hax`, `hax-engine` and the custom rustc driver.
 - Minor changes:
    + use `prettyplease` instead of `rustfmt` to format our internal Rust-ish AST (that's much faster)
    + re-enable cargo cache by default
    + errors are displayed with `annotate_snippets`, which is Rustc's crate for reporting errors. Closes https://github.com/hacspec/hax/issues/739.
    + simplify diagnostics
    + remove the linter (closes #611)
    + clean up some type in the CLI options
    + remove `pretty_print_diagnostics`
    + use a different `target` directory to avoid recompilation
[...]

PR #753, opened 2024-07-09
 exporter: make it need nightly only when feature `rustc` is on

### LINKED ISSUE #749: hax-frontend-exporter: make it non-nightly when feature `rustc` is disabled
Currently `hax-frontend-exporter`, even when the `rustc` feature is disabled, requires nightly. That should not be the case.

W95Psp:
This PR removes some use of nightly features or feature-gate them, based on the `rustc` feature introduced by #743.
This is useful for Rust projects that depend on the ASTs of hax but not on librustc_driver (e.g. Testify).

Fixes #749

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,6 +30,9 @@ jobs:
       - name: Test
         run: cargo test --workspace --exclude hax-engine-names-extract --verbose
 
+      - name: Test `hax-frontend-exporter` with feature `rustc` off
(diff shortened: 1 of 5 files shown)


franziskuskiefer (approved):
Should this (the non-nightly build) be tested on CI as well? Or do we do that already?

W95Psp:
Ah, that's a good idea. Should I add a `cargo check --no-default-feature -p hax-frontend-exporter`?

PR #758, opened 2024-07-10
fix color printing in util script

paulmure:
The rebuild script uses the `\e` argument to print in different colors.
But the `\e` encoding is not supported in all shells, whereas `\033` should be widely supported.


W95Psp (approved):
Looks good to me, thank you! :)

Issue #757, opened 2024-07-09
Frontend: panic on function item types.


paulmure:
The following code:
```rust
fn foo<T>() {}

fn bar() {
    let foo_ptr = foo::<i32>;
    foo_ptr();
}
```
produces the following error message:
```
disabled backtrace
error[E9999]: Supposely unreachable place in the Rust AST. The label is "CallNotTyFnDef".
              This error report happend because some assumption about the Rust AST was broken.
              
              Context:
               - e: Expr {
                  kind: VarRef {
                      id: LocalVarId(
                          HirId(DefId(0:5 ~ scratch[5c53]::bar).7),
                      ),
                  },
                  ty: FnDef(
                      DefId(0:3 ~ scratch[5c53]::foo),
                      [
                          i32,
                      ],
                  ),
                  temp_lifetime: Some(
                      Node(11),
                  ),
                  span: src/lib.rs:5:5: 5:12 (#0),
              }
               - kind: VarRef {
                  id: LocalVarId(
                      HirId(DefId(0:5 ~ scratch[5c53]::bar).7),
                  ),
              }
               - ty_kind: FnDef(
                  DefId(0:3 ~ scratch[5c53]::foo),
                  [
                      i32,
                  ],
              )
 --> src/lib.rs:5:5
  |
5 |     foo_ptr();
  |     ^^^^^^^
  |
  = note: ⚠️ This is a bug in Hax's frontend.
          Please report this error to https://github.com/hacspec/hax/issues with some context (e.g. the current crate)!

disabled backtrace
error[E9999]: RefCallNotTyFnPtr
 --> src/lib.rs:3:1
  |
3 | fn bar() {
  | ^^^^^^^^
  |
  = note: ⚠️ This is a bug in Hax's frontend.
          Please report this error to https://github.com/hacspec/hax/issues with some context (e.g. the current crate)!

error: could not compile `scratch` (lib) due to 2 previous errors
```

For more reference on function item types: [https://doc.rust-lang.org/reference/types/function-item.html](https://doc.rust-lang.org/reference/types/function-item.html).

W95Psp:
Hi, thank you for the bug report!
Indeed, the current handling of what is a callable function is a bit bizarre; I recall I wanted to change that at some point but never did. So thanks again, that's a good motivation for doing that change :)
I pushed a PR!

PR #759, opened 2024-07-10
fix(frontend): crashes on fn ptr

W95Psp:
This commit fixes the issue #757: the frontend was assuming functions expressions to be of a very specific shape. This commit fixes that.


frontend/exporter/src/types/copied.rs
```diff
@@ -2412,66 +2412,50 @@ pub enum ExprKind {
...
+                    id: def_id.sinto(gstate)
+                });
+                let mut translated_generics = generics.sinto(gstate);
+                let tcx = gstate.base().tcx;
+                r#trait = (|| {
+                    let assoc_item = tcx.opt_associated_item(*def_id)?;
+                    let assoc_trait = tcx.trait_of_item(assoc_item.def_id)?;
+                    let trait_ref = ty::TraitRef::new(tcx, assoc_trait, generics.iter());
+                    let impl_expr = {
+                        // TODO: we should not wrap into a dummy binder
```
franziskuskiefer, line 2430: Should this todo be addressed?

franziskuskiefer (approved):
Thanks.

It looks like the following two functions get translated very differently. Is that on purpose?

```rust
fn do_twice(f: fn(i32) -> i32, arg: i32) -> i32 {
    f(arg) + f(arg)
}

fn do_again<F>(f: F, arg: i32) -> i32
where
    F: Fn(i32) -> i32,
{
    f(arg) + f(arg)
}
```

```ocaml
let do_again
      (#v_F: Type0)
      (#[FStar.Tactics.Typeclasses.tcresolve ()] i1: Core.Ops.Function.t_Fn v_F i32)
      (f: v_F)
      (arg: i32)
    : i32 =
  (Core.Ops.Function.f_call #v_F #i32 f (arg <: i32) <: i32) +!
  (Core.Ops.Function.f_call #v_F #i32 f (arg <: i32) <: i32)

let do_twice (f: (i32 -> i32)) (arg: i32) : i32 = f arg +! f arg
```

frontend/exporter/src/types/copied.rs
W95Psp: @Nadrieril and I are looking at those things in parallel!

W95Psp:
Yes, higher-order functions are a bit problematic, that's related to what @paulmure says in https://github.com/hacspec/hax/issues/549#issuecomment-2218838180

PR #763, opened 2024-07-10
fix(hax-lib): refinements: various fixes

W95Psp:
This PR:
 - fixes a bad F* file in the library
 - fixes a bug where `f(y.get()).get()` would rewrite `f(y.get()).get()`, but not the inner `y.get()`, and adds a test for that
 - fixes an unrelated bug abound prettyplease panicking on `unimplemented!`

### DIFF
--- a/tests/attributes/src/lib.rs
+++ b/tests/attributes/src/lib.rs
@@ -225,7 +225,7 @@ mod refinement_types {
 }
 mod nested_refinement_elim {
     use hax_lib::*;
-    #[hax_lib::refinement_type(|x| true)]
+    #[refinement_type(|x| true)]
     pub struct DummyRefinement(u16);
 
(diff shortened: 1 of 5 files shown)


tests/attributes/src/lib.rs
```diff
@@ -223,3 +223,12 @@ mod refinement_types {
     #[hax_lib::refinement_type(|x| x == 4 || x == 5 || x == 10 || x == 11)]
     pub struct CompressionFactor(u8);
 }
+mod nested_refinement_elim {
+    use hax_lib::*;
+    #[hax_lib::refinement_type(|x| true)]
```
franziskuskiefer, line 228:
```suggestion
    #[refinement_type(|x| true)]
```

franziskuskiefer (approved):

Issue #549, opened 2024-03-05
Associated types are not extracted correctly in general


W95Psp:
The F* backend (and probably others) is not handling "non-trivial" associated type, e.g.:
```rust
pub trait Foo {
    type FooType;
}
pub trait Bar {
    type BarType: Foo;
    fn f() -> <Self::BarType as Foo>::FooType;
}
```

This is related to #536

paulmure:
There is a similar issue with projections. Here's an important example:
```rust
fn foo<F>(f: F) -> i32
where
    F: Fn(u64) -> i32,
{
    f(42)
}
```
The `Fn` trait is implemented as an extension of `FnMut`, which is itself an extension of `FnOnce`, and only `FnOnce` stores the output type as an associated type.

In the example above, figuring out the return type of `f` would require looking at the projection clause from the list of bounds on generic parameters.

This information is [exported by the frontend here](https://github.com/hacspec/hax/blob/main/frontend/exporter/src/types/copied.rs#L3638) but [ignored by the importer in the backend](https://github.com/hacspec/hax/blob/main/engine/lib/import_thir.ml#L1099).

I don't think any backend correctly translates the example above at the moment.

W95Psp:
Hi @paulmure, thanks for your comment! That's true, the engine doesn't support at all equality constraints. That would be great to add support for that!
Let me open a new issue. This one was more about the F* extraction which is broken for even less complicated things :sweat_smile:

maximebuyse:
I had an issue that seems to be related. The equality constraints inside a trait type argument are not translated to f*: https://hax-playground.cryspen.com/#fstar/3b38e61b99/gist=cc40e5e2c6579d215293f9378b9e871f

PR #765, opened 2024-07-10
fix(engine/lib): import associated item projection on generic bounds

paulmure:
Partially address #549.

The importer in the engine now imports projections defined in `hax_frontend_exporter::types::copied::ClauseKind`.

This is done by adding a variant to the `Ast.Rust.generic_constraint` enum to include projections, which necessitates changing the interface of `import_clause` in `engine/lib/import_thir.mli` to return `Ast.Rust.generic_constraint` instead of `Ast.Rust.impl_ident`.

As a result, all functions that previously expect an `Ast.Rust.impl_ident` from `c_clause_kind` now do some additional filtering.

**Remaining issues:**

- Each backend still need to add their own support to process associated item projections. For now, I just added a call to `Error.unimplemented` with reference to #549.
- Printing generic constraints still need to be polished as part of [https://github.com/hacspec/hax/pull/533](https://github.com/hacspec/hax/pull/533).

### DIFF
--- a/cli/subcommands/src/cargo_hax.rs
+++ b/cli/subcommands/src/cargo_hax.rs
@@ -229,7 +229,11 @@ fn run_engine(
                 }
                 FromEngine::PrettyPrintRust(code) => {
                     let code = match syn::parse_file(&code) {
-                        Ok(file) => Ok(prettyplease::unparse(&file)),
+                        Ok(file) => match std::panic::catch_unwind(|| prettyplease::unparse(&file))
+                        {
+                            Ok(pp) => Ok(pp),
+                            Err(err) => Err(format!("prettyplease panicked with: {:#?}", err)),
+                        },
                         Err(err) => Err(format!("{}", err)),
                     };
                     send!(&ToEngine::PrettyPrintedRust(code));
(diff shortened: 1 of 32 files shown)


paulmure:
@W95Psp, the fstar backend currently fails the traits test as I do not know how this additional feature should be dealt with there. Please let me know if you have any ideas on how to push this forward.

W95Psp:
Hi! Thanks for your PR!

It looks good to me, I can add the missing bits for the F* backend, or we can just filter that out for now (I'm not entirely sure about what is the best encoding in F*). Do you want to take care of the filtering in the F* and Coq backends?

For now, let's not feature-gate this new variant, let's see if that makes sense at some point for some backends.

(What's up with the git history though? :eyes:)

paulmure:
The coq backend is surprisingly unaffected by this change. The ssprove backend is though, so I just called `Error.unimplemented` there.

For the F* backend, I filtered `GCProjection`s out with a TODO comment linking to https://github.com/hacspec/hax/issues/785.

The git history was probably because I accidentally merged some changes in main...

W95Psp (approved):
Great, thanks!
Ok, that will be a merge commit anyway, let's merge as is!

PR #776, opened 2024-07-16
feat(book): fix checkboxes

W95Psp:
The book uses custom checkboxes in the quickstart chapter. When we moved the book to this repo (see
https://github.com/hacspec/hax/pull/729), we dropped the theme.

mdbook documentation suggest (see
https://rust-lang.github.io/mdBook/format/theme/index.html) copying and patching the default theme.

Since here it's such a small change, I added a small script that post-process the HTML instead.

### DIFF
--- a/book/default.nix
+++ b/book/default.nix
@@ -9,6 +9,7 @@ stdenv.mkDerivation {
   buildPhase = ''
             mdbook build
             mdbook build archive -d ../book/archive
+            bash ./postprocess.sh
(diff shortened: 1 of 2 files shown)


franziskuskiefer (approved):
We have check boxes? Fine with me.

W95Psp:
Yeah, on the quick start! So that users follow the different steps one by one, as if that was a task list!

PR #779, opened 2024-07-17
Support let-else pattern.

### LINKED ISSUE #155: `let .. = .. else { .. }`
See https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html.

I think we should just transform those into `match`es.

maximebuyse:
Fixes #155


W95Psp (approved):

W95Psp (changes requested):
Thanks! Can you add a test?

engine/lib/import_thir.ml
```diff
@@ -370,6 +370,84 @@ end) : EXPR = struct
...
+                              };
+                            body = { else_block with typ = body.typ };
+                          };
+                        span = else_block.span;
+                      };
+                    ];
+                  scrutinee = rhs;
+                }
+            in
+            { e; typ; span = Span.union rhs.span lhs_body_span }
```
W95Psp, line 439: I think `Span.union rhs.span lhs_body_span` is equivalent to `span`

W95Psp (approved):

PR #788, opened 2024-07-18
Fix #296

### LINKED ISSUE #296: Add support for `dyn`
From a dependent type perspective, a `dyn` can be seen as a dependent tuple.
Given a trait `Foo`, `dyn Foo` denotes an existentially quantified type: there exists a type that implements trait `Foo`.

Thus, an inhabitant of `dyn Foo` is simply a type `T`, a value of type `T`, and an impl of `Foo` for `T`. In other terms, `dyn Foo` is `t: Type & t & foo t`.

Note that `Foo` (and any trait, actually) should be implemented for such dependent tuples.

maximebuyse:
This fixes #296. Some work still needs to be done on the f* backend (see https://github.com/hacspec/hax/issues/787)
As part of this MR: `dyn` is added to the hax AST and created by the frontend. A (broken) translation is produced by the f* backend: for example `dyn (A + B<C>)` is translated as `dyn 2 (fun z -> A z) (fun z -> B z C)`.

### DIFF
--- a/engine/backends/coq/coq/coq_backend.ml
+++ b/engine/backends/coq/coq/coq_backend.ml
@@ -40,7 +40,8 @@ module SubtypeToInputLanguage
              and type while_loop = Features.Off.while_loop
              and type for_index_loop = Features.Off.for_index_loop
              and type quote = Features.Off.quote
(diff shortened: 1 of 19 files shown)


W95Psp (approved):
LGTM, thanks!

PR #792, opened 2024-07-18
feat(frontend): Add option to control output directory

### LINKED ISSUE #791: Frontend: Add option to control output directory
Currently, the extracted code gets put into `proofs/[backend]/extraction` and there does seem to be an option to change that.

We can add an optional `--output-dir` option to control where the output file goes. The default behavior can remain unchanged.

paulmure:
Close #791.

### DIFF
--- a/hax-types/src/cli_options.rs
+++ b/hax-types/src/cli_options.rs
@@ -329,7 +329,8 @@ pub struct BackendOptions {
     #[command(flatten)]
     pub translation_options: TranslationOptions,
 
-    /// Where to put the output files.
+    /// Where to put the output files resulting from the translation.
+    /// Defaults to "<crate folder>/proofs/<backend>/extraction".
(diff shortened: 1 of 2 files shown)


hax-types/src/cli_options.rs
```diff
@@ -328,6 +328,10 @@ pub struct BackendOptions {
 
     #[command(flatten)]
     pub translation_options: TranslationOptions,
+
+    /// Where to put the output files.
```
W95Psp, line 332: Maybe add a "defaults to `<crate folder>/proofs/<backend>/extraction`" or something along those lines?

W95Psp (approved):
Thanks, LGTM!
I only have one remark.

hax-types/src/cli_options.rs
paulmure: Done.

PR #794, opened 2024-07-22
Resugar asserts

### LINKED ISSUE #285: Resugar `assert!`s
Rust desugars `assert!(condition)` as something like `if !condition {panic!()}`, which is not very nice looking.

We should detect and rewrite those as calls to a hax-lib function, say, `hax::assert(condition)`

maximebuyse:
Fixes #285


cli/subcommands/src/cargo_hax.rs
```diff
@@ -199,7 +199,7 @@ fn run_engine(
 
         let stdout = std::io::BufReader::new(engine_subprocess.stdout.take().unwrap());
         for msg in stdout.json_lines() {
-            let msg = msg.unwrap();
+            let msg = msg.expect("hax engine sent an invalid json value");
```
W95Psp, line 202: Can you add something like: "This might be caused by debug messages on stdout, which is reserved for JSON communication with `cargo-hax`"?

W95Psp (approved):
Nice, thank you, that looks good to me!
I just have one very small thing on the error message.

cli/subcommands/src/cargo_hax.rs
maximebuyse: Added now.

PR #795, opened 2024-07-22
fix: precondition constraints should be of the shape `true => pred`

W95Psp:
PR #790 was outputting `pre_instance => pre_trait` for precondition.
We need the opposite: the precondition of the instance should always be implied by the one defined on the trait.


mamonet:
I can confirm this PR solves the issue. Thanks, Lukas.

mamonet (approved):

PR #796, opened 2024-07-23
Add nicer error messages when rustc emits an error in detect_forking.

### LINKED ISSUE #640: `features.rs`: `detect_forking` generates ugly error messages
`detect_forking` intent is to discover which feature is enabled by running a `cargo build`.

It uses a very simple custom driver, which stops right after parsing. 
This makes hax fail with a quite horrible error message right now when running on a crate with a bad path or something like that.

maximebuyse:
Fixes #640


W95Psp (approved):
LGTM!

PR #813, opened 2024-07-29
feat(hax-lib/f*): add `verif_status` and `options`

W95Psp:
@karthikbhargavan @mamonet that's what we discussed:

```rust
#[hax_lib::fstar::verification_status(panic_free)]
#[hax_lib::requires(x < 30)]
#[hax_lib::ensures(|_|
    /*something hard to prove, symbolized by false here*/
    false
)]
fn f(x: u8) -> u8 {
    x + x
}

#[hax_lib::fstar::verification_status(lax)]
fn g(x: u8) -> u8 {
    x + x
}

#[hax_lib::fstar::options("--z3rlimit 1234")]
fn h(x: u8) -> u8 {
    x + x
}


```
<sup>&nbsp;&nbsp;&nbsp;[Open this code snippet in the playground](https://hax-playground.cryspen.com/#fstar/526fc073dc/gist=09e66ea206ccaca45ff4021b553515fa)</sup>

### DIFF
--- a/hax-lib-macros/src/lib.rs
+++ b/hax-lib-macros/src/lib.rs
@@ -23,6 +23,71 @@ use impl_fn_decoration::*;
 use prelude::*;
 use utils::*;
 
+/// When extracting to F*, wrap this item in `#push-options "..."` and
+/// `#pop-options`.
+#[proc_macro_error]
+#[proc_macro_attribute]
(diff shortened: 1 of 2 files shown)


karthikbhargavan (approved):
How about being verbose with "verification_status",
and how about allowing us to use other statuses like: "verified" (does nothing), or "ignored" (does not generate any code in F*) ? Or maybe you have some other ideas.

Issue #818, opened 2024-07-31
Add span for guards in pattern matchings


maximebuyse:
#814 introduces guards in pattern matchings but no span is associated to them (only the arm span), it would be better to have a type `guard'` with a span in it.

maximebuyse:
Implemented in #821

Issue #807, opened 2024-07-29
Matches: guards are dropped!


W95Psp:
The following example:
```rust
fn f(x: u8) -> bool {
    match x {
        _ if x > 20 => true,
        _ => false,
    }
}
```
[Open this code snippet in the playground](https://hax-playground.cryspen.com/#fstar/d6cc1888b7/gist=3687f970d20c0230ecde83cf89064111)

Is translated without any guards!

Action items:
 - [ ] add guards in arms in the AST, feature-gate them
 - [ ] add a rejection phase and reject them for most backends (at least for now)
 - [ ] investigate how mutation behaves there

W95Psp:
Closed by #821

Issue #806, opened 2024-07-29
`match`s simplification phase


W95Psp:
- https://github.com/hacspec/hax/issues/463
 - https://github.com/hacspec/hax/issues/804

maximebuyse:
A first solution that we designed to deal with these issues would be to add `if let` guards for patterns in the Hax AST. When a guard is present in the initial rust code, we would import it in the AST as an if let guard (`if let true = guard`). Then we would add the following phases:

- Destruct opaque type patterns: This one would be parameterized by the list of opaque types to deal with (for example usize, isize, arrays for the f* backend) and how to rewrite the patterns for these types. For example `match x { [a,..] => a, _ => 0}` would rewrite as `match x { v if let Some(a) = if v.len() >= 1 {Some(v[0])} else {None}=> a, _ => 0}` 
- A phase eliminating these `if let` patterns: 
```
match x {
  None => 0, 
  Some(v) if let A(y) = v => y, 
  Some(B(y)) => y,
  _ => 1}
``` 
would rewrite as 
```
match x {
  None => 0, 
  _ => 
    let r = match x {
      Some(v) => match v {A(y)=> Some(y), _ => None},
      _ => None};
    match r {
       Some(y) => y,
       None => match x {
         Some(B(y)) => y,
         _ => 1}
    }
}
```
- A phase eliminating nested disjunctive patterns: Rewrite them as shallow patterns. `match x { Some(1|2) => 1, _ => 0}` would rewrite as  `match x { Some(1)|Some(2) => 1, _ => 0}` 
- A phase eliminating shallow disjunctive patterns (for backends that don't support them): Duplicate the arms. `match x { Some(1)|Some(2) => 1, _ => 0}`  would rewrite as `match x { Some(1) => 1, Some(2) => 1, _ => 0}`

maximebuyse:
Sub-issue for if let guards: #814

maximebuyse:
> A first solution that we designed to deal with these issues would be to add `if let` guards for patterns in the Hax AST. When a guard is present in the initial rust code, we would import it in the AST as an if let guard (`if let true = guard`). Then we would add the following phases:
> 
[...]

PR #821, opened 2024-08-01
Implement (if-let) guards in pattern matchings

### LINKED ISSUE #814: Support if let guards in pattern matchings
[Guards](https://doc.rust-lang.org/rust-by-example/flow_control/match/guard.html) in pattern matchings are supported in Rust but not necessarily in hax backends. We want to support them, and even a generalization with [if-let](https://doc.rust-lang.org/rust-by-example/flow_control/if_let.html) guards (if-lets in guards are a nightly feature in Rust). They should be rewritten by a dedicated phase (see #806).

maximebuyse:
Closes #814

This adds support for guards in pattern matchings. They are added to the AST (with a corresponding feature guard), and handled in the THIR import. A phase rewrites them as an equivalent nested pattern matching. Apart from supporting guards as specified in #807, this will also be useful for future phases that will generate guards (see #806).

Example of code with if-let guard (which is currently a nightly feature) and Rust code for the equivalent after the new phase:
```
pub fn if_let_guard(x: Option<Result<i32, i32>>) -> i32 {
    match x {
        None => 0,
        Some(v) if let Ok(y) = v => y,
        Some(Err(y)) => y,
        _ => 1,
    }
}

pub fn equivalent(x: Option<Result<i32, i32>>) -> i32 {
    match x {
        None => 0,
        _ => match match x {
            Some(v) => match v {
                Ok(y) => Some(y),
                _ => None,
            },
            _ => None,
        } {
            Some(y) => y,
            None => match x {
                Some(Err(y)) => y,
                _ => 1,
            },
        },
    }
}
```
Regular if guards are also supported (but for now desugared as if-let guards):
```
pub fn if_guard(x: Option<i32>) -> i32 {
    match x {
        Some(v) if v > 0 => v,
        _ => 0,
    }
}

pub fn equivalent(x: Option<i32>) -> i32 {
    match x {
        Some(v) if let true = v > 0 => v,
        _ => 0,
    }
}
```


engine/lib/ast.ml
```diff
@@ -338,8 +338,11 @@ functor
           witness : F.nontrivial_lhs;
         }
 
+    and guard = { guard : guard'; span : span }
```
W95Psp, line 341: Can you add:
 - a doc comment that explains what is a `guard` (with an example)
 - a comment on `guard'` explaining why there is only one variant currently, and a link to an issue that explains the roadmap for the work left on guards (that is, simple lets and let-chains), with link to Rust documentation on that

engine/lib/ast_utils.ml
```diff
@@ -410,8 +410,12 @@ module Make (F : Features.T) = struct
           inherit [_] Visitors.reduce as super
           inherit [_] Sets.Local_ident.monoid as _m
 
-          method! visit_arm' env { arm_pat; body } =
-            shadows ~env [ arm_pat ] body super#visit_expr
+          method! visit_arm' env { arm_pat; body; guard } =
+            match guard with
```
W95Psp, line 414: Here, you can remove duplicate code with a `shadows ~env [ arm_pat ] rhs super#visit_expr ++ (Option.map ~f:(...) guard |> Option.value ~default_or_thunk:self#zero))`

engine/lib/ast_utils.ml
```diff
@@ -601,8 +607,17 @@ module Make (F : Features.T) = struct
...
+          match guard with
+          | Some { guard = IfLet { lhs; rhs; _ }; _ } ->
+              let rhs_vars =
+                without_pat_vars (self#visit_expr () rhs) arm_pat
+              in
+              let body_vars =
+                without_pats_vars (self#visit_expr () body) [ arm_pat; lhs ]
+              in
+              Set.union rhs_vars body_vars
+          | None -> without_pat_vars (self#visit_expr () body) arm_pat
```
W95Psp, line 620: You can use a similar strategy as above, with an `Option.map`

engine/lib/ast_utils.ml
```diff
@@ -777,6 +792,9 @@ module Make (F : Features.T) = struct
 
   let make_wild_pat (typ : ty) (span : span) : pat = { p = PWild; span; typ }
 
+  let make_unguarded_arm (arm_pat : pat) (body : expr) (span : span) : arm =
+    { arm = { arm_pat; body; guard = None }; span }
```
W95Psp, line 796: Can you make this into a `make_arm`, with an optional (defaulting to `None`) `?guard` argument?

engine/lib/generic_printer/generic_printer.ml
```diff
@@ -430,10 +430,17 @@ module Make (F : Features.T) (View : Concrete_ident.VIEW_API) = struct
...
+          fun { arm_pat; body; guard } ->
             let pat = print#pat_at Arm_pat arm_pat |> group in
             let body = print#expr_at Arm_body body in
-            pat ^^ string " => " ^^ body ^^ comma
+            let g =
+              match guard with
+              | Some { guard = IfLet { lhs; rhs; _ }; _ } ->
+                  string " if let " ^^ print#pat_at Arm_pat lhs ^^ string " = "
+                  ^^ print#expr_at Arm_body rhs
+              | None -> string ""
```
W95Psp, line 441: Can you:
 - use `empty` instead of `string ""`
 - use a `Option.map ~f:(...) guard |> Option.value ~default:empty`?

W95Psp: Oh and why not renaming `g` into `guard` here?

engine/lib/import_thir.ml
```diff
@@ -418,7 +418,10 @@ end) : EXPR = struct
                 {
                   arms =
                     [
-                      { arm = { arm_pat = lhs; body }; span = lhs_body_span };
+                      {
+                        arm = { arm_pat = lhs; body; guard = None };
+                        span = lhs_body_span;
+                      };
```
W95Psp, line 424: In this file we should probably use `make_(unguarded_)arm` now :)

engine/lib/import_thir.ml
```diff
@@ -1091,7 +1095,36 @@ end) : EXPR = struct
...
+                  IfLet
+                    {
+                      lhs =
+                        { p = PConstant { lit = Bool true }; span; typ = TBool };
+                      rhs = c_expr e;
+                      witness = W.match_guard;
+                    };
+                span = Span.of_thir e.span;
+              })
+        arm.guard
```
W95Psp, line 1125: Can you put the match in a `let` binding? e.g. a `let guard = match ... with ... in {guard; span = Span.of_thir ...}`?

engine/lib/phases/phase_drop_match_guards.ml
```diff
@@ -0,0 +1,200 @@
...
+    and transform_arms (scrutinee : B.expr) (remaining : A.arm list)
+        (treated : B.arm list) : B.arm list =
+      match remaining with
+      | [] -> treated
+      | { arm = { arm_pat; body; guard = None }; span } :: remaining ->
+          let new_arm : B.arm =
+            {
+              arm = { arm_pat = dpat arm_pat; body = dexpr body; guard = None };
+              span;
+            }
```
W95Psp, line 50: Maybe use `U.make_arm`?

maximebuyse:
This also closes #818 and #807

engine/lib/phases/phase_drop_match_guards.ml
```diff
@@ -0,0 +1,200 @@
...
+              e =
+                Match
+                  {
+                    scrutinee;
+                    arms =
+                      [
+                        UB.make_unguarded_arm (dpat arm_pat)
+                          {
+                            e =
+                              Match
```
W95Psp, line 117: Can you hoist this inner match to a let binding? I mean in OCaml, not in the generated code.
Then, can you add a comment that explains what is this sub-match exactly, with an example (e.g. a quote of the example you'll have earlier in the code)

engine/lib/phases/phase_drop_match_guards.ml
```diff
@@ -0,0 +1,200 @@
...
+          arm =
+            {
+              arm_pat;
+              body;
+              guard = Some { guard = IfLet { lhs; rhs; _ }; span = guard_span };
+            };
+          span;
+        }
+        :: remaining ->
+          let result_typ = dty span body.typ in
```
W95Psp, line 63: Here, can you add a comment with an example in Rust syntax (you can use `<lhs>`, `<rhs>`, `<body>` et `<arm_pat>` in the example)?
I.e. an example of the matched arm + what will be the rewrite?

engine/lib/phases/phase_drop_match_guards.ml
```diff
@@ -0,0 +1,200 @@
...
+      include Features.SUBTYPE.Id
+    end
+
+    [%%inline_defs dmutability]
+
+    let rec dexpr' (span : span) (expr : A.expr') : B.expr' =
+      match expr with
+      | [%inline_arms "dexpr'.*" - Match] -> auto
+      | Match { scrutinee; arms } ->
+          let arms = transform_arms (dexpr scrutinee) (List.rev arms) [] in
```
W95Psp, line 38: If you have `match <scrut> { ... if ... => ..., <other arms> }`, this would generate `match <scrut> { _ => match <scrut>} {...}}`, we should optimize that

engine/lib/phases/phase_drop_match_guards.ml
```diff
@@ -0,0 +1,200 @@
...
+                    is_record = false;
+                    is_struct = false;
+                  };
+              span = guard_span;
+              typ = opt_result_typ;
+            }
+          in
+
+          let expr_none = mk_opt_expr None in
+          let r : B.expr =
```
W95Psp, line 106: Can you add a comment here, with a example, explaining what AST we're building?

engine/lib/phases/phase_drop_match_guards.ml
```diff
@@ -0,0 +1,200 @@
+(* This phase removes guards from pattern matchings. It rewrites
```
W95Psp, line 1: We should include a succinct rewrite example here.

engine/lib/print_rust.ml
```diff
@@ -278,8 +278,14 @@ module Raw = struct
...
           List.map
-            ~f:(fun { arm = { arm_pat; body }; _ } ->
-              ppat arm_pat & !" => {" & pexpr body & !"}")
+            ~f:(fun { arm = { arm_pat; body; guard }; _ } ->
+              let g =
+                match guard with
+                | Some { guard = IfLet { lhs; rhs; _ }; _ } ->
+                    !" if let " & ppat lhs & !" = " & pexpr rhs
+                | None -> !""
+              in
```
W95Psp, line 287: Similar comments that on the generic printer: maybe use `Option.map`&`Option.value` and name `g` `guard`?

W95Psp (changes requested):
Thank you, the output looks like what we want!
However, we need:
 - more tests (e.g. a match where more than one arm as an `if`/`if let`, exercise `if`, exercise a non-trivial `if let` with a computation in the RHS... you don't need to have the equivalent desugar for everything)
 - some description on the PR. Your PR should:
    + explain the problem you solve (referring to an issue is fine, iff the issue has a nice and intelligible description about the exact problem we are solving, with examples --but in your case the issue has no description, please add some as requested in the issue)
    + then give an overview of technically what is our approach
    + then give examples of what one can do if your PR is merged (possibly with links to playground)
    + note that those points above should basically be copy pastes from code documentations or tests cases

engine/lib/ast_utils.ml
maximebuyse: I think it is not possible here as in the `None` branch the expression on which we call `shadows` is `body`. But maybe another simplification is possible

engine/lib/phases/phase_drop_match_guards.ml
maximebuyse: Yes, I also noticed that with tests. This phase can generate a lot of nested pattern matchings, some of them with only one Wildcard branch or other simplifiable situations. Do you think we should optimize in this phase (trying to simpler code in specific situations) or in another one (that would catch them and simplify)?

maximebuyse:
> Thank you, the output looks like what we want! However, we need:
> 
>     * more tests (e.g. a match where more than one arm as an `if`/`if let`, exercise `if`, exercise a non-trivial `if let` with a computation in the RHS... you don't need to have the equivalent desugar for everything)
> 
>     * some description on the PR. Your PR should:
>       
>       * explain the problem you solve (referring to an issue is fine, iff the issue has a nice and intelligible description about the exact problem we are solving, with examples --but in your case the issue has no description, please add some as requested in the issue)
>       * then give an overview of technically what is our approach
>       * then give examples of what one can do if your PR is merged (possibly with links to playground)
>       * note that those points above should basically be copy pastes from code documentations or tests cases

Thank you. I addressed most of your feedback but I can add more tests / documentation if you think it is not clear enough.

engine/lib/ast_utils.ml
W95Psp: Ah, right, that makes sense!

engine/lib/phases/phase_drop_match_guards.ml
W95Psp: I think we should do that in this phase: if a user writes a match like that in Rust, it's better not to rewrite it.
We should collapse matches only when a rewrite was done I think

maximebuyse:
> Thanks you, that's great! :) I don't know how you want to do this optimization for "one wildcard arm" matches: in this PR or in a subsequent one?

I'll do it in this one

maximebuyse:
> Thanks you, that's great! :) I don't know how you want to do this optimization for "one wildcard arm" matches: in this PR or in a subsequent one?

@W95Psp That's done now

engine/lib/phases/phase_drop_match_guards.ml
W95Psp: Most of the unnecessary `match ... { _ => ... }` seems to be gone :smiley: but if one appears at top-level, it seems like it's still there, e.g. https://hax-playground.cryspen.com/#fstar/9dc02d2817/gist=f490560c1ca7bb7a0e30f0e133fd8202
I guess you need to do something right there, when calling `transform_arms` in `dexpr'`, no?

maximebuyse: Good catch! Should be fixed now.

engine/lib/phases/phase_drop_match_guards.ml
```diff
@@ -60,12 +60,19 @@ module%inlined_contents Make (F : Features.T) = struct
...
+      (* General case *)
+      | _ -> Match { scrutinee; arms }
+
     let rec dexpr' (span : span) (expr : A.expr') : B.expr' =
       match expr with
       | [%inline_arms "dexpr'.*" - Match] -> auto
       | Match { scrutinee; arms } ->
           let arms = transform_arms (dexpr scrutinee) (List.rev arms) [] in
-          Match { scrutinee = dexpr scrutinee; arms }
+          maybe_simplified_match (dexpr scrutinee) arms
```
W95Psp, line 75: Sorry for being picky :grin: but can do `maybe_simplified_match` iff the user did not type a match with one arm with `_` LHS and no guard? otherwise such user typed matches are optimized away, e.g. https://hax-playground.cryspen.com/#fstar/365ad6f8bf/gist=be0c4a2ebe27079803f981fc39c45283

maximebuyse: No problem, I fixed that now!

W95Psp:
Thanks, I think that's great now! Let's merge that PR :)

Issue #624, opened 2024-04-24
Default method in trait blocks


W95Psp:

W95Psp:
Was fixed by https://github.com/hacspec/hax/pull/824

chrysn:
I'm still getting an error when running the latest hax version on the latest [lakers](https://github.com/openwsn-berkeley/lakers) main:

```
error: [HAX0008] (reject_TraitItemDefault) ExplicitRejection { reason: "a node of kind [Trait_item_default] have been found in the AST" }
  --> shared/src/crypto.rs:41:5
   |
41 |       fn supported_suites(&self) -> EdhocBuffer<MAX_SUITES_LEN> {
   |  _____^
42 | |         EdhocBuffer::<MAX_SUITES_LEN>::new_from_slice(&[EDHOCSuite::CipherSuite2 as u8])
43 | |             .expect("This should never fail, as the slice is of the correct length")
44 | |     }
   | |_____^
   |
```

Is this a regression on this issue, or is this a new one?

<details><summary><tt>cargo-hax --version</tt></summary>

```
hax 
version=0.3.1
commit=unknown
```

(built from 6254bb38f5dbb4c9459a13ccc82070bc4e98da4c)

</details>

PR #824, opened 2024-08-06
feat(engine): add feature for trait item defaults

### LINKED ISSUE #823: Importer currently does not import trait item defaults
If a trait item is given a default value/implementation, the importer will fail.

This can probably be feature gated.

paulmure:
Address #624 in the importer.

This PR adds a `TIDefault` variant to `trait_item'` in the AST to support trait items with default values/implementations.

Currently, only defaults in `const` and `fn` are supported where `const`s are encoded with a arity-0 function just like normal trait items.

Defaults for associated types are not currently supported as they are still an unstable feature in upstream rust [https://github.com/rust-lang/rust/issues/29661](https://github.com/rust-lang/rust/issues/29661).

This addition is feature gated with the `trait_item_default` feature and all backends have been adjusted to reject this feature for the moment.

### DIFF
--- a/engine/backends/coq/coq/coq_backend.ml
+++ b/engine/backends/coq/coq/coq_backend.ml
@@ -42,7 +42,8 @@ module SubtypeToInputLanguage
              and type quote = Features.Off.quote
              and type state_passing_loop = Features.Off.state_passing_loop
              and type dyn = Features.Off.dyn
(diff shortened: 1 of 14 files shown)


W95Psp (approved):
Thank you! That looks good to me, and that's a good thing to add support for, thanks! :smiley:

Issue #773, opened 2024-07-15
Pre/post `&mut` argument: remove `future` in favor of `old`


W95Psp:
- ~make the proc macro parse both `::hax_lib::_final` and `_final`~
 - ~detect if (locally) there is any name clash, and ask the user to use `::hax_lib::_final` instead~

We chatted about that, and we'll adopt the style of Verus / Prusti / [Rustc proposal](https://github.com/rust-lang/compiler-team/issues/759): use `old` to refer to the old value, and otherwise that's the future value.

W95Psp:
Still something we want, we need to schedule it before too much code is written in that style.

W95Psp:
I'll bump priority for that one, it's still important to get that done soon.

PR #826, opened 2024-08-07
Add unsafe versions of inlining macros

### LINKED ISSUE #825: Add macros for inlining F* (and other backends) of any type
The macro `hax_lib::fstar!` is for computationally irrelevant code, and thus forces the expression inlined to be of type `unit`.
In the context of a pre-condition, we're in a computationally irrelevant context, thus we should be able to use a variant of `hax_lib::fstar!` that doesn't force `unit`.

Roadmap:
 - [x] add a `fstar_unsafe` macro, that can produce any type;
 - [x] add an `use` item from `fstar_unsafe` to `fstar` in requires and ensures.

cc @karthikbhargavan @mamonet

W95Psp:
Closes #825.

(note: replace BACKEND by `fstar`, `coq`, etc. in the text of this PR)

This PR:
 - adds `hax_lib::inline_unsafe` function, hidden in documentation, that allows to inline BACKEND code of any type directly as an expression. This is an internal function, it not supposed to be used directly. ~It exists only when compiling through hax~ (see my last reverted commit). The body of this function is an `unreachable!`.
 - adds a `hax_lib_macro:BACKEND_unsafe_expr` macro (exposed as `hax_lib::BACKEND_unsafe_expr`, hidden in doc) which leverages `inline_unsafe` to produce inlined BACKEND code of any type.
 - exposes `hax_lib_macro:BACKEND_unsafe_expr` locally as `BACKEND` in pre and post conditions. Thus we can use e.g. `fstar!("forall x. ...")` in ensures and requires clauses.
 - Disallow `future(...)` within a quote (e.g. `fstar!("... ${future(arg)} ...")`): due to the way macros expand in a fixed sequence (the `fstar!("...")` expands after `requires` expands, so the `requires` macro cannot rewrite future nodes in `"..."` -- note that will be the same thing, but with `old` instead of `future`, when #773 will be fixed)
 - adds tests

Example of what you can do:
```rust
/// `ensures` and `requires` with inlined code (issue #825)
mod inlined_code_ensures_requires {
    #[hax_lib::requires(fstar!("forall i. FStar.Seq.index $v i <. ${254u8}"))]
    #[hax_lib::ensures(|()| {
        let future_v = future(v);
        fstar!("forall i. FStar.Seq.index ${future_v} i >. ${0u8}")
    })]
    fn increment_array(v: &mut [u8; 4]) {
        v[0] += 1;
        v[1] += 1;
        v[2] += 1;
        v[3] += 1;
    }
}
```
https://hax-playground.cryspen.com/#fstar/88b4821690/gist=49d80be217020e9fd41f9e0a6a23714e

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -556,6 +556,7 @@ version = "0.1.0-pre.1"
 dependencies = [
  "hax-lib",
  "hax-lib-macros-types",
+ "paste",
  "proc-macro-error",
  "proc-macro2",
  "quote",
--- a/engine/lib/phases/phase_transform_hax_lib_inline.ml
+++ b/engine/lib/phases/phase_transform_hax_lib_inline.ml
@@ -80,7 +80,8 @@ module%inlined_contents Make (F : Features.T) = struct
     and quote_of_expr' span (expr : A.expr') =
       match expr with
       | App { f = { e = GlobalVar f; _ }; args = [ payload ]; _ }
-        when Global_ident.eq_name Hax_lib__inline f ->
+        when Global_ident.eq_name Hax_lib__inline f
+             || Global_ident.eq_name Hax_lib__inline_unsafe f ->
           let bindings, str = dexpr payload |> UB.collect_let_bindings in
           let str =
             match
--- a/engine/names/src/lib.rs
+++ b/engine/names/src/lib.rs
@@ -51,6 +51,7 @@ fn dummy_hax_concrete_ident_wrapper<I: core::iter::Iterator<Item = u8>>(x: I, mu
     }
 
     let _ = hax_lib::inline("");
+    let _: () = hax_lib::inline_unsafe("");
     use hax_lib::{RefineAs, Refinement};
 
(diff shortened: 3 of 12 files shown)


karthikbhargavan (approved):
Very useful, but it croaks when you use `hax_lib::fstar` instead of `fstar` in the requires/ensures. Would this be easy to fix?

I see that `future` cannot be used within F*, which may also be confusing for people. In a future PR, let's figure out the recommended style for this, once we have some experience.

W95Psp:
Thanks!
 - for future, it is OK to use it with `fstar!`, you just need to put it in a let (the macro signals that to the user, if you fstar!("${future(x)}"), you get a message saying you need to put future(x) in a let binding an use the binding instead of future(x) directly, see https://hax-playground.cryspen.com/#fstar/88b4821690/gist=9b5b4923c0717cca519a323176fbe2b3)
 - I could make it work for hax_lib::fstar, but it is going to be a bit odd. Maybe I should rename the local fstar! of ensures/requires to something else?

mamonet (approved):

PR #830, opened 2024-08-07
Eliminate deep disjunctive patterns.

### LINKED ISSUE #463: (deep) disjunctive patterns in backend that don't support them (e.g. F*)
@franziskuskiefer wrote in [#462](https://github.com/hacspec/hax/issues/462)
> Disjunctive patterns aren't supported in the F* backend. They are supported since [#306](https://github.com/hacspec/hax/issues/306).
> 
> There is a test in https://github.com/hacspec/hax/blob/main/tests/pattern-or/src/lib.rs

This issue covers the case "deep" disjunctive patterns, eg:
```rust
fn f(x: &[u8]) -> u8 {
    match x {
        [] => 0,
        [(1|2|3),..] => 1,
        _ => 2
    }
}
```

# Roadmap
F* doesn't support such patterns, we need a phase that will turn such such a `match` into multiple nested ones.

maximebuyse:
Closes #463.

We add a new monomorphic phase that lifts up all deep disjunctive patterns so that the resulting patterns may only contain shallow disjunctions.

For example the pattern `(1 | 2, Some(3 | 4))` is transformed as `(1, Some(3)) | (1, Some(4)) | (2, Some(3)) | (2, Some(4))`

### DIFF
--- a/engine/lib/phases/phase_hoist_disjunctive_patterns.ml
+++ b/engine/lib/phases/phase_hoist_disjunctive_patterns.ml
@@ -19,7 +19,7 @@ module Make (F : Features.T) =
 
       let hoist_disjunctions =
         object (self)
-          inherit [_] Visitors.map as super
+          inherit [_] Visitors.map
 
           method! visit_pat () p =
             let return_pat p' = { p = p'; span = p.span; typ = p.typ } in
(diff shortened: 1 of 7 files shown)


engine/lib/phases/phase_hoist_disjunctive_patterns.ml
```diff
@@ -0,0 +1,115 @@
...
+            | PAscription { typ; typ_span; pat } ->
+                treat_subpat pat (fun pat -> PAscription { typ; typ_span; pat })
+            | PBinding { subpat = Some (pat, as_pat); mut; mode; typ; var } ->
+                treat_subpat pat (fun pat ->
+                    PBinding
+                      { subpat = Some (pat, as_pat); mut; mode; typ; var })
+            | PDeref { subpat; witness } ->
+                treat_subpat subpat (fun subpat -> PDeref { subpat; witness })
+            | PWild | PConstant _ | PBinding { subpat = None; _ } ->
+                super#visit_pat () p
```
W95Psp, line 111: We can just return `p`

W95Psp (approved):

PR #831, opened 2024-08-07
fix(backend/fstar): pre/post are `Type0` predicates, not boolean predicates

W95Psp:
For a method `f` in a trait, we generate two fields: `f_pre` and `f_post`. Those used to be boolean predicates, because we inherited the type from Rust. This commit changes that: we now enforce `f_pre` and `f_post` to be arrow type with `Type0` as last component.

### DIFF
--- a/engine/backends/fstar/fstar_backend.ml
+++ b/engine/backends/fstar/fstar_backend.ml
@@ -693,9 +693,13 @@ struct
       let ident = F.id ("x" ^ Int.to_string nth) in
       { kind = Explicit; ident; typ = pty span typ }
 
-    let of_named_typ span name typ : t =
+    (** Makes an F* binder from a name and an F* type *)
+    let of_named_fstar_typ span name typ : t =
       let ident = plocal_ident name in
-      { kind = Explicit; ident; typ = pty span typ }
+      { kind = Explicit; ident; typ }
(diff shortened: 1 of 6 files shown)


mamonet (approved):

karthikbhargavan:
This still needs a couple of changes. For now, the default value of the pre/post conditions is "true" not "True"
This causes F* type errors. E.g. see: https://github.com/cryspen/libcrux/blob/39a1178e4b0004fbb2a6dbdb36447e7be61643f4/libcrux-ml-kem/proofs/fstar/extraction/Libcrux_ml_kem.Hash_functions.Portable.fsti#L81

This causes the following type error
```
* Error 19 at Libcrux_ml_kem.Hash_functions.Portable.fsti(81,8-81,12):
  - Subtyping check failed
  - Expected type
      pred:
      Type0
        { pred ==>
          Rust_primitives.Integers.v v_LEN < Prims.pow2 32 ==>
          out == Spec.Utils.map_array (Spec.Utils.v_PRF v_LEN) input }
    got type Prims.logical
  - The SMT solver could not prove the query. Use --query_stats for more
    details.
  - See also Libcrux_ml_kem.Hash_functions.fsti(37,32-37,93)
```

PR #835, opened 2024-08-09
fix(engine/importer): use correct field to import concrete implexprs

### LINKED ISSUE #834: Engine: `Concrete` `ImplExpr`s are not imported correctly
Consider this example:
```rust
trait Foo {
    fn do_thing(self: &Self) -> u64;
}

struct Bar {
    n: u64,
}

impl Foo for Bar {
    fn do_thing(self: &Self) -> u64 {
        self.n
    }
}

fn call_foo<F: Foo>(foo: &F) -> u64 {
    foo.do_thing()
}

fn use_bar() -> u64 {
    let b = Bar { n: 42 };
    call_foo(&b)
}
```

One of the trait bounds in the call to `call_foo` in the function `use_bar` is the fact that `Bar` implements `Foo`.

This information is exported as an `ImplExpr::Concrete`. The exported json looks something like this:
```json
{
  "trait": {
    "def_id": {
      "krate": "scratch",
      "path": [
        {
          "data": {
            "TypeNs": "Foo"
          },
          "disambiguator": 0
        }
      ],
      "index": [
        0,
        3
      ]
    },
    "generic_args": [
      {
        "Type": {
          "Adt": {
            "generic_args": [],
            "trait_refs": [],
            "def_id": {
              "krate": "scratch",
              "path": [
                {
                  "data": {
                    "TypeNs": "Bar"
                  },
                  "disambiguator": 0
                }
              ],
              "index": [
                0,
                5
              ]
            }
          }
        }
      }
    ]
  },
  "impl": {
    "Concrete": {
      "id": {
        "krate": "scratch",
        "path": [
          {
            "data": "Impl",
            "disambiguator": 0
          }
        ],
        "index": [
          0,
          7
        ]
      },
      "generics": []
    }
  },
  "args": []
}
```

The information about what type was used to implement what trait is stored in the `trait` field, whereas the `impl` field stores some useless marker.

The importer incorrectly imports the `impl` field as the final trait goal. Resulting in a junk `impl_expr` like this:
```json
[
  "Concrete",
  {
    "trait": {
      "def_id": {
        "krate": "scratch",
        "path": [
          {
            "data": [
              "Impl"
            ],
            "disambiguator": 0
          }
        ]
      },
      "kind": [
        "Impl"
      ]
    },
    "args": []
  }
]
```

paulmure:
Close #834.

@W95Psp some of the existing insta tests are failing and I'm not knowledgeable enough about them to see if they need to be updated.

### DIFF
--- a/engine/backends/fstar/fstar_backend.ml
+++ b/engine/backends/fstar/fstar_backend.ml
@@ -343,7 +343,7 @@ struct
     let some = Option.some in
     let hax_unstable_impl_exprs = false in
     match ie with
-    | Concrete tr -> c_trait_goal span tr |> some
+    | Concrete { impl = tr; _ } -> c_trait_goal span tr |> some
     | LocalBound { id } ->
         let local_ident =
           Local_ident.{ name = id; id = Local_ident.mk_id Expr 0 }
--- a/engine/lib/ast.ml
+++ b/engine/lib/ast.ml
@@ -143,7 +143,7 @@ functor
 
     and impl_expr =
       | Self
-      | Concrete of trait_goal
+      | Concrete of { trait : trait_goal; impl : trait_goal }
       | LocalBound of { id : string }
       | Parent of { impl : impl_expr; ident : impl_ident }
       | Projection of {
--- a/engine/lib/import_thir.ml
+++ b/engine/lib/import_thir.ml
@@ -1026,7 +1026,7 @@ end) : EXPR = struct
   (* fun _ -> Ok Bool *)
 
   and c_impl_expr (span : Thir.span) (ie : Thir.impl_expr) : impl_expr =
-    let impl = c_impl_expr_atom span ie.impl in
+    let impl = c_impl_expr_atom span ie.trait ie.impl in
     match ie.args with
     | [] -> impl
     | args ->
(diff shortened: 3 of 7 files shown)


W95Psp:
Hi, thanks for your PR! Sorry for the delay, the week is quite busy... will try to look into those tests tomorrow!

W95Psp (approved):
Sorry for the delay... LGTM.

W95Psp:
(forgot to push, doing that in a minute)

W95Psp:
Ok, I'm in a rush and so I rushed the review as well... I wanted to merge too fast, and actually I think there is a problem with the PR. Let me comment on the issue.

paulmure:
@W95Psp
As discussed in the [issue](https://github.com/hacspec/hax/issues/834#issuecomment-2324987173), I'm trying to import the trait goal alongside the information stored in `ImplExprAtom`.

However, this seems to break the uses of `ppx_inline` in many of the modules in `lib/phases`. I guess it's caused by incorrect inlining of the code in `subtype.ml`. And I'm not sure how to fix this.

Any suggestions would be greatly appreciated!

paulmure:
> @W95Psp As discussed in the [issue](https://github.com/hacspec/hax/issues/834#issuecomment-2324987173), I'm trying to import the trait goal alongside the information stored in `ImplExprAtom`.
> 
> However, this seems to break the uses of `ppx_inline` in many of the modules in `lib/phases`. I guess it's caused by incorrect inlining of the code in `subtype.ml`. And I'm not sure how to fix this.
> 
> Any suggestions would be greatly appreciated!

Forget what I said here, I think I just needed to run `dune clean` before building.

paulmure:
After some back and forth, I decided to only import the trait goal for the `concrete` case for now.

I'm not sure if it's actually needed for the other cases. If such a need arises in the future, we can reconsider how to import this information again.

All tests should pass now. The only change is that the traits test for f* added a line of implicit dependency.

W95Psp:
Hi @paulmure, I'm reviewing this PR this morning, and I think I did something similar recently... https://github.com/hacspec/hax/pull/827
Sorry for the duplicate work... (the last month has been pretty work intensive toward a verification project...)

Wdyt, shall we close this one or is it still needed?

paulmure:
That looks like the information we need. Thank you!

PR #838, opened 2024-08-12
Bump rustc version

Nadrieril:
Bump the version of rustc by 7 weeks, to get the benefits of https://github.com/rust-lang/rust/pull/127028 and https://github.com/rust-lang/rust/pull/128815.

### DIFF
--- a/frontend/exporter/src/types/copied.rs
+++ b/frontend/exporter/src/types/copied.rs
@@ -1686,7 +1686,7 @@ pub struct GenericParamDef {
         match self.kind {
             ty::GenericParamDefKind::Lifetime => GenericParamDefKind::Lifetime,
             ty::GenericParamDefKind::Type { has_default, synthetic } => GenericParamDefKind::Type { has_default, synthetic },
-            ty::GenericParamDefKind::Const { has_default, is_host_effect } => {
+            ty::GenericParamDefKind::Const { has_default, is_host_effect, .. } => {
                 let ty = s.base().tcx.type_of(self.def_id).instantiate_identity().sinto(s);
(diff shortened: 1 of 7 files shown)


frontend/exporter/src/types/copied.rs
```diff
@@ -2982,6 +2983,25 @@ pub struct AnonConst<Body: IsBody> {
...
+}
+
+/// Reflects [`rustc_hir::ConstArgKind`]
+#[derive_group(Serializers)]
+#[derive(AdtInto, Clone, Debug, JsonSchema)]
+#[args(<'tcx, S: UnderOwnerState<'tcx>>, from: rustc_hir::ConstArgKind<'tcx>, state: S as s)]
+pub enum ConstArgKind<Body: IsBody> {
+    Path(QPath),
+    Anon(AnonConst<Body>),
+}
```
Nadrieril, line 3003: We could get rid of all that by using `ty::GenericParamDefKind` instead of `hir::GenericParamKind`. I'm generally in favor of using things in `ty` rather then their HIR equivalent since the `ty` stuff is more processed.

W95Psp (approved):
LGTM!
I don't know for `GenericParamDefKind` :thinking: you're probably right, can you open an issue or something? let's chat about that later and see, this week is a bit too busy :/

Nadrieril:
We can talk about `GenericParamDefKind` etc once you start using `FullDef` for the engine. For now it's not hard to maintain, we can keep it.

Nadrieril:
Opened https://github.com/hacspec/hax/issues/852 to discuss it

W95Psp:
Great, let's merge then!

PR #841, opened 2024-08-13
fix(engine/fstar): add `string` to the list of protected names

### LINKED ISSUE #839: function argument clashes with type name
```rust
fn f(string:&str, n:&str) {
}
```
[Open this code snippet in the playground](https://hax-playground.cryspen.com/#fstar/46df00505b/gist=ee9578777fe9f1172adb77819f335711)

Since `&str` is translated to `string` in F*, this results in a name clash (F* assumes that the second `&str` is a dependent type referring to the first argument).

W95Psp:
Fixes #839


karthikbhargavan (approved):
Thanks!

PR #848, opened 2024-08-13
Add some more information to `FullDef`

Nadrieril:
Some small additions to `FullDef` for use in charon.


engine/lib/import_thir.ml
```diff
@@ -109,12 +109,12 @@ let c_attr (attr : Thir.attribute) : attr =
...
+            { args = Eq (_, Hir { symbol; _ }); path = "doc"; tokens = None; _ };
           tokens = None;
         } ->
         DocComment { kind = DCKLine; body = symbol }
         (* Looks for `#[doc = "something"]` *)
-    | Normal { item = { args; path; tokens = subtokens }; tokens } ->
+    | Normal { item = { args; path; tokens = subtokens; _ }; tokens } ->
         let args_tokens =
           match args with Delimited { tokens; _ } -> Some tokens | _ -> None
         in
```
Nadrieril, line 120: `as_str` might be able to replace the various `tokens`?

frontend/exporter/src/types/copied.rs
```diff
@@ -2366,6 +2379,15 @@ pub struct AttrItem {
     pub path: String,
     pub args: AttrArgs,
     pub tokens: Option<TokenStream>,
+    /// String representation of the item. This may not exactly match what the user wrote because
```
W95Psp, line 2382: This is redundant with `tokens`, right? I guess it's because it's nicer to process?

W95Psp (approved):
Sorry for the long delay, that looks good to me!
I have only an interrogation about `as_str`, I wonder if we want that, in general more structured data is nicer, I don't know...

frontend/exporter/src/types/copied.rs
Nadrieril: Sometimes `tokens` is `None` and `as_str` isn't

PR #851, opened 2024-08-14
fix(engine): expose type of local_ident.id

### LINKED ISSUE #850: Engine: `local_ident`'s `id` type is opaque
The interface for `local_ident` in the engine is defined like this:
```ocaml
  type id [@@deriving show, yojson, hash, compare, sexp, eq]

  val mk_id : kind -> int -> id

  type t = { name : string; id : id }
  [@@deriving show, yojson, hash, compare, sexp, eq]
```
The `id` type is opaque so backends cannot inspect the `kind` of a `local_ident`.

Without this information, certain things become impossible.

paulmure:
Close #850


W95Psp (approved):
As discussed in https://github.com/hacspec/hax/issues/850, that looks ok to me, let's merge.

PR #854, opened 2024-08-16
Simplify the unsize definition

karthikbhargavan:
This PR switches the definition of the unsize trait and type in the F* library to become the identity function.
This matches what we currently do and significantly improves verification performance in ML-KEM.


W95Psp:
Closing in favor of #856

PR #855, opened 2024-08-17
fix(engine/backend): expose raw view on concrete_ident

### LINKED ISSUE #793: Raw view on `concrete_ident`
Currently, `concrete_ident` is exported to engines as an opaque type. To get its underlying content, one would need to make a `VIEW_API` which transform the underlying path and definition into a view.

While the `VIEW_API` can accept options such as a `NAME_POLICY`, it seems intractable to have an API that is flexible to accommodate all possible name transformation schemes. For example, there is currently no way to configure the prefix scheme built-in to the default view api.

It seems appropriate to just export the raw underlying content without any transformations, giving each engine the option to have full control over how to translate `concrete_ident`s.

paulmure:
Close #793

### DIFF
--- a/engine/lib/concrete_ident/concrete_ident.mli
+++ b/engine/lib/concrete_ident/concrete_ident.mli
@@ -1,10 +1,28 @@
-type t [@@deriving show, yojson, compare, sexp, eq, hash]
+module Imported : sig
+  type def_id = { krate : string; path : path }
+  and path = disambiguated_def_path_item list
 
-type name = Concrete_ident_generated.t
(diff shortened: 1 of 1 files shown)


franziskuskiefer:
@W95Psp what's up with this. Review has been waiting for 2 months now.

W95Psp:
As discussed in #793, instead of exposing a raw view on `concrete_ident`, we prefer to improve the API on names to allow more flexibility.
Let's close for now, and try to come up with a revisit of the concrete_ident API.

PR #865, opened 2024-08-30
i16 funs

karthikbhargavan:


W95Psp (approved):
LGTM, but why not defining `impl__i16__pow` directly, instead of having a post condition here?

PR #877, opened 2024-09-04
Avoid recursing twice on the arguments.

### LINKED ISSUE #875: Stack overflow for deep expressions in bincode serialization/deserialization
With rust code containing deep expressions, hax fails with a stack overflow. For example:
```
fn main() -> i32 {
        1 + 1 + 1 + 1 + 1 + 1 + 1 + 
        1 + 1 + 1 + 1 + 1 + 1 + 1 + 
        1 + 1 + 1 + 1 + 1 + 1 + 1 + 
        1 + 1 + 1 + 1 + 1 + 1 + 1
}
```
The stack trace suggests that this has to do with serialization/deserialization with bincode that was introduced in #743. However this problem cannot be reproduced with the merge commit of #743 itself (but it loops for a long time). The version of bincode hasn't changed since then so there might be an interaction with another later change that triggers the stack overflow.

maximebuyse:
Fixes #875 

The recursive call on the arguments was done twice in some cases which leads to an exponential complexity. The fix is just to avoid the double recursive call.

### DIFF
--- a/engine/lib/phases/phase_specialize.ml
+++ b/engine/lib/phases/phase_specialize.ml
@@ -113,7 +113,14 @@ module Make (F : Features.T) =
 
           method! visit_expr () e =
             match e.e with
-            | App { f = { e = GlobalVar f; _ } as f'; args = l; _ } -> (
+            | App
+                {
+                  f = { e = GlobalVar f; _ } as f';
+                  args = l;
+                  trait;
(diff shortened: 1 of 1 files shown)


franziskuskiefer (approved):
The produced F* doesn't look very nice. But that's a separate issue.

PR #880, opened 2024-09-05
Support float literals

R1kM:
This PR adds support for float literals, such as `1.34f32`.

Note, I wasn't sure how to add tests directly in Hax for this feature, so tests are currently in Charon only.
I'm happy to add some here if somebody can give me a pointer.

### DIFF
--- a/frontend/exporter/src/constant_utils.rs
+++ b/frontend/exporter/src/constant_utils.rs
@@ -75,7 +75,10 @@ pub enum ConstantExprKind {
         impl_expr: ImplExpr,
         name: String,
     },
+    /// A shared reference to a static variable.
     Borrow(ConstantExpr),
+    /// A `*mut` pointer to a static mutable variable.
+    MutPtr(ConstantExpr),
     ConstRef {
         id: ParamConst,
     },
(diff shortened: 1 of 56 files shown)


frontend/exporter/src/constant_utils.rs
```diff
@@ -255,6 +259,52 @@ mod rustc {
...
+                    ty::FloatTy::F16 => {
+                        let f = scalar.to_f16().unwrap_or_else(|_| {
+                            fatal!(
+                                s[span],
+                                "Type is primitive, but the scalar {:#?} is not a [Float]",
+                                scalar
+                            )
+                        });
+                        rustc_apfloat::ieee::Half::to_bits(f)
+                    }
```
Nadrieril, line 274: You can go directly from the scalar to its bits without round-tripping through a structured representation: `scalar.assert_scalar_int().to_bits_unchecked()`

R1kM: Thanks! Modified in https://github.com/hacspec/hax/pull/880/commits/005594778f2aff4111e4f47fd893f52e29f84263, I went through `try_to_scalar_int` to provide a better error message

frontend/exporter/src/constant_utils.rs
```diff
@@ -255,6 +259,19 @@ mod rustc {
                 });
                 ConstantExprKind::Literal(scalar_int_to_constant_literal(s, scalar_int, ty))
             }
+            ty::Float(float_type) => {
+                let scalar_int = scalar.try_to_scalar_int().unwrap_or_else(|_| {
+                    fatal!(
+                        s[span],
+                        "Type is primitive, but the scalar {:#?} is not a [Float]",
```
Nadrieril, line 266:
```suggestion
                        "Type is [Float], but the scalar {:#?} is not a number",
```

R1kM:
It is unclear why the ocamlformat CI target is failing, this seems more related to infrastructure (it exits 1 when setting up the Nix environment, possibly because the CI machine is out of memory/disk space?)
I confirmed locally that the diff on the modified file is empty with the latest commit.

PR #882, opened 2024-09-10
fix(fstar-core): add support for `Option::map`

ROMemories:
This is an attempt at adding support for `Option::map` in the F* model of the Rust core library.
The model is copied from the existing one of `Option::and_then`, with the return type of the closure being adapted.

Is there any additional integration/unit tests to write for these models?


maximebuyse:
The implementation looks correct to me.

There are no tests for fstar-core AFAIK.

franziskuskiefer:
> There are no tests for fstar-core AFAIK.

Unfortunately not yet.

PR #884, opened 2024-09-11
Fix incorrect line numbers in error messages

### LINKED ISSUE #879: Incorrect line numbers in error messages
The line numbers on extraction errors don't seem to be correct in all cases.
For example, running `cargo hax into fstar` in the `libcrux-ml-dsa` crate at [this commit](https://github.com/cryspen/libcrux/pull/558/commits/7e674c6d82f120eb174ea98314e356613a85c980) results in a number of errors, e.g.:
```
error: [HAX0010] (RefMut) At this position, Hax was expecting an expression of the shape `&mut _`. Hax forbids `f(x)` (where `f` expects a mutable reference as input) when `x` is not a place expression[1] or when it is a dereference expression.

[1]: https://doc.rust-lang.org/reference/expressions.html#place-expressions-and-value-expressions

  --> libcrux-ml-dsa/src/encoding/signature.rs:95:1
   |
95 |             hint_serialized[MAX_ONES_IN_HINT + i] = true_hints_seen as u8;
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
```
and
```
error: [HAX0010] (RefMut) At this position, Hax was expecting an expression of the shape `&mut _`. Hax forbids `f(x)` (where `f` expects a mutable reference as input) when `x` is not a place expression[1] or when it is a dereference expression.

[1]: https://doc.rust-lang.org/reference/expressions.html#place-expressions-and-value-expressions

  --> libcrux-ml-dsa/src/encoding/signature.rs:87:1
   |
87 |                     hint_serialized[true_hints_seen] = j as u8;
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
```

but these two lines are lines [44](https://github.com/cryspen/libcrux/blob/7e674c6d82f120eb174ea98314e356613a85c980/libcrux-ml-dsa/src/encoding/signature.rs#L44) and [48](https://github.com/cryspen/libcrux/blob/7e674c6d82f120eb174ea98314e356613a85c980/libcrux-ml-dsa/src/encoding/signature.rs#L48) in `signature.rs`, respectively.

maximebuyse:
Fixes #879 

The crate `annotate-snippets` that we use for reporting error messages has a parameter `line_start` that corresponds to the line number of the first line that we pass (the source code is passed as a string in another parameter). In our case we always give the full source code so `line_start` should be 1 (and not the line number at which the error is located).

### DIFF
--- a/hax-types/src/diagnostics/report.rs
+++ b/hax-types/src/diagnostics/report.rs
@@ -74,20 +74,21 @@ impl Diagnostics {
                 let start = compute_offset(&source, span.lo.line, span.lo.col);
                 let end = compute_offset(&source, span.hi.line, span.hi.col);
                 let origin = format!("{}", path.display());
-                snippets_data.push((source, origin, span.lo.line, start..end));
+                snippets_data.push((source, origin, start..end));
             };
         }
 
         let title = format!("[{}] {self}", self.kind.code());
-        let message = level.title(&title).snippets(snippets_data.iter().map(
-            |(source, origin, line, range)| {
-                Snippet::source(source)
-                    .line_start(*line)
-                    .origin(&origin)
-                    .fold(true)
-                    .annotation(level.span(range.clone()))
-            },
-        ));
+        let message =
+            level
+                .title(&title)
+                .snippets(snippets_data.iter().map(|(source, origin, range)| {
+                    Snippet::source(source)
+                        .line_start(1)
+                        .origin(&origin)
+                        .fold(true)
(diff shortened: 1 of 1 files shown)


jschneider-bensch (approved):
Ah, the docs for `annotate_snippet` are really a bit sparse. This looks good!

franziskuskiefer (approved):
Thanks for the quick fix!

PR #887, opened 2024-09-13
Initial cleanup of binders-related code

Nadrieril:
This is a bunch of small fixes we made while working together with @W95Psp on the binders issue. I kept the non-controversial changes and cleaned them up.


Nadrieril:
The charon breakage is expected (I changed a type), the fix is straightforward.

W95Psp (approved):
Great, thanks for packing those changes, let's merge before it's awful to rebase :sweat_smile:

PR #891, opened 2024-09-16
fix: fstar: default trait

### LINKED ISSUE #585: Core library: F*: support for the `Default` trait
hax [doesn't seem to support](https://github.com/future-proof-iot/RIOT-rs/actions/runs/8372464440/job/22923475247?pr=205) Rust's [`Default` trait](https://doc.rust-lang.org/stable/std/default/trait.Default.html). Would it be possible to add support for it?

_Originally posted by @ROMemories in https://github.com/future-proof-iot/RIOT-rs/pull/205#discussion_r1534117627_

W95Psp:
Fixes the definition of the `default` trait in the F* core lib.
The previous definition was incorrect: names were wrong, and pre/post fields were missing.

Fixes #585.

### DIFF
--- a/proof-libs/fstar/core/Core.Default.fsti
+++ b/proof-libs/fstar/core/Core.Default.fsti
@@ -1,5 +1,7 @@
 module Core.Default
 
 class t_Default (t: Type0) = {
-  v_default: unit -> t;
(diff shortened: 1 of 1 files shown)


franziskuskiefer (approved):
Thanks!

PR #892, opened 2024-09-17
Support other variants of MIR

Nadrieril:
Intermediate variants of MIR involve promoted constants, which are normal expressions turned into anonymous constants, which is a necessary step for cases like:
```rust
fn static_lifetime() -> &'static u32 {
    let x = &0;
    x
}
```
Hax crashes on them today (the call to `ucv.shrink()` panics); this PR fixes that.

### DIFF
--- a/frontend/exporter/src/body.rs
+++ b/frontend/exporter/src/body.rs
@@ -104,7 +104,7 @@ mod module {
                         did.to_def_id(),
                     ))
                 });
-                mir.unwrap()
+                mir.s_unwrap(s)
             }
         }
     }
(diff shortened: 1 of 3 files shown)


frontend/exporter/src/body.rs
```diff
@@ -95,8 +95,16 @@ mod module {
...
+                let mir = MirKind::get_mir(s.base().tcx, did, |body| {
+                    let body = Rc::new(body.clone());
+                    body.sinto(&with_owner_id(
+                        s.base(),
+                        thir,
+                        body.clone(),
+                        did.to_def_id(),
+                    ))
+                });
+                mir.unwrap()
```
W95Psp, line 107: Can you use `s_unwrap` here? If the unwrap fails, that will show a spanned error.
(i.e. `mir.s_unwrap(s)`)

W95Psp (approved):
Thanks, that looks good to me, I have one comment in the code.

Also, we should add support in the CLI of hax for outputting those various MIR variants. But I can do that; it's done by some weird code (macro `with_kind_type` in `hax-types/src/driver_api.rs`). I can add an issue.

PR #894, opened 2024-09-18
Detangle trait solving from `SInto`

Nadrieril:
This PR makes it so trait solving never calls into `SInto`, as that was causing headaches with binders. There is still more work to do to clarify trait solving but that's less of a priority.


Nadrieril:
As usual, the charon failure is expected, I have a fix ready

W95Psp (approved):
Thank you, that's really great! :smiley:
Let's merge that! :shipit:

PR #898, opened 2024-09-19
Fix some binder-related crashes

Nadrieril:


frontend/exporter/src/traits.rs
```diff
@@ -466,13 +479,15 @@ pub mod rustc {
             }
             .with_args(impl_exprs(tcx, owner_id, &nested)?, *tref),
             ImplSource::Param(nested) => {
+                let nested = impl_exprs(tcx, owner_id, &nested)?;
                 let predicates = tcx.predicates_defined_on_or_above(owner_id);
                 let Some((path, apred)) =
                     search_clause::path_to(tcx, &predicates, tref.clone(), param_env)
                 else {
-                    return Err(format!(
+                    return Ok(ImplExprAtom::Error(format!(
```
W95Psp, line 487: Should we display a warning here as well?

W95Psp (approved):
LGTM, thanks :D 
I only have one remark on a potential other warning to add

PR #901, opened 2024-09-23
Determine emuerate value by index

mamonet:
In the loop body of `fold_enumerated_slice` I noticed that enumerate value is not equal to array element which is necessary for some use cases. An example of use case is here https://github.com/cryspen/libcrux/blob/dev-serialize/libcrux-ml-kem/proofs/fstar/extraction/Libcrux_ml_kem.Ind_cpa.fst#L218 where the enumerate value `re` has to be equal to the index element of `input` array.


W95Psp (approved):
LGTM, thanks!

PR #906, opened 2024-09-23
Cleanup trait resolution

Nadrieril:
This PR methodically cleans up trait resolution. It also tracks the index of locally-bound predicates for use in Charon. This solves many binder- and trait-resolution-related issues in Charon.


W95Psp (approved):
I looked into the PR a bit fast (I don't have much time), but that looks great to me, thank you!

PR #907, opened 2024-09-23
Add RewriteControlFlow phase.

### LINKED ISSUE #393: Add a `RewriteControlFlow` phase (`return`/`continue`/`break`)
We want a phase that eliminates explicit control flow (say a `return` or a `break`) by moving or duplicating code.

We want to rewrite the control flow so that every `return`, `break` or `continue` always appears in a return position in our AST. Returns in return positions can be dropped: e.g. `if ... then {return 3;} else {return 5;}` can be simplified in `if ... then {3} else {5}`.

 - [x] rewrite control flow for `return`s appearing outside of a loop (this was PR #907);
 - [ ] rewrite control flow for `return`, `break` and `continue` within a loop.

See *Details* below for a lengthy explanation.

<details>
  <summary>Details</summary>
  
[Question marks](https://doc.rust-lang.org/rust-by-example/std/result/question_mark.html) (but also `return` and friends) used to be transformed in a monadic style.

In this style, computations are promoted to a `Result` or `Option` monad, and `pure` and `bind` nodes are inserted heavily.

For instance, let's consider the following function:
```rust
fn test(x: Option<u8>) -> Option<u8> {
    Some(f(x?) + 4)
}
```

This would be transformed to this monadic F* function, where `(let*)` is basically a `bind` node in the option monad:
```fstar
let test (x: option u8) : option u8 =
  run_monad (
    let* hoist1:u8 = x in
      Some
      (let hoist2:u8 = f hoist1 in
        let hoist3:u8 = hoist2 + 4 in
        Some hoist3 <: option u8)
      <:
      option (option u8))
```

While this is a correct translation, lifting everything to a monad and running the monad is a bit heavy.

Instead, the control flow of the Rust function `test` can be rewritten into:
```rust
match x {
   Some(x) => Some(f(x) + 4),
   None => None
}
``` 

Rewriting control flow is about pushing `?`s to return positions. When `?` appears deep in a `if`/`match` node itself nested in a sequential expression (e.g. a sequence or a function call), then such rewriting can result in code duplication. Example:
```rust
fn fun(cond: bool, x: Option<u8>) -> Option<u8> {
    let b = x + 10;
    let a = if b > 20 { x? + long_code } else { 0 };
    let b = other_long_code;
    Some(a + b + 3)
}

fn fun_rewrite_step(x: Option<u8>) -> Option<u8> {
    let b = x + 10;
    if b > 20 {
        let a = x? + long_code;
        let b = other_long_code;
        Some(a + b + 3)
    } else {
        let b = other_long_code;
        Some(a + b + 3)
    };
}

fn fun_rewrited(x: Option<u8>) -> Option<u8> {
    let b = x + 10;
    if b > 20 {
        match x {
            Some(x) => {
                let a = x? + long_code;
                let b = other_long_code;
                Some(a + b + 3)
            }
            None => None,
        }
    } else {
        let b = other_long_code;
        Some(a + b + 3)
    }
}
```

  
</details>

maximebuyse:
Closes #393

### DIFF
--- a/engine/lib/phases/phase_rewrite_control_flow.ml
+++ b/engine/lib/phases/phase_rewrite_control_flow.ml
@@ -34,143 +34,75 @@ module Make (F : Features.T) =
           method! visit_expr () e =
             match e.e with
             | _ when not (has_return#visit_expr () e) -> e
-            | Let
-                {
-                  monadic = None;
-                  lhs;
-                  rhs = { e = If { cond; then_; else_ }; _ } as rhs;
-                  body;
-                } ->
-                let cond = self#visit_expr () cond in
-                let then_has_return = has_return#visit_expr () then_ in
-                let else_has_return =
-                  Option.map else_ ~f:(has_return#visit_expr ())
-                  |> Option.value ~default:false
+            (* Returns in loops will be handled by issue #196 *)
+            | Loop _ -> e
+            | Let _ -> (
+                (* Collect let bindings to get the sequence
+                   of "statements", find the first "statement" that is a
+                   control flow containing a return. Rewrite it.
+                *)
+                let stmts, final = U.collect_let_bindings e in
+                let before_after i =
+                  let stmts_before, stmts_after = List.split_n stmts i in
+                  let stmts_after = List.tl_exn stmts_after in
+                  (stmts_before, stmts_after)
                 in
-                let rewrite = then_has_return || else_has_return in
-                if rewrite then
-                  let then_ =
-                    {
-                      e with
-                      e = Let { monadic = None; lhs; rhs = then_; body };
-                    }
+                let inline_in_branch branch p stmts_after final =
+                  let branch_stmts, branch_final =
+                    U.collect_let_bindings branch
                   in
-                  let then_ = self#visit_expr () then_ in
-                  let else_ =
-                    Some
-                      (match else_ with
-                      | Some else_ ->
-                          self#visit_expr ()
-                            {
-                              e with
-                              e = Let { monadic = None; lhs; rhs = else_; body };
-                            }
-                      | None -> body)
-                  in
-
-                  { rhs with e = If { cond; then_; else_ } }
-                else
-                  let body = self#visit_expr () body in
-                  {
-                    e with
-                    e =
-                      Let
-                        {
-                          monadic = None;
-                          lhs;
-                          rhs = { rhs with e = If { cond; then_; else_ } };
-                          body;
-                        };
-                  }
-            (* We need this case to make sure we take the `if` all the way up a sequence of nested `let`
-               and not just one level. *)
-            | Let { monadic = None; lhs; rhs = { e = Let _; _ } as rhs; body }
-              -> (
-                let body = self#visit_expr () body in
-                match self#visit_expr () rhs with
-                | { e = If { cond; then_; else_ = Some else_ }; _ } ->
-                    (* In this case we know we already rewrote the rhs so we should take the `if` one level higher. *)
-                    let rewrite_branch branch =
-                      {
-                        branch with
-                        e = Let { monadic = None; lhs; rhs = branch; body };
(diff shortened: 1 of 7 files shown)


maximebuyse:
@W95Psp I refactored like we discussed using `Ast_utils.collect_let_bindings` to treat directly with a sequence of "statements". This is indeed simpler and seems to work well.

engine/lib/phases/phase_rewrite_control_flow.ml
```diff
@@ -34,143 +34,75 @@ module Make (F : Features.T) =
...
-                let body = self#visit_expr () body in
-                match rhs.e with
-                | Return _ -> rhs
-                | _ -> { e with e = Let { monadic = None; lhs; rhs; body } })
+                match
+                  List.findi stmts ~f:(fun _ (_, e) ->
+                      match e.e with
+                      | (If _ | Match _) when has_return#visit_expr () e -> true
+                      | Return _ -> true
+                      | _ -> false)
```
W95Psp, line 63: I think here you can use `split_while`, so that you don't manipulate indexes explicitly and you get rid of `before_after`.

W95Psp (changes requested):

engine/lib/phases/phase_rewrite_control_flow.ml
```diff
@@ -0,0 +1,103 @@
...
+                let stmts, final = U.collect_let_bindings e in
+                let inline_in_branch branch p stmts_after final =
+                  let branch_stmts, branch_final =
+                    U.collect_let_bindings branch
+                  in
+                  U.make_lets
+                    (branch_stmts @ ((p, branch_final) :: stmts_after))
+                    final
+                in
+                match
```
W95Psp, line 53: I'm nitpicking a little, but maybe you should put that split_while in a let binding so that you can extract `stmts_before` before matching? (since you don't actually match on `stmts_before`)

W95Psp (changes requested):
Looks great to me, thank you!
I have one small code remark, but also, as we discussed yesterday, this phase sometimes adds extra `let _ = () in`. Can you try to spot where those are produced and prevent it?
For an example see https://hax-playground.cryspen.com/#fstar/595d6dddc9/gist=9c4e9735d9e4055b3407c81bc4238e8f

maximebuyse:
> Looks great to me, thank you! I have one small code remark, but also, as we discussed yesterday, this phase sometimes adds extra `let _ = () in`. Can you try to spot where those are produced and prevent it? For an example see https://hax-playground.cryspen.com/#fstar/595d6dddc9/gist=9c4e9735d9e4055b3407c81bc4238e8f

I fixed that. I noticed that in your example there is also a `let _ = 3`. Should we remove that as well?

W95Psp:
Yeah, the `3` is useless, and since all side effects are moved away, I would say such values will always be useless.
However, since the user typed it, I'm in favor of keeping it: in general I think it's a great thing to keep user inputs, even if the input make no much sense.
So let's keep that as it is!

engine/lib/phases/phase_rewrite_control_flow.mli
```diff
@@ -0,0 +1,6 @@
+(** This phase finds control flow expression (`if` or `match`) with a `return` expression
+    in one of the branches. We replace them by replicating what comes after in all the branches. 
+    This allows the `return` to be eliminated by `drop_needless_returns`.
+    This phase should come after phase_local_mutation. *)
```
W95Psp, line 4: Can you add ticks around `phase_local_mutation`?

W95Psp (approved):
Thanks, I have one tiny tiny stupid remark (sorry for that :sweat_smile:), and then let's merge it!

engine/lib/phases/phase_rewrite_control_flow.mli
maximebuyse: Done!

PR #908, opened 2024-09-23
fix(engine/deps): stop re-computing over and over the set of assoc items

W95Psp:
While debugging some bug, I noticed the dependency analysis was very slow, it was basically about precomputing the set of associated items instead of computing it over and over.


franziskuskiefer (approved):
🚀

PR #935, opened 2024-10-01
Handle cyclic module dependencies

### LINKED ISSUE #396: Add utility function to deal with mutual dependency across different modules
(this issue was extracted from https://github.com/hacspec/hax/issues/203#issuecomment-1637539912)

Rust modules are really about namespacing. Rust allows items to be mutually recursive **across** modules.

This is rarely allowed in other statically typed languages, and even less in proof assistants. Thus, we need a strategy for removing such recursive bundles that span across multiple modules.

# Example
## Situation A
```rust
mod a {
    pub fn f (x: u8) -> u8 {
        if x < 0 { x } else { super::b::f(x - 1) }
    }
}
mod b {
    pub fn f (x: u8) -> u8 {
        if x < 0 { x } else { super::a::f(x - 1) }
    }
}
```

In this case, we should probably transform to:
```rust
mod bundle_a_b {
    pub fn a_f (x: u8) -> u8 {
        if x < 0 { x } else { b_f(x - 1) }
    }
    pub fn b_f (x: u8) -> u8 {
        if x < 0 { x } else { a_f(x - 1) }
    }
}
mod a { pub fn f (x: u8) -> u8 { super::bundle_a_b::a_f(x) } }
mod b { pub fn f (x: u8) -> u8 { super::bundle_a_b::b_f(x) } }
```

## Situation B
We can also write something like:
```rust
fn f() {}
mod b {
    use super::*;
    pub fn g() {super::f ()}
}
fn h() { b::g() }
```

Here we have:
 - a dependency from `b` to `a` (`a::b::g` uses `a::f`);
 - a dependency from `a` to `b` (`f::h` uses `a::b::g`)

In this case, we can either:
 - make a bundle for `f`, `g` and `h` just like above;
 - make a bundle only for `f` and `g` and use that in the rest of the top-level module.

# Solution
We want our module-level dependency graph to be a tree.
For this, we should identify every group of items that span across multiple modules and isolate them in newly named modules.

This is partially implemented by `engine/lib/dependencies.ml`, but we need to revisit it and test it heavily.
This issue is blocking for extracting Core to anything (F*/Coq/whatever) useful.

maximebuyse:
Closes #396 

The idea is to put together objects that create a cyclic dependency at the module level. We bundle them in a newly created module (one for each cycle) and then use https://github.com/FStarLang/FStar/pull/3369 to selectively include these objects in the module they actually belong to.
List of changes compared to the existing implementation (that was present in `dependencies.ml` but not used by any backend):
- we solve name clashes in the bundles (by adding a hash to the names that clash)
- we replace the `let` and `type` keywords by `and` for mutually recursive items
- we replace the `unfold let` by `include` to make use of https://github.com/FStarLang/FStar/pull/3369
- we bundle not only mutually recursive items but also items that create a cyclic dependency without being recursive. This is done by computing the transitive closure of the item dependency graph. Looking for paths of three objects a1 -> b -> a2 such that a1, a2 belong to module A and b belongs to B (with B != A). Adding an edge between these items for each such path in a new graph. After finding all such paths, the bundles are exactly the connected components of this new graph.

### DIFF
--- a/engine/backends/coq/coq/coq_backend.ml
+++ b/engine/backends/coq/coq/coq_backend.ml
@@ -679,8 +679,8 @@ let hardcoded_coq_headers =
    Open Scope Z_scope.\n\
    Open Scope bool_scope.\n"
 
-let translate _ (_bo : BackendOptions.t) (items : AST.item list) :
-    Types.file list =
+let translate _ (_bo : BackendOptions.t) ~(bundles : AST.item list list)
+    (items : AST.item list) : Types.file list =
   U.group_items_by_namespace items
   |> Map.to_alist
   |> List.map ~f:(fun (ns, items) ->
--- a/engine/backends/coq/ssprove/ssprove_backend.ml
+++ b/engine/backends/coq/ssprove/ssprove_backend.ml
@@ -2405,8 +2405,8 @@ let hardcoded_coq_headers =
    Import choice.Choice.Exports.\n\n\
    Obligation Tactic := (* try timeout 8 *) solve_ssprove_obligations.\n"
 
-let translate _ (_bo : BackendOptions.t) (items : AST.item list) :
-    Types.file list =
+let translate _ (_bo : BackendOptions.t) ~(bundles : AST.item list list)
+    (items : AST.item list) : Types.file list =
   let analysis_data = StaticAnalysis.analyse items in
   U.group_items_by_namespace items
   |> Map.to_alist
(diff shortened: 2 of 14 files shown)


W95Psp (changes requested):
As we discussed, we should rename constructors in enumeration when their name are clashing, see https://hax-playground.cryspen.com/#fstar/0559a9e0ba/gist=02128b3051501d2c361787c887a82d26

maximebuyse:
> As we discussed, we should rename constructors in enumeration when their name are clashing, see https://hax-playground.cryspen.com/#fstar/0559a9e0ba/gist=02128b3051501d2c361787c887a82d26

As discussed, it is easier (and maybe better as we keep the naming convention) to always rename the variants when we rename the enum. We also need to include the new variants in the module of origin with an alias (to make sure the original name is available).

I implemented that. Note that the import of variants also happens when there is no renaming (not sure if it is good or not).

W95Psp (approved):
Thanks, that's great!
Let's merge it! :smiley:

PR #945, opened 2024-10-02
feat(engine/f*): allow unsafe code

W95Psp:
This PR follows from
https://github.com/hacspec/hax/issues/417#issuecomment-2387860612, and from PR #867.

It turns on the `unsafe` feature for the F* backend.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -94,9 +94,9 @@ dependencies = [
 
 [[package]]
 name = "autocfg"
-version = "1.3.0"
+version = "1.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


franziskuskiefer (changes requested):
This crashes again on use.
Please update the `Cargo.lock` and add a process that ensures that this doesn't happen.

W95Psp:
Can you copy paste the error you get?
Can you also `cargo clean` in the `tests` folder and in the main folder before running the tests?

This dependency things makes no sense to me, I don't get it.

W95Psp:
@franziskuskiefer can you try locally to run hax on `examples/chacha20`?

franziskuskiefer:
Let's just update the lock file and get this in, and open an issue to investigate this. We shouldn't block this PR on that.

PR #947, opened 2024-10-02
fix(exporter+engine): drop default parameters on traits

### LINKED ISSUE #310: Defaults traits bounds
```rust
trait Foo: Bar {
    type U;
}
trait Bar<T = <Self as Foo>::U> {}
```

W95Psp:
This PR disable default parameters on traits, both in the engine and in the exporter, following the conversation in #310.

Explanation (quoting my comment in code):

> On use site, Rust always give us all the generic
> parameters, no matter the defaultness. This information is
> thus not so useful. At the same time, as discussed in
> https://github.com/hacspec/hax/issues/310, extracting this
> default type causes failures when querying Rust for trait
> resolution. We thus decided to disable this feature. If
> this default type information is useful to you, please
> open an issue on https://github.com/hacspec/hax.

Fixes #310

### DIFF
--- a/engine/backends/coq/coq/coq_backend.ml
+++ b/engine/backends/coq/coq/coq_backend.ml
@@ -362,10 +362,8 @@ struct
 
   let pgeneric_param_as_argument span : generic_param -> C.AST.argument =
     function
-    | { ident; kind = GPType { default }; _ } ->
-        C.AST.Explicit
-          ( C.AST.Ident ident.name,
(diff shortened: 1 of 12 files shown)


Nadrieril (approved):
The changes in `copied.rs` look good to me; can't vouch for the rest.

W95Psp:
Thanks, yeah, no worries, the rest is just dumb propagation;:)

PR #955, opened 2024-10-02
feat(frontend): collect regular comments

### LINKED ISSUE #951: Recover comments whenever this is possible
@Nadrieril told me about https://github.com/AeneasVerif/charon/pull/359 and https://github.com/AeneasVerif/charon/pull/358: [Charon](https://github.com/AeneasVerif/charon/pull/358) is recovering comments inside functions via spans!

That's great, I would love to have that as well in hax!

If we could output a list of spanned comments, with the generic printer and the source spans, it should be fairly easy to insert comments at the right place, in a backend-agnostic way!

W95Psp:
This commit adds support for regular comments in hax. Comments are dumped with the command `cargo hax json -E -o -`.

To parse comments, this commit uses `rustc_lexer`, so the parsing is very dumb yet very reliable.

Fixes #951

### DIFF
--- a/cli/driver/src/exporter.rs
+++ b/cli/driver/src/exporter.rs
@@ -264,12 +264,21 @@ impl Callbacks for ExtractionCallbacks {
                 <Body>|| {
                     let (spans, def_ids, impl_infos, items) =
                         convert_thir(&self.clone().into(), self.macro_calls.clone(), tcx);
+                    let files: HashSet<PathBuf> = HashSet::from_iter(
+                        items
+                            .iter()
+                            .flat_map(|item| item.span.filename.to_path().map(|path| path.to_path_buf()))
+                    );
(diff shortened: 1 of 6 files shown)


franziskuskiefer (approved):
Thanks! We should add this to F* and test it at some point to know that it's working well. But this looks like it's doing the right thing.

W95Psp:
Yeah, I want to directly make support for the generic printer for this.
The goal would be to have comments without having to do anything special in the backend, just specify how comments are rendered and how to format them.

Nadrieril:
Looks nice, I didn't think of using `rustc_lexer`!

W95Psp:
:smiley: yeah, I was recalling our discussion about rustfmt maybe keeping comments, I dug a bit there, and found out rustc_lexer does what we want!

PR #957, opened 2024-10-03
Ignore structs in variants renaming for bundling.

maximebuyse:
Fixes a regression introduced by #935  (see https://github.com/cryspen/libcrux/actions/runs/11153943502/job/31002385466).

That was caused by trying to rename a variant of a struct, but only enum variants should be renamed.


W95Psp (approved):
LGTM :shipit:

PR #959, opened 2024-10-03
Prefer `try_normalize_erasing_regions`

Nadrieril:
There are some contexts where normalization can fail (particularly type aliases), which causes a compiler panic.


W95Psp (approved):
Thanks! That seems like a good thing to do generally, indeed.
Let's merge it!

PR #970, opened 2024-10-07
Rework trait resolution to support GATs

Nadrieril:
This reworks trait resolution as discussed offline: we only explore item bounds insofar as the `Self` type of the searched predicate is itself an associated type. This makes the search bounded, and makes it possible to support GATs in trait resolution.


W95Psp (approved):
Thanks! That looks pretty nice to me.
I really like the redesign with `PredicateSearcher`, it's super clean and nice! :smiley:

Let's merge it!

PR #977, opened 2024-10-08
More improvements to trait resolution

Nadrieril:
This fixes a few more things about trait resolution. In particular, I added a step that removes all regions so we can finally use normal equality between predicates instead of comparing strings. This should hopefully fix most binder issues, including https://github.com/hacspec/hax/issues/974.


Nadrieril:
Charon breakage is because this fixes come bugs which is great

W95Psp (approved):
That PR looks great to me, thank you! :tada:
You did a great job there, that's great :)

Let's merge that, @maximebuyse told me at the moment this is indeed fixing the bug!

PR #978, opened 2024-10-08
experiments with using refined integers to model machine integers.

karthikbhargavan:


W95Psp (commented):
That seems good to me, but what is the status of this PR?

PR #981, opened 2024-10-09
Make options extensible

W95Psp:
The CLI is defined with the rust library clap. This PR introduce a
trait `Extension` which define an associated type per type of
`cli_options`.

For now, not all the types have been added to that trait: let's do
that lazily.

With this PR, it is easy to define a superset of hax' CLI.


franziskuskiefer (approved):
Thanks

Issue #962, opened 2024-10-07
Engine: move Coq backend to the generic printer


W95Psp:
Similar to https://github.com/hacspec/hax/issues/942

PR #987, opened 2024-10-10
Coq generic printer

cmester0:
Transition coq to using the Generic printer.


cmester0:
WIP for #962

franziskuskiefer:
@cmester0 what does it take to get this in a reviewable state? The plan was to get this done this week.

W95Psp:
Status: it's basically done, actions items:
 - [ ] separate the bits for core (e.g. usize -> t_usize)
 - [ ] tests: review snapshot changes for coq
 - [ ] tests: enable coq for more tests

cmester0:
This should now be a stable version of the Coq generic printer. There is still some naming issues, which I have fixed directly in Concrete ident and the generic printer. We should figure out a better solution for this.

engine/lib/generic_printer/generic_printer.ml
```diff
@@ -435,37 +435,44 @@ module Make (F : Features.T) = struct
...
+                ^^ (self#_do_not_override_lazy_of_global_ident
+                      Generated_generic_printer_base
+                      .AstPos_pat'_PConstruct_constructor constructor)
+                     #p ^^ fields_or_empty space
+              else
+                (self#_do_not_override_lazy_of_global_ident
+                   Generated_generic_printer_base
+                   .AstPos_pat'_PConstruct_constructor constructor)
+                  #p ^^ fields_or_empty space
+            else string "(* TODO: Record? *)"
```
W95Psp, line 475: This is Coq-specific, right?
Do you need any specific logic more than tuple/record/struct here?
If yes, what do you need exactly?

engine/lib/generic_printer/generic_printer.ml
```diff
@@ -477,51 +484,49 @@ module Make (F : Features.T) = struct
             self#expr'_GlobalVar_concrete ~super concrete
         | `Primitive primitive ->
             self#expr'_GlobalVar_primitive ~super primitive
+        | `TupleCons 0 -> string "tt"
```
W95Psp, line 487: Same, this is Coq specific.
When do we call `_do_not_override_expr'_GlobalVar` or a `TupleCons _`?
We should fix that!

W95Psp: Ah, it's the constructor for `()` that ends up as a simple `GlobalVar ()`?
If so, we should call `_do_not_override_expr'_Construct` here!

engine/lib/generic_printer/generic_printer.ml
```diff
@@ -477,51 +484,49 @@ module Make (F : Features.T) = struct
...
-                  self#_do_not_override_lazy_of_pat AstPos_field_pat__pat
-                    field#v.pat)
-                fields
-            in
-            self#pat'_PConstruct_tuple ~super ~components
-        | `Primitive _ | `TupleType _ | `TupleField _ | `Projector _ ->
-            self#assertion_failure "Construct unexpected constructors"
+        match (constructor, fields) with
+        | `TupleCons 0, [] -> string "'tt (* const unit *)" (* Unit pat *)
+        | `TupleCons 1, [ p ] -> p#p
```
W95Psp, line 497: Same here, the Rust expression `(1u8,)` (a tuple of size 1) is different from `(1u8)` (a raw `u8`)
Also, unit should not be specific at the level of this file, that's a detail of implementation that should be decided by the backend, in `pat'_PConstruct_tuple`.
I don't understand what is wrong with the `_do_not_override_pat'_PConstruct` we have in main currently

engine/lib/generic_printer/generic_printer.ml
```diff
@@ -594,6 +600,37 @@ module Make (F : Features.T) = struct
...
+            (*     { *)
+            (*       kind = UnsupportedTupleSize { tuple_size = Int64.of_int n; reason }; *)
+            (*       span; *)
+            (*     } *)
+            (* | `TupleField _ | `Projector _ -> *)
+            (*   Error.assertion_failure span *)
+            (*     ("pglobal_ident: expected to be handled somewhere else: " *)
+            (*      ^ show_global_ident id) *)
+            | _ -> string "(* TODO *) error global ident type?")
+          ast_position id
```
W95Psp, line 632: `_do_not_override_lazy_of_global_ident` was undefined?

engine/backends/coq/coq/coq_backend.ml
```diff
@@ -12,6 +12,14 @@ include
...
       include On.Macro
       include On.Construct_base
+      (* include On.Trait_item_default *)
+
+      (* include On.Loop *)
+      (* include On.For_loop *)
+      (* include On.While_loop *)
+      (* include On.For_index_loop *)
+      (* include On.State_passing_loop *)
+      (* include On.Fold_like_loop *)
```
W95Psp, line 22: Clean up comments?

engine/backends/coq/coq/coq_backend.ml
```diff
@@ -88,598 +104,2219 @@ module Context = struct
...
+  let default_string_for s = "TODO: please implement the method `" ^ s ^ "`"
+  let default_document_for = default_string_for >> string
 
-    let __TODO_pat__ span s =
-      Error.unimplemented ~details:("[pat] node " ^ s) span
+  let is_document_empty doc : bool =
+    (* Replace the following with: [PPrint.is_empty], however this currently does not work, bug? *)
+    let buf = Buffer.create 0 in
+    PPrint.ToBuffer.pretty 1.0 80 buf doc;
+    String.equal "" (Buffer.contents buf)
```
W95Psp, line 135: As we discussed the other day, that will break stuff.
Rendering a document alters the state of the printer, we should *never* do that manually.

Whenever we render, we also write in some internal state some span information; this will break it.

engine/backends/coq/coq/coq_backend.ml
```diff
@@ -88,598 +104,2219 @@ module Context = struct
...
+            @
+            if is_document_empty requires then []
+            else [ string "`" ^^ braces requires ])
+            typ#p body#p (* ^^ TODO: ensures? *)
+        else
+          CoqNotation.definition name#p generics#p
+            (List.map ~f:(fun x -> x#p) params
+            @
+            if is_document_empty requires then []
+            else [ string "`" ^^ braces requires ])
```
W95Psp, line 457: With requires being a option, that will be easy to replace those `is_document_empty` with a match

engine/backends/coq/coq/coq_backend.ml
```diff
@@ -88,598 +104,2219 @@ module Context = struct
...
-    make { current_namespace = U.Concrete_ident_view.to_namespace item.ident }
-  in
-  List.map ~f:C.decl_to_string @@ Print.pitem item |> String.concat ~sep:"\n"
+    method _do_not_override_expr'_App :
+      super:expr ->
+      f:expr Generic_printer.LazyDoc.lazy_doc ->
+      args:expr Generic_printer.LazyDoc.lazy_doc list ->
+      generic_args:generic_value Generic_printer.LazyDoc.lazy_doc list ->
+      bounds_impls:impl_expr Generic_printer.LazyDoc.lazy_doc list ->
+      trait:
```
W95Psp, line 834: I can take a look at this, we really don't want to inline that

engine/lib/concrete_ident/concrete_ident.ml
```diff
@@ -346,7 +346,10 @@ module View = struct
         path
     in
     let sep = "__" in
-    let subst = String.substr_replace_all ~pattern:sep ~with_:(sep ^ "_") in
+    let subst =
```
W95Psp, line 349: Are those changes for the new core library?

cmester0: This is just to get consistent naming in Coq, since I cannot have the same name twice. However, I still have some cases, where a constructor is refered to as `<name>` instead of `t_<name>`.

W95Psp: I don't get it, can you provide a small example and the translation we want?

engine/lib/generic_printer/generic_printer.ml
cmester0: I could not find a definition, but it seems most places it was only `Concrete` that was used, and then call directly to `_do_not_override_lazy_of_concrete_ident`

engine/lib/generic_printer/generic_printer.ml
cmester0: This should now be reverted, and moved into Coq.

engine/lib/generic_printer/generic_printer.ml
cmester0: I think my issue might have been that `field` was evaluated, now that it is lazy, the errors I had, no longer happens, since I never call `#p`.

cmester0:
I could not figure out how to get the sourcemap, however, if possible we should also return the sourcemap.

engine/lib/concrete_ident/concrete_ident.ml
cmester0: The naming is not consistent with how it is used in Coq backend, furthermore I get collisions of names. This disambiguates the names and makes it (more) consistent across different uses. I could move this into the Coq backend, but then the naming logic would be in many different functions.

engine/lib/generic_printer/generic_printer.ml
cmester0: I should now call the correct function, instead of doing this hack. This was mostly done for debugging.

engine/lib/generic_printer/generic_printer.ml
cmester0: Yes, it was a problem in some of the snapshots. I changed it to use `_do_not_override_expr'_Construct`

engine/lib/concrete_ident/concrete_ident.ml
cmester0: The line `>> String.substr_replace_all ~pattern:"___" ~with_:""` was to fix issues with `___self` occuring in require statements. However, this was a fix for annotated core lib. The rest is issues with naming for the coq backend.

engine/lib/generic_printer/generic_printer.ml
```diff
@@ -594,6 +591,17 @@ module Make (F : Features.T) = struct
...
 
+      method _do_not_override_lazy_of_global_ident ast_position
+          (id : global_ident) : global_ident lazy_doc =
+        lazy_doc
+          (fun (id : global_ident) ->
+            match id with
+            | `Concrete cid ->
+                (self#_do_not_override_lazy_of_concrete_ident ast_position cid)
+                  #p
+            | _ -> self#assertion_failure "[global_ident]")
```
W95Psp, line 602:
```suggestion
            | _ -> self#assertion_failure ("_do_not_override_lazy_of_global_ident: expected [`Concrete _] got [" ^ [%show: global_ident] id ^ "]")
```

engine/backends/coq/coq/coq_backend.ml
```diff
@@ -71,615 +71,791 @@ module CoqNamePolicy = Concrete_ident.DefaultNamePolicy
...
 
-module C = Coq (CoqLibrary)
+  type ('get_span_data, 'a) object_type =
+    ('get_span_data, 'a) BasePrinter.Gen.object_type
 
-module Context = struct
-  type t = { current_namespace : string * string list }
-end
+  class printer =
+    (* let associated_expr = let open m in associated_expr in *)
```
W95Psp, line 127: Stale comment?

engine/backends/coq/coq/coq_backend.ml
```diff
@@ -71,615 +71,791 @@ module CoqNamePolicy = Concrete_ident.DefaultNamePolicy
...
-        "TODO (global ident) tuple field"
-    | _ -> .
-
-  module TODOs_debug = struct
-    let __TODO_pat__ _ s = C.AST.Ident (s ^ " todo(pat)")
-    let __TODO_ty__ _ s : C.AST.ty = C.AST.NameTy (s ^ " todo(ty)")
-    let __TODO_item__ _ s = C.AST.Unimplemented (s ^ " todo(item)")
-    let __TODO_term__ _ s = C.AST.Const (C.AST.Const_string (s ^ " todo(term)"))
-  end
+      (* BEGIN GENERATED *)
```
W95Psp, line 138:
```suggestion
```

engine/backends/coq/coq/coq_backend.ml
```diff
@@ -71,615 +71,791 @@ module CoqNamePolicy = Concrete_ident.DefaultNamePolicy
...
+      (*            ^^ space) *)
+      (*          base) *)
+      (*   else *)
+      (*     (\* Tuple struct *\) *)
+      (*     string "Build_" ^^ constructor#p *)
+      (*     ^^ concat_map (fun (ident, exp) -> space ^^ parens exp#p) fields *)
+      (* else *)
+      (*   (\* Indutive type *\) *)
+      (*   constructor#p *)
+      (*   ^^ concat_map (fun (ident, exp) -> space ^^ parens exp#p) fields *)
```
W95Psp, line 277: Stale comments?

engine/backends/coq/coq/coq_backend.ml
```diff
@@ -71,615 +71,791 @@ module CoqNamePolicy = Concrete_ident.DefaultNamePolicy
...
+        (*   | S64 -> ("u64", "U64") *)
+        (*   | S128 -> ("u128", "U128") *)
+        (*   | SSize -> ("usize", "U64") *)
+        (*   (\* Dependens on architecture.. *\) *)
+        (* in *)
+        (* string ("Build_t_" ^ outer) *)
+        (* ^^ space *)
+        (* ^^ parens *)
+        (*      (string ("Build_t_" ^ inner) *)
+        (*      ^^ space ^^ string value ^^ string "%N") *)
```
W95Psp, line 620: Stale comment?

engine/backends/coq/coq/coq_backend.ml
```diff
@@ -71,615 +71,791 @@ module CoqNamePolicy = Concrete_ident.DefaultNamePolicy
...
+      method pat'_PConstruct_tuple ~super:_ ~components =
+        (* TODO: Only add `'` if you are a top-level pattern *)
+        (* string "'" ^^ *)
+        parens (separate_map comma (fun x -> x#p) components)
+
+      method pat'_PDeref ~super:_ ~subpat:_ ~witness:_ =
+        default_document_for "pat'_PDeref"
+
+      (* method pat'_POr ~super ~subpats = *)
+      (*   parens( subpats ) *)
```
W95Psp, line 681: Stale?

engine/lib/concrete_ident/concrete_ident.ml
W95Psp: Mmmh. That's not clear to me.
Let's pull those changes out of this PR, make an issue describing the problem with an example of a collision, and then push those changes to a separate PR.
That will help I think.

W95Psp (commented):
Thanks, I have a few comments, but not big things.
The only thing is about concrete_ident: let's pull the changes out of this PR to a separate PR, it is going to be nicer to review, currently things are not clear to me.

W95Psp (approved):
That's great, thank you! Let's merge that :D

PR #988, opened 2024-10-10
feat(engine) Rewrite control flow inside loops.

### LINKED ISSUE #393: Add a `RewriteControlFlow` phase (`return`/`continue`/`break`)

maximebuyse:
Closes #393.

Loops are already translated as folds. This PR wraps their accumulators to add control flow information (in case of `break` `return` or `continue`). Some specific fold operators are also added to handle these wrapped accumulators.

This PR puts `LocalMutation` before `RewriteControlFlow`. In `RewriteControlFlow`, we now deal with the body of loops (to push control flow in return position, like this phase already does for `return`s outside loops). We also wrap loops that have a `return` inside a pattern matching over the result (to propagate a returned value), and apply the phase to inline what comes after in the branches of the pattern matching. Phase `DropNeedlessReturns` is now `DropREturnBreakContinue`. Inside loops it replaces `return` , `break` and `continue` by their encoding in the `ControlFlow` enum. It treats a value in return position of a loop body like a `continue`. Phase `FunctionalizeLoops` selects the correct `fold` operator (each fold should now come in 3 versions: the normal one, the one supporting `break` and `continue`, and the one supporting all type of control flow).


maximebuyse:
Code generation should be almost stable. The remaining part is to write the specific folds as part of the fstar `Rust_primitives`

maximebuyse:
@W95Psp still testing but you can start to review.

engine/lib/phases/phase_functionalize_loops.ml
```diff
@@ -38,6 +43,18 @@ struct
...
+        inherit [_] Visitors.reduce as super
+        method zero = false
+        method plus = ( || )
+
+        method! visit_expr' () e =
+          match e with
+          | Return _ | Break _ | Continue _ -> true
+          | _ -> super#visit_expr' () e
+      end
+
```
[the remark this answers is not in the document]
W95Psp: It appears to be unused, some how OCaml doesn't complains :thinking:

engine/lib/ast.ml
```diff
@@ -307,13 +313,23 @@ functor
 
     and loop_kind =
       | UnconditionalLoop
-      | WhileLoop of { condition : expr; witness : F.while_loop }
-      | ForLoop of { pat : pat; it : expr; witness : F.for_loop }
+      | WhileLoop of {
+          condition : expr;
+          has_return : bool;
```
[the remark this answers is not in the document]
W95Psp: I think at some point we discussed a `fold_like_loops`, replacing the `state_passing_loop` feature?

engine/lib/phases/phase_functionalize_loops.ml
```diff
@@ -129,29 +140,45 @@ struct
         Some (ChunksExact { size; slice })
       else None
 
-    let fn_args_of_iterator (it : iterator) :
-        (Concrete_ident.name * B.expr list * B.ty) option =
+    let fn_args_of_iterator (has_cf : bool) (has_return : bool) (it : iterator)
```
W95Psp, line 143: I feel like those boolean flags are not very nice, we only have three states: no control flow, only returns, both returns and breaks. But if we want to go from those two bools to this enum, we need to lift that to the AST or something

engine/lib/phases/phase_reject.ml
```diff
@@ -41,6 +41,45 @@ end
...
+        module B = FB
+        include Feature_gate.DefaultSubtype
+
+        let question_mark = reject
+        let metadata = make_metadata QuestionMark
+      end)
+end
+
+module _ (FA : Features.T) : Phase_utils.PHASE = Continue (FA)
+module _ (FA : Features.T) : Phase_utils.PHASE = Break (FA)
```
W95Psp, line 81: Ideally those should be ensured statically by the phase removing breaks and question marks

engine/lib/phases/phase_rewrite_control_flow.ml
```diff
@@ -11,31 +14,127 @@ module Make (F : Features.T) =
...
           method zero = false
           method plus = ( || )
 
           method! visit_expr' () e =
-            match e with Return _ -> true | _ -> super#visit_expr' () e
+            match e with
+            | Return _ | Break _ | Continue _ -> true
+            | Loop _ -> false
+            | _ -> super#visit_expr' () e
         end
```
W95Psp, line 60: What is the difference between this and `has_return` with the `in_loop` set to `true`

engine/lib/phases/phase_rewrite_control_flow.ml
```diff
@@ -11,31 +14,127 @@ module Make (F : Features.T) =
...
+          inherit [_] Visitors.reduce as super
+          method zero = false
+          method plus = ( || )
+
+          method! visit_expr' in_loop e =
+            match e with
+            | Return _ -> true
+            | (Break _ | Continue _) when in_loop -> true
+            | _ -> super#visit_expr' in_loop e
+        end
```
W95Psp, line 35: Why is that OK to return true if we face e.g. `loop { loop { return 3 }}`? I mean, shouldn't `has_return` return true only if there is a return in the current loop/function?
If it is OK, can you explain with a comment?

W95Psp (changes requested):
Thanks! I have a first set of comments!

engine/lib/phases/phase_drop_return_break_continue.ml
```diff
@@ -0,0 +1,134 @@
+(* This phase removes `return`s in exit position. Inside loops,
```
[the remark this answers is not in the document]
W95Psp: We chatted about this: we should make this phase typed, but since the visitor by design doesn't need to visit every nodes, rejecting within the rewrite is difficult.
Let's then just add a treatment after the visitor that takes care of rejecting (with an assertion failure) any return/break/continue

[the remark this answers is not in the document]
W95Psp: Also, we want `Loop` in AST to carry a ``([ `BreakOnly | `BreakOrReturn ] * F.fold_like_loop) option``, and we want this phase to enable `fold_like_loop`.

engine/lib/phases/phase_drop_return_break_continue.ml
```diff
@@ -0,0 +1,134 @@
...
+
+      open Ast.Make (F)
+      module U = Ast_utils.Make (F)
+      module Visitors = Ast_visitors.Make (F)
+
+      module Error = Phase_utils.MakeError (struct
+        let ctx = Diagnostics.Context.Phase phase_id
+      end)
+
+      type loop_info = { return_type : ty option; break_type : ty }
```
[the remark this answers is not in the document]
maximebuyse: We rewrite the loops during `LocalMutation` which means the type of the loop is the type of the accumulator

engine/lib/phases/phase_rewrite_control_flow.ml
maximebuyse: Even nested `return`s apply to the top level loop but `break`s and `continue`s don't. I'll add a comment.

maximebuyse:
@W95Psp I think I fixed everything according to your feedback!

engine/lib/ast.ml
```diff
@@ -217,6 +217,7 @@ functor
 
     and pat = { p : pat'; span : span; typ : ty }
     and field_pat = { field : global_ident; pat : pat }
+    and cf_kind = BreakOnly | BreakOrReturn
```
W95Psp, line 220: Can you add a comment here to explain what it is, where it is used and why (e.g. we need that to track what control flow was present in the loop after removing that control flow in favor of simple values representing this control flow)?

engine/lib/ast_builder.ml
```diff
@@ -22,17 +22,173 @@ module Make (F : Features.T) = struct
     let ty_tuple_or_id : ty list -> ty = function
       | [ ty ] -> ty
       | types -> ty_tuple types
+
+    let ty_control_flow ~(continue_type : ty) ~(break_type : ty) : ty =
```
W95Psp, line 26: Can you rename to `ty_cf` for consistency?

W95Psp: (+ add a comment)

engine/lib/ast_builder.ml
```diff
@@ -22,17 +22,173 @@ module Make (F : Features.T) = struct
...
+      TApp
+        {
+          ident = Global_ident.of_name Type Core__ops__control_flow__ControlFlow;
+          args = [ GType break_type; GType continue_type ];
+        }
+
+    let ty_cf_return ~(acc_type : ty) ~(break_type : ty)
```
W95Psp, line 33: We need a comment that explains we encode `return` as `Break(Break(...))` and `break` as `Break(Continue)`, to justify the existence of this funciton

engine/lib/ast_builder.ml
```diff
@@ -22,17 +22,173 @@ module Make (F : Features.T) = struct
...
+      }
+
+    let call_Constructor (constructor_name : Concrete_ident.name)
+        (is_struct : bool) (args : expr list) span ret_typ =
+      call_Constructor'
+        (`Concrete
+          (Concrete_ident.of_name (Constructor { is_struct }) constructor_name))
+        is_struct args span ret_typ
+
+    let expr'_Constructor_CF ~(span : span) ~(break_type : ty)
```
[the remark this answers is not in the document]
W95Psp: Ah, I see that's basically what you have below, nethermind then! :)

engine/lib/phases/phase_drop_return_break_continue.ml
```diff
@@ -0,0 +1,164 @@
+(* This phase removes `return`s in exit position. Inside loops,
```
W95Psp, line 1: Should be a doc comment (** *)

engine/lib/ast_builder.ml
```diff
@@ -22,17 +22,173 @@ module Make (F : Features.T) = struct
...
+    (* We use the following encoding of return, break and continue in the `ControlFlow` enum:
+       Return e -> Break (Break e)
+       Break e -> Break ((Continue(e, acc)))
+       Continue -> Continue(acc)
+
+       In case there is no return we simplify to:
+        Break e -> (Break (e, acc))
+        Continue -> (continue (acc))
+    *)
+    let expr_Constructor_CF ~(span : span) ~(break_type : ty option)
```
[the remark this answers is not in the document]
W95Psp: +  I'd prefer a full path to fully qualified std::ops::ControlFlow<break, continue>

engine/lib/phases/phase_drop_return_break_continue.ml
```diff
@@ -0,0 +1,164 @@
...
+                 in exit nodes. **)
+      end
+
+    [%%inline_defs dmutability + dsafety_kind]
+
+    let rec dexpr' (span : span) (expr : A.expr') : B.expr' =
+      match expr with
+      | [%inline_arms "dexpr'.*" - Return - Break - Continue - Loop] -> auto
+      | Return _ | Break _ | Continue _ ->
+          Error.raise { kind = NonTrivialAndMutFnInput (*TODO Correct*); span }
```
W95Psp, line 127: This TODO is about changing the error?
Here you should use `Error.assertion_failure span "Return/Break/Continue are expected to be gone as this point"` or something like that

engine/lib/phases/phase_rewrite_control_flow.ml
```diff
@@ -11,31 +14,97 @@ module Make (F : Features.T) =
...
+                    (U.make_lets stmts_after final |> self#visit_expr in_loop);
+                ]
+              in
+              MS.expr_Match ~scrutinee:loop ~arms ~typ:return_type
+            in
             match e.e with
-            | _ when not (has_return#visit_expr () e) -> e
-            (* Returns in loops will be handled by issue #196 *)
-            | Loop _ -> e
+            (* This is supposed to improve performance but it might actually make it worse in some cases *)
```
W95Psp, line 85: Can you file an issue and reference it there so that we track that?

maximebuyse: https://github.com/hacspec/hax/issues/1033

W95Psp (approved):

Issue #747, opened 2024-07-01
Panic: type parameter out of range when instantiating


Nadrieril:
Minimized reproduction:
```rust
pub trait Trait {}

impl<T, U> Trait for Result<T, U>
where
    for<'a> &'a Result<T, U>: IntoIterator,
    for<'a> <&'a Result<T, U> as IntoIterator>::Item: Copy,
{
}
```
[Open this code snippet in the playground](https://hax-playground.cryspen.com/#fstar/latest-main/gist=b3d9ce3e20013336bfc60b554145151d)

panics with
```
thread 'rustc' panicked at /rustc/6b0f4b5ec3aa707ecaa78230722117324a4ce23c/compiler/rustc_type_ir/src/binder.rs:713:9:
type parameter `U/#1` (U/#1/1) out of range when instantiating, args=[&'^0.Named(DefId(0:8 ~ derive_visitor[edb9]::{impl#0}::'a#1), "'a") std::result::Result<T/#0, U/#1>]
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
error: could not compile `derive-visitor` (lib)
```

This does not panic with a similar `Option` impl, which suggests there's something weird happening with the generics. This also does not panic if I replace the HRTBs with simple `Result<T, U>: IntoIterator` bounds.

W95Psp:
I suspect that's the same issue as https://github.com/hacspec/hax/issues/683

Nadrieril:
With https://github.com/hacspec/hax/pull/906, the error becomes:
```text
warning[E9999]: Hax frontend found a projected type with escaping bound vars. Please report https://github.com/hacspec/hax/issues/495
  |
  = note: ⚠️  This is a bug in Hax's frontend.
          Please report this error to https://github.com/hacspec/hax/issues with some context (e.g. the current crate)!

[...]

PR #990, opened 2024-10-11
Remove workaround in trait resolution for associated items

Nadrieril:
Now that we thoroughly erase lifetimes in trait resolution, we shouldn't have any binder problems left.

Fixes https://github.com/hacspec/hax/issues/747


Nadrieril:
Charon failure is because I had #747 as a failing test and it now passes :D

W95Psp (approved):
Thanks, having `erase_and_norm` doing the right job really helps a lot :sweat_smile:
Let's merge! :)

Issue #754, opened 2024-07-09
Avoid translating the same thing many times


Nadrieril:
Rustc interns most of its datastructures. For example, `Ty<'tcx>` is a reference that is guaranteed to be unique for each (structurally-equal) type, and hashes instantaneously. We could save a lot of translation by keeping a `HashMap<Ty<'tcx>, Rc<hax::Ty>>` around and re-using what we've already translated.

Nadrieril:
Fixed in #996

PR #996, opened 2024-10-14
Cache many things

Nadrieril:
This PR adds caching to hax. We now cache the translation of `Span`, DefId`, ImplExpr` and `Ty`, cache the resolution of traits (`TraitRef` -> `ImplExpr`) within a given item, and deduplicate `Ty`s by adding a reference-counted indirection.

This is all to try and run `cargo-hax` on crates such as https://github.com/cryptocorrosion/cryptocorrosion/tree/master/hashes/blake which today results in an OOM crash. With this PR (particularly the hash-consing of `Ty`) we no longer OOM on `cryptocorrosion/hashes/blake`. We still don't terminate however, so there remains more to investigate.


Nadrieril:
Charon failure is expected because this changes `Ty` to `TyKind`

W95Psp:
Sounds great! Fixes #754

W95Psp (approved):
Thank you! That looks good to me!
The split of `trait.rs` is a great thing.

This should improve a lot perfs :partying_face: :tada:  

Let's merge ASAP!

Issue #995, opened 2024-10-14
Circular dependency handling is not working on libcrux ml-dsa


maximebuyse:
The implementation for #396 probably has a bug.

maximebuyse:
I found another issue. Here is a reproducer:
```rust


pub mod a {
    pub fn f () {super::b::h()}
    pub fn g () {}
}
pub mod b {
    pub fn h () {}
    pub fn i () {super::a::g()}
}



```
[Open this code snippet in the playground](https://hax-playground.cryspen.com/#fstar/1ba7ce3e4b/gist=8862d6578465f047726fce3a435f7419)

The problem is that the we look for chains of items making a module depend on itself but we don't need a chain, we can have the dependency from a to b and from b to a use completely disjoint items.

In this example we need to bundle only one item to break the cycle (`g` or `h`). 

One way to break these cycles that has an arbitrary part is to look for module level cycles (`M1 -> M2 -> ... -> Mn -> M1`). Take one module dependency that is in the cycle ( for example `Mn -> M1`). Look for all items in `M1` that are in the dependencies of `Mn`. Bundle them together with all there dependencies in any module that is part of the cycle. This is not so nice because choosing arbitrarily which part of the cycle we should break can lead to bundle many things when there can be another solution minimizing the amount of items to bundle.

A less naive solution would be to look for the best edges to break in the module dependency graph (minimum number of edges, we could even do a weighted version where an edge in the module dependency graph has a weight that is the number of edges in the item dependency graph that correspond). This is the minimal feedback arc set problem. And there is a [solution](https://backtracking.github.io/ocamlgraph/ocamlgraph/Graph/Cycles/Fashwo/index.html) for that in the ocamlgraph library.

maximebuyse:
> A less naive solution would be to look for the best edges to break in the module dependency graph (minimum number of edges, we could even do a weighted version where an edge in the module dependency graph has a weight that is the number of edges in the item dependency graph that correspond). This is the minimal feedback arc set problem. And there is a [solution](https://backtracking.github.io/ocamlgraph/ocamlgraph/Graph/Cycles/Fashwo/index.html) for that in the ocamlgraph library.

This is still an imperfect solution because the module dependency graph is missing the information of how much subsequent items will need to be bundled to break an edge.

PR #1002, opened 2024-10-14
Fix bug with recursive bundles depending on items from original modules.

maximebuyse:
Fixes a bug found while investigating #995 . More generally this fixes many problems in the bundling of cyclic dependencies


maximebuyse:
As described [here](https://github.com/hacspec/hax/issues/995#issuecomment-2413253296) the problem is more complex than I usually thought if one wants to minimize the amount of bundled items. So this PR implements a conservative approach (that is much more simple than what we did before): when there is a cycle at the module level, it bundles together in a new module everything the content of all the modules in the cycle. https://github.com/hacspec/hax/issues/1005 will track whether this is practical enough and if we should find an algorithm to minimize bundles.

W95Psp (approved):
LGTM, thanks!

W95Psp:
Looks like tests snapshot are not up to date? :thinking:

maximebuyse:
> Looks like tests snapshot are not up to date? 🤔

Yes, noticed that too. I am looking into it.

PR #1004, opened 2024-10-16
Restore neg_equiv_lemma

mamonet:
This PR restores `neg_equiv_lemma` which is needed for portable compress proofs in this PR https://github.com/cryspen/libcrux/pull/631.

### DIFF
--- a/proof-libs/fstar/rust_primitives/Rust_primitives.Integers.fsti
+++ b/proof-libs/fstar/rust_primitives/Rust_primitives.Integers.fsti
@@ -315,6 +315,10 @@ let abs_int (#t:inttype) (a:int_t t{minint t < v a}) =
 let neg (#t:inttype{signed t}) (a:int_t t{range (0 - v a) t}) =
     mk_int #t (0 - (v a))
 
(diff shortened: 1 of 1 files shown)


W95Psp (changes requested):
LGTM, but shouldn't this lemma be two lemmas? :thinking: 
Also, can you add a definition in `proof-libs/fstar/rust_primitives/Rust_primitives.Integers.fst`? (a actual proof would be cool, but if that's taking too much time, put at least an `let ... = admit ()`)

mamonet:
Definition is already in `proof-libs/fstar/rust_primitives/Rust_primitives.Integers.fst`,  some signatures were dropped while refactoring that module. I will check why it has no proof and see if it's easy to reason that lemma

W95Psp:
Ahh! Sorry I misunderstood! Sounds good to me!

W95Psp (approved):

karthikbhargavan (approved):

franziskuskiefer:
@karthikbhargavan double check if this got merged already elsewhere.

PR #1006, opened 2024-10-16
Support more features

Nadrieril:
This PR is a series of small fixes to support more rust.


.envrc
```diff
@@ -1 +0,0 @@
-use flake
```
W95Psp, line 1: Mmh, why do you remove the envrc?
I never exactly know what to think about committing `envrc`s to repos, but in the case of hax, not having an envrc will cause me pain: I use a lot git workdirs, whence I'll need to `git worktree add whatever && cd whatever && echo 'use flake' > .direnv`, which annoys me :(

W95Psp (approved):
Thanks, that looks great!
Just as a note: this adds the revealed type to `Opaque` types, but has nothing to do with trait resolution on opaque types (and thus doesn't impact https://github.com/hacspec/hax/issues/684)

Nadrieril:
The corresponding charon fix is ready

.envrc
Nadrieril: I will annoy me too tbf, because I'll need to add the envrc file to all my worktrees too, but the save in evaluation time is 1000% worth the pain

W95Psp: yeah, makes sense. I was using lorri a while ago, but had issues with it, I think it was not supporting flakes at that time :thinking:
I could try lorri again, but the issue at the end is the annoying "shall we commit envrc" situation :'(

Nadrieril: > that works across worktrees?

no idea, it just makes it possible to keep the `.envrc` committed like today while I use my own version locally.

PR #1008, opened 2024-10-16
Support `impl Trait` in trait resolution

Nadrieril:
Fixes https://github.com/hacspec/hax/issues/684 by fully revealing opaque types during trait resolution.


W95Psp (approved):
Thanks for that fix, calling `with_reveal_all_normalized` did the trick, that's good news :smiley:

PR #1017, opened 2024-10-21
Minimal example of ProVerif extraction

jschneider-bensch:
This provides a minimal example of an extraction to ProVerif, with documentation on how to run proverif on the extracted model.

I've done a tiny bit of clean-up on the PV backend, also removing the `Process` and `Toplevel` outputs. They are not actually populated at the moment and letting users remove them (especially `analysis.pv`) is unneccessary work they would have to do.


W95Psp (commented):
Sound good to me, but I guess we should wait @karthikbhargavan and you chat about the PV example itself?
I guess one thing we should do is add that to CI

jschneider-bensch:
Closing this in favor of #1023

PR #1022, opened 2024-10-22
some fixes for iterators.f_next and update_at_usize

karthikbhargavan:
The iterator model in the hax proof-libs used F* options instead of `Core.Option.t_Option`. This PR fixes that.
There is also a drive by fix remove the duplicate definition of `update_at_usize`


W95Psp (approved):
LGTM, thanks!

PR #1023, opened 2024-10-22
PV backend example: Simple PSK based protocol

jschneider-bensch:
There are some updates to the PV backend to make this example work:
- Handle `Quote`d types and functions
- Handle `opaque_type` annotations

Additionally, we remove the (empty) output `analysis.pv` and the empty subprinters `Processes` and `Toplevel`.


W95Psp (approved):
Looks good to me! :tada:

PR #1024, opened 2024-10-22
Reorganize `frontend/exporter/types`

Nadrieril:
This reorganizes the big bag of copied types into thematic submodules. I also removed a number of unused definitions.


W95Psp (approved):
Thanks, that looks great!
Seeing this huge `copied.rs` disappear makes me happy :smile:

I have no objection, let's merge it!

PR #1029, opened 2024-10-23
feat: deduplicate types in haxmeta and JSON

### LINKED ISSUE #863: Frontend Performance Problem
# Issue
We discovered a performance issue in the frontend when using the `tock_registers` crate.
[https://docs.rs/tock-registers/latest/tock_registers/](https://docs.rs/tock-registers/latest/tock_registers/)

# Details
There is a macro called `register_bitfields` which makes it easy to defined type-checked memory mapped registers with bitfields. And it can be used like so:
```rust
tock_registers::register_bitfields![u64,
    REG [
        FIELD1 8,
        FIELD2 8,
        FIELD3 8,
        FIELD4 8,
        FIELD5 8,
        FIELD6 8,
        FIELD7 8,
        FIELD8 8,
    ],
];
```

We observed an exponential blowup in the size of the JSON exported by the frontend w.r.t. to the number of bitfields in a register. The example here generates a JSON that takes 2GB on disk. However, `rustc` has no problem compiling the code above and only takes a fraction of a second. The extracted code from the backend also looks normal as expected, where the final output only grows linearly w.r.t. to the number of bitfields.

In a realistic project that uses this macro, hax simply gets killed by the OS due to OOM after using over 80GB of memory.

# Potential Causes

- Lack of interning: Neither the frontend nor the backend employs any form of interning to curb memory use.
- JSON serialization: JSON is an inherently inefficient serialization format.

There might also be issues with how macros are expanded that I'm not aware of.

# Solution
Unfortunately there doesn't seem to be an easy solution. Solving issues like this and similar scaling problems would likely involve major refactoring across the hax codebase.

### LINKED ISSUE #875: Stack overflow for deep expressions in bincode serialization/deserialization

### LINKED ISSUE #1001: Frontend: haxmeta with bincode is very slow because of duplication
We output haxmeta files, and those contain a lot of duplicate information.
We should get rid of this redundancy, similarly to what #996 does.

For https://github.com/hacspec/hax/issues/686, the haxmeta then JSON encoding takes around 4 minutes... This is huge.

W95Psp:
This PR introduces a module `id_table` in `hax-frontend-exporter`: a table mapping identifiers to various types, currently only to `Ty`.
A type is now not only a `TyKind` (that disappears during serialization if requested), but also an `Id`. `Ty` is defined as `id_table::Node<Ty>`.
Together with types provided by `id_table`, this enables a stateful serialization and deserialization that deduplicates types.

It is easy to extend the table to support more types.
This PR also adds support for tables in the engine.

This PR also drops bincode (fixes #875).

Fixes #1001, fixes #863

Note: I will make a PR to Charon, it needs a small patch.


frontend/exporter/src/state.rs
```diff
@@ -319,16 +321,28 @@ pub trait WithGlobalCacheExt<'tcx>: BaseState<'tcx> {
     }
     /// Access the cache for a given item. You must not call `sinto` within this function as this
     /// will likely result in `BorrowMut` panics.
-    fn with_item_cache<T>(&self, def_id: RDefId, f: impl FnOnce(&mut ItemCache<'tcx>) -> T) -> T {
-        self.with_global_cache(|cache| f(cache.per_item.entry(def_id).or_default()))
+    fn with_item_cache<T>(
+        &self,
+        def_id: RDefId,
+        f: impl FnOnce(&mut ItemCache<'tcx>, &mut id_table::Session) -> T,
```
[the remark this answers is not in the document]
W95Psp: Sure, makes sense!

frontend/exporter/src/types/copied.rs
```diff
@@ -1832,29 +1788,20 @@ impl<'tcx, S: UnderOwnerState<'tcx>> SInto<S, Box<Ty>> for rustc_middle::ty::Ty<
...
-pub struct Ty {
-    pub kind: Arc<TyKind>,
-}
-
-impl Ty {
-    pub fn kind(&self) -> &TyKind {
-        self.kind.as_ref()
-    }
-}
+pub type Ty = id_table::Node<TyKind>;
```
Nadrieril, line 1791: Just for the sake of encapsulation, I would prefer something like:

```
#[derive_group(Serializers)]
#[derive(Clone, Debug, JsonSchema, Hash, PartialEq, Eq, PartialOrd, Ord)]
pub struct Ty {
    pub(crate) kind: id_table::Node<TyKind>,
}

impl Ty {
    pub fn inner(&self) -> &Arc<TyKind> {
        self.kind.inner()
    }

    pub fn kind(&self) -> &TyKind {
        self.inner().as_ref()
    }
}
```

frontend/exporter/src/id_table.rs
```diff
@@ -0,0 +1,375 @@
...
+        let id = self.next_id.id;
+        self.next_id.id += 1;
+        Id { id }
+    }
+}
+
+impl<T: Sync + Send + Clone + 'static + SupportedType> Node<T> {
+    pub fn new(value: T, session: &mut Session) -> Self {
+        let id = session.fresh_id();
+        let kind = Arc::new(value);
```
Nadrieril, line 132: why not use `Arc::as_ptr(&kind).addr()` as unique identifier instead of making and storing a fresh one?

frontend/exporter/src/id_table.rs
```diff
@@ -0,0 +1,375 @@
...
+impl Session {
+    pub fn table(&self) -> &Table {
+        &self.table
+    }
+}
+
+/// The different types of values one can store in an ID table.
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub enum Value {
+    Ty(Arc<TyKind>),
```
Nadrieril, line 49: This will keep every type ever translated in memory. I suggest using `Weak` ref-counted pointers instead, so that when all the "real uses" of a given type go out of scope, the type can be deallocated.
```suggestion
    Ty(Weak<TyKind>),
```

frontend/exporter/src/id_table.rs
```diff
@@ -0,0 +1,375 @@
...
+
+impl Session {
+    pub fn table(&self) -> &Table {
+        &self.table
+    }
+}
+
+/// The different types of values one can store in an ID table.
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub enum Value {
```
Nadrieril, line 48: Regarding the enum, I was planning to copy my [`type_map`](https://github.com/AeneasVerif/charon/blob/b38710c4655e9eaf91beac02c15030e9cbf87999/charon/src/common.rs#L58) module to hax to make it possible to cache `FullDef<Body>`. We could use it here, it would nicely separate the "caching" part from the "type-generic" part.

frontend/exporter/src/id_table.rs
```diff
@@ -0,0 +1,375 @@
...
+}
+
+/// Helper function that makes sure no nested deserializations occur.
+fn full_id_deserialization<T>(f: impl FnOnce() -> T) -> T {
+    let _lock = *DESERIALIZATION_STATE_LOCK.lock().expect("CACHE_MAP_LOCK: only one WithTable deserialization can occur at a time (nesting is forbidden)");
+    let result = f();
+    result
+}
+
+/// The deserializer of `WithTable<T>` is special. First, we need to continuously
```
Nadrieril, line 195: This comment is cut off

frontend/exporter/src/id_table.rs
W95Psp: Yeah, I was thinking about that also, but that required more serde custom serializer and deserializer, so I decided an enum would do the trick for now.
But definitively, that could be a nice improvement :)

frontend/exporter/src/id_table.rs
```diff
@@ -0,0 +1,375 @@
...
+    }
+    pub fn destruct(self) -> (T, Table) {
+        let Self { value, table: map } = self;
+        (value, map)
+    }
+}
+
+/// Helper function that makes sure no nested deserializations occur.
+fn full_id_deserialization<T>(f: impl FnOnce() -> T) -> T {
+    let _lock = *DESERIALIZATION_STATE_LOCK.lock().expect("CACHE_MAP_LOCK: only one WithTable deserialization can occur at a time (nesting is forbidden)");
```
Nadrieril, line 190: Should you dereference here? Doesn't that copy the `()` and drop the guard? I might be missing a level

W95Psp: Uuh, thanks! was doing something else with that before, now that makes it useless!
I will drop that and add a type annotation to make it clear

Nadrieril: Also, `lock` blocks waiting for someone else holding the lock, so this will deadlock in case of nested deserializations. You want `try_lock` instead. (the `Result` that you call `expect` on is for lock poisoning)

W95Psp: ah yes right, thank you, that's a good catch, this would have been pretty horrible to debug!

frontend/exporter/src/id_table.rs
Nadrieril: Actually your thing is neat, it just took me a moment to follow what bit did what. I would find it clearer if there was a separate module for the type-generic map and if the table was a wrapper around it

Nadrieril:
I would really appreciate if you merged https://github.com/hacspec/hax/pull/1024 first to make conflicts easier to fix

frontend/exporter/src/id_table.rs
W95Psp: Mmh, sure, but here that's the state that holds those `Arc`: we allocate a state only during the translation, then we can drop it, and that would drop the table as well, dropping the types held in the table.
Concerning the state of the deserializer, I take care of emptying things after deserialization.

Also, during deserialization, we construct a table incrementally: we'd need to store `Rc`s somewhere before downgrading them to store them in the table.

frontend/exporter/src/id_table.rs
```diff
@@ -0,0 +1,375 @@
...
+                    .ok_or_else(|| {
+                        Self::Error::custom(&format!(
+                            "Stateful deserialization failed for id {:?}: wrong type",
+                            id
+                        ))
+                    })?
+            };
+            Ok(Self { value: kind, id })
+        }
+    }
```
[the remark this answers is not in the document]
W95Psp: You mean calling `DeserializeSeed::*` from a `Deserialize` method using the global state? Or something fancier?

frontend/exporter/src/id_table.rs
```diff
@@ -0,0 +1,375 @@
...
+        struct WithTableVisitor<T>(PhantomData<T>);
+
+        impl<'de, T: Deserialize<'de>> Visitor<'de> for WithTableVisitor<T> {
+            type Value = WithTable<T>;
+
+            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+                formatter.write_str("struct WithTable")
+            }
+
+            fn visit_seq<V>(self, mut seq: V) -> Result<Self::Value, V::Error>
```
Nadrieril, line 220: why do you need both `visit_seq` and `visit_map`? shouldn't one of them be enough?

frontend/exporter/src/id_table.rs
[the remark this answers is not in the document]
Nadrieril: No, I was hoping to get rid of global state entirely. I was thinking something like
```rust
impl<T> DeserializeSeed for &mut Option<Table> {
    type Value = WithTable<T>;
    ...
}
```
that would set the table in the `Self` mutable reference and use that throughout. Not entirely sure how that works to get access to the table when deserializing `Node` tbh

frontend/exporter/src/id_table.rs
Nadrieril: I guess my usecase caches all the `FullDef`s so never drop types without dropping the whole state. If you don't need to drop intermediate types either we can leave this as-is :+1:

frontend/exporter/src/id_table.rs
Nadrieril: I see!

frontend/exporter/src/id_table.rs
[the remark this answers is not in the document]
W95Psp: So, from what I understand (see https://github.com/serde-rs/serde/issues/881 e.g.), this composes badly with `Deserialize` traits: if a you want to use `DeserializeSeed` on a type `T` that appears on a bigger type `U`, then everything in the middle should have a `DeserializeSeed`.
Obviously, that is pretty terrible in our case, given the number of types.

frontend/exporter/src/id_table.rs
W95Psp: Ah, that's a pleasant idea!
But looks like that requires a nightly feature.
I could do `(kind.as_ptr() as *const _) as usize` :thinking: Arc pins in memory I guess, that should be OK :thinking:

Nadrieril: That feature got stabilized 2 days ago, we can safely use it :)

Nadrieril: (`.as_ptr()` shouldn't work, you have to call it with `Arc::as_ptr()`. And the `as` cast is redundant, that method returns a `*const T` already)

frontend/exporter/src/id_table.rs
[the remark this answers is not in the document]
Nadrieril: I was just reading that same issue ^^ Indeed I am now convinced this would require support from the derive macro but [that's not happening](https://github.com/serde-rs/serde/issues/881#issuecomment-1627865581).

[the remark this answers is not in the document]
W95Psp: ^^'
Yeah, exactly, so we have to keep this global state thing sadly :'( :vomiting_face:

frontend/exporter/src/id_table.rs
W95Psp: hax itself is using nightly, but we have a few other projects that use `hax-types` as a dependency, and those project are not nightly, that's a bit convoluted ^^'
Nice, will do that! That remove a type, that's good.

frontend/exporter/src/id_table.rs
W95Psp: Cool! Just pushed b17c8db8b35e680

frontend/exporter/src/id_table.rs
W95Psp: Arf, no, I can't do that as such: I need an order on IDs!

Nadrieril: I mean, you can order `usize`s, do you also need that order to be stable or something like that?

W95Psp: Currently, if a node has an id `N`, I know that its value might hold nodes whose id are strictly below `N`

So when I deserialize, I need to make sure to deserialize first the "leaf" nodes in the graph of nodes, so that I can incrementally store those in the state table. Then, nodes with higher IDs will be serialized, and might refer to nodes with lower IDs.
Do you see the point?

frontend/exporter/src/id_table.rs
```diff
@@ -0,0 +1,411 @@
...
+                        let value = seq.next_element::<T>().and_then(|inner| {
+                            inner.ok_or_else(|| de::Error::invalid_length(1, &self))
+                        });
+                        let table = std::mem::replace(
+                            &mut *DESERIALIZATION_STATE.lock().unwrap(),
+                            previous,
+                        );
+                        let value = value?;
+                        Ok((table, value))
+                    })?;
```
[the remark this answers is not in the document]
W95Psp: That's true! That'd be much easier. When implementing that, I was thinking about storing the table in between the deserialization of table and value, but now we don't need that any longer.
The only think I'm worried about is the order of parsing of elements: in JSON here we require the field `table` to appear before `value`, which is quite odd, JSON-wise. 

With this (rather complex) code, we will get an explicit error saying that `table` should come before `value`. So I think the extra complexity is OK: debugging swapped fields like that without this explicit error would be very tricky

W95Psp:
@maximebuyse: no, actually I'm using another binary format for haxmeta files (don't recall the name :sweat_smile:)
But the JSON files are now drastically smaller!
On the AES crate the JSON used to be more than 30Gigs, now it is... 91MB :partying_face:

frontend/exporter/src/id_table.rs
Nadrieril: Ah I see :thinking: You could use https://docs.rs/indexmap/latest/indexmap/ to remember the order of insertion: you're guaranteed that a leaf will be inserted before its parent. And just serialize in the order of insertion.

W95Psp:
@Nadrieril your PR is in main and I merge it into that PR :) so we should be good to go!

frontend/exporter/src/id_table.rs
W95Psp: Ah, that's tempting indeed! Gonna give that a shot.

W95Psp: ah yes, I was thinking also I could just use a Vec and use the indices as fresh IDs

W95Psp: That's even simpler than the addresses

W95Psp: I don't know actually. I kind of like the fact table is the same thing during serialization and deserialization.
I think the current ID and Session thing is fine, I agree that we could optimize that a bit more, but I'm not sure wether that'd be much more clear.
Let's stick with that design for now

Nadrieril: Sure, it works :+1:

frontend/exporter/src/id_table.rs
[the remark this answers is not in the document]
Nadrieril: If you want the order to be fixed, just use a tuple, no? In fact you can do that just to avoid needing to implement both `seq` and `map`.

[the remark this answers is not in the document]
W95Psp: oh, that's a pretty good point :sweat_smile:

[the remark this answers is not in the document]
W95Psp: let me try that and kill of that code!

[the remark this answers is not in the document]
W95Psp: Actually, I can even `#[serde(into = ...)]`/from to a tuple struct, and do my `DESERIALIZATION_STATE`/full_id_deserialization business there!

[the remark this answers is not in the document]
W95Psp: Thanks, that was a great observation! I simplified things a lot, preserving the same behavior!

Nadrieril (approved):
Looks great!

Issue #272, opened 2023-10-04
Engine: import thir: investigate `Use` ident


W95Psp:
We're currently doing something hackish in the `Use` case in `import_thir` with `Dummy`.

W95Psp:
Still relevant!

maximebuyse:
We probably want to ignore `Use` items.

PR #1030, opened 2024-10-23
Avoid bundling use items.

maximebuyse:
This avoids adding `include` statements in the generated f* with `DUMMY` idents (see #272)

### DIFF
--- a/engine/lib/dependencies.ml
+++ b/engine/lib/dependencies.ml
@@ -393,6 +393,8 @@ module Make (F : Features.T) = struct
     in
     let mut_rec_bundles =
       let mod_graph_cycles = ModGraph.of_items items |> ModGraph.cycles in
+      (* `Use` items shouldn't be bundled as they have no dependencies
+          and they have dummy names. *)
       let non_use_items =
         List.filter
           ~f:(fun item -> match item.v with Use _ -> false | _ -> true)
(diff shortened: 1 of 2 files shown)


W95Psp:
I think this makes a lot of sense.
The `Use` items are exposed in the AST, but that is a bit useless: `use` and their name indirections are erased by Rust earlier than where we hook.
Thus no any item used by Rust will ever refer to such `use` paths.

We should revisit `use` in the AST. But definitively, that's the `use` items in hax' AST which are to revisit, not the bundles.
I'd be fine merging this PR if you had a comment that explains a bit this situation, I think that's a useful PR. Wdyt @maximebuyse?

maximebuyse:
> I think this makes a lot of sense. The `Use` items are exposed in the AST, but that is a bit useless: `use` and their name indirections are erased by Rust earlier than where we hook. Thus no any item used by Rust will ever refer to such `use` paths.
> 
> We should revisit `use` in the AST. But definitively, that's the `use` items in hax' AST which are to revisit, not the bundles. I'd be fine merging this PR if you had a comment that explains a bit this situation, I think that's a useful PR. Wdyt @maximebuyse?

Yes, I think it will become irrelevant if we try to minimize the bundles (`use` items shouldn't have or be dependencies) but it is a good way to get rid of this problem for now. I'll add a comment.

W95Psp (approved):
Looks good to me, :shipit:!

PR #1039, opened 2024-10-24
Fix 914

### LINKED ISSUE #914: Tuple struct constructors are not always usable
It seems that tuple struct constructors are not usable in every context.

For instance, the following does not lax check:

```rust
struct Id(u8);

fn main() {
    let ids = [1u8].into_iter().map(Id);
}
```

[view in hax playground](https://hax-playground.cryspen.com/#fstar+tc/dc2725b99a/gist=4b7cb7dead92899c8305a4bdd2004e60)

However, the following does:

```rust
struct Id(u8);

fn main() {
    let id = Id(1u8);
}
```

maximebuyse:
Fixes #914


frontend/exporter/src/rustc_utils.rs
```diff
@@ -257,3 +257,15 @@ pub fn inline_macro_invocations<'t, S: BaseState<'t>, Body: IsBody>(
         })
         .collect()
 }
+
+pub fn closest_parent_type(
```
W95Psp: Please add a comment here

W95Psp (approved):

karthikbhargavan:
Note: the charon job is failing on the CI with this PR.

Nadrieril:
Charon's CI was red on this PR because I was in the process of fixing the impact of the recent rustc update. Please wait for CI to get fixed before merging, especially when I'm in the next room.

frontend/exporter/src/types/thir.rs
```diff
@@ -201,8 +201,24 @@ impl<'tcx, S: ExprState<'tcx>> SInto<S, Expr> for thir::Expr<'tcx> {
                         let params: Vec<rustc_middle::ty::Ty> =
                         params.map(|i| tcx.erase_late_bound_regions(i)).collect();
                         */
+                        let tcx = s.base().tcx;
+                        let constructor = if tcx.is_constructor(*def_id) {
+                            let adt_def =
+                                tcx.adt_def(rustc_utils::get_closest_parent_type(&tcx, *def_id));
```
Nadrieril, line 207: Is there a case where this isn't `tcx.adt_def(tcx.parent(def_id))`?

Nadrieril: Oh I see, if it's a tuple struct then it's only a single parent

W95Psp:
Mmh, I thought this was a THIR-only change, my bad! There is an extra field in the constant, is that it? (I'm not home I can really check)

Nadrieril:
Yeah that was an easy fix

PR #1049, opened 2024-10-28
fix(engine) Move variant constructors in recursive bundling.

### LINKED ISSUE #1021: No renaming of enum variant constructor as closure in cyclic dependencies bundling
```rust
pub mod a {
    pub enum Context {
        A(i32),
        B(i32)
    }
    pub fn f () -> Context {super::b::h()}
    impl Context {
        pub fn test (x: Option<i32>) -> Option<Context> {
            x.map(Self::A)
        }
    }
}
pub mod b {
    pub fn h () -> super::a::Context {
        super::a::Context::A(1)
    }
}
```
[Open this code snippet in the playground](https://hax-playground.cryspen.com/#fstar/21672cd0da/gist=02e1f1d6a4c867fd25f808a5c95689e5)

We produce `Playground.A.Context.v_A` for `Self::A` but it should be renamed to belong to the bundle.

maximebuyse:
Fixes #1021

### DIFF
--- a/engine/lib/concrete_ident/concrete_ident.mli
+++ b/engine/lib/concrete_ident/concrete_ident.mli
@@ -29,6 +29,7 @@ val to_debug_string : t -> string
 module Create : sig
   val fresh_module : from:t list -> t
   val move_under : new_parent:t -> t -> t
+  val ctor : t -> t
 
   val map_last : f:(string -> string) -> t -> t
   (** [map_last f ident] applies [f] on the last chunk of [ident]'s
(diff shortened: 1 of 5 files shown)


engine/lib/concrete_ident/concrete_ident.mli
```diff
@@ -29,6 +29,7 @@ val to_debug_string : t -> string
 module Create : sig
   val fresh_module : from:t list -> t
   val move_under : new_parent:t -> t -> t
+  val ctor : t -> t
```
W95Psp, line 32: Can you add a comment to explain that this adds a `Ctor` to a concrete ident, and that this is useful to craft a constructor from a variant name?

W95Psp: Let's name that `constructor` instead of `ctor` btw

W95Psp (approved):
I have only one remark, otherwsie that looks good, thanks :)

W95Psp (approved):

PR #1053, opened 2024-10-28
Avoid a panic in `FullDef`

Nadrieril:
`def_span` and `get_attrs_unchecked` panic on some foreign modules like `extern "rustc-call" { ... }` in `core`.


W95Psp (approved):
LGTM, thanks!

PR #1055, opened 2024-10-28
Include bodies in `FullDef`

Nadrieril:
So that we can start using `FullDef` instead of `hir::ItemKind`.


W95Psp (approved):
Fantastic, thank you!
Will try to migrate import_thir when I have some time!

PR #1056, opened 2024-10-28
fix(engine) Avoid wrapping return in a tuple.

### LINKED ISSUE #1044: Regression from #393 - ? not working
#393 introduced a regression, breaking Bertie F* extraction.

https://github.com/inria-prosecco/circus-green/actions/runs/11510030233

```
  bertie> error: [HAX0002] (DropReturnBreakContinue) Fatal error: something we considered as impossible occurred! Please report this by submitting an issue on GitHub!
  bertie> Details: Return/Break/Continue are expected to be gone as this point
  bertie>    --> src/tls13crypto.rs:362:1
  bertie>     |
  bertie> 362 |             algorithm.libcrux_scheme()?,
  bertie>     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  bertie>     |
  bertie> error: [HAX0002] (DropReturnBreakContinue) Fatal error: something we considered as impossible occurred! Please report this by submitting an issue on GitHub!
  bertie> Details: Return/Break/Continue are expected to be gone as this point
  bertie>    --> src/tls13crypto.rs:368:1
  bertie>     |
  bertie> 368 |             algorithm.libcrux_scheme()?,
  bertie>     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  bertie>     |
```

maximebuyse:
Fixes #1044.

This fixes phase `LocalMutation` that was wrapping `return` expressions inside a tuple (to pass a state) but the state is lost in case of a `return` so we should not try to pass the state (otherwise it is not caught by `DropReturnBreakContinue`).


W95Psp (approved):
Ah, that makes sense, thanks, looks great!

W95Psp:
The test failure looks suspicious though, do you have an idea why we have this change @maximebuyse?

maximebuyse:
@W95Psp I fixed the issue with the typing of `return`s and the bug introduced for the proverif backend is now gone. Should we merge now?

W95Psp:
Nothing changed in the generation of PV, but the verification doesn't work any longer.
The proverif TC is not under CI and relies on patches, so we can't say what broke it.
Let's go ahead and merge, it will be hard to diagnostic what broke bertie PV.
We need to (1) remove the patches and then (2) add CI there

karthikbhargavan:
@jschneider-bensch may want to look at why PV verification fails

PR #1062, opened 2024-10-29
fix(engine) Remove attributes on bundle aliases and filter out NotImplementedYet.

### LINKED ISSUE #1045: Engine: rec. bundles: attributes are preserved
Seems like attributes are preserved: the attributes of an item `i` are transported to the `alias` item.
This is likely to cause issues, especially with UUID-based attributes, which are required to be unique.

```rust
mod a {
/// A comment on `a::f`
pub fn f() {super::b::f()}
}
mod b {
/// A comment on `b::f`
pub fn f() {super::a::f()}
}
```
[Open this code snippet in the playground](https://hax-playground.cryspen.com/#fstar/7cf8c4885c/gist=ac7f2956fefba2895b1a872b9e39bc29)

cc @maximebuyse

### LINKED ISSUE #1058: Recursive bundling tries to import submodules
```rust
mod a {
    pub fn f () {
        super::b::i()
    }
    pub mod aa {
        pub fn g(){}
    }
}

mod b {
    pub fn i() {super::a::f()}
}

```
[Open this code snippet in the playground](https://hax-playground.cryspen.com/#fstar/7cf8c4885c/gist=e733e19f5d4940c1b835e5e408ddf407)

In the generated code for module `a` we have `include Playground.A.Rec_bundle_927482861 {aa as aa}`. This shouldn't be here and doesn't lax-check.

The fix for this should be similar to https://github.com/hacspec/hax/pull/1030

maximebuyse:
Fixes #1045. Fixes #1058

### DIFF
--- a/engine/lib/dependencies.ml
+++ b/engine/lib/dependencies.ml
@@ -383,7 +383,11 @@ module Make (F : Features.T) = struct
     let aliases =
       List.map (old_new :: variants_renamings old_new)
         ~f:(fun (old_ident, new_ident) ->
-          { item with v = Alias { name = old_ident; item = new_ident } })
+          {
+            item with
+            v = Alias { name = old_ident; item = new_ident };
+            attrs = [];
+          })
     in
     item' :: aliases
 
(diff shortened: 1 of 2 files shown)


W95Psp (approved):
Looks good, thanks.
Do you have an idea why we were generating so many `include`s in module `Cyclic_modules.fst` before?

PR #1064, opened 2024-10-29
Move to annotated core lib

cmester0:


W95Psp:
This probably is a bit stale name-wise, but this should be easy to patch.

W95Psp (approved):

PR #1065, opened 2024-10-29
HaxLib Interface Macro

cmester0:
Add interface macro to replace body of hax functions with todo!()

### DIFF
--- a/hax-lib-macros/src/lib.rs
+++ b/hax-lib-macros/src/lib.rs
@@ -381,6 +381,28 @@ pub fn ensures(attr: pm::TokenStream, item: pm::TokenStream) -> pm::TokenStream 
     .into()
 }
 
+/// Exclude this item from the Hax translation.
+#[proc_macro_error]
+#[proc_macro_attribute]
+pub fn interface(attr: pm::TokenStream, item: pm::TokenStream) -> pm::TokenStream {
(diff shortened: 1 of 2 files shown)


cmester0:
I have been using this macro in the annotated-core. I would expect `-i` flag to do this functionality, however it seems not to work for removing just the body with `+:`.

W95Psp:
We just chatted about this with @cmester0, let's add a proc-macro crates for macro helpers in the annotated lib core.

W95Psp (changes requested):
As explained in my previous comment, let's add a proc-macro crates for macro helpers in the annotated lib core, and move this there. That makes more sense, and we will be able to name it "backend_primitive" or something like that, which makes sense in the context of the annotated core.

hax-lib-macros/src/lib.rs
```diff
@@ -1000,7 +1022,7 @@ pub fn refinement_type(mut attr: pm::TokenStream, item: pm::TokenStream) -> pm::
             }
 
             #[::hax_lib::exclude]
-            impl #generics ::std::ops::Deref for #ident <#generics_args> {
+            impl #generics ::core::ops::Deref for #ident <#generics_args> {
```
cmester0, line 1025: This is still relevant

PR #1067, opened 2024-10-29
feat(book): allow F* extraction + TC from within the book

W95Psp:
This PR:
 - fixes the book (things were broken because we dropped the theme, that contained custom code I wrote)
 - add buttons to extract and typecheck from within the book, using the API of the playground
 - do some refactor, add a introduction for the examples


book/src/examples/intro.md
```diff
@@ -0,0 +1,12 @@
+# Examples
+
+This chapter contains various examples that demonstrates how hax can
+be used to prove properties about programs. Each example is
+self-contained. hax being a tool that can extract Rust to various
+backends, this section provides examples for each backends..
```
maximebuyse, line 6: "each backend" (without s), and probably only one dot at the end.

maximebuyse (changes requested):
Just noticed a typo but it looks great otherwise!

book/src/examples/intro.md
W95Psp: Good catch, thanks!

maximebuyse (approved):

karthikbhargavan (approved):
This needs some text polish, but let's merge this PR so that the code/checking features are merged.

PR #1069, opened 2024-10-29
Add `ImplExpr` information to associated types in an impl block

Nadrieril:
Along with constness information, this is one of the last remaining queries we were doing in charon.


W95Psp (approved):
That looks very nice :D 
I might give a shot at modifying import_thir to this FullDef type tomorrow if I find time! :smiley:
Let's merge if CI is happy!

PR #1073, opened 2024-10-30
Concrete ident for coq

cmester0:
Move concrete ident modifications for Coq backend (#987) to separate PR.

### DIFF
--- a/test-harness/src/snapshots/toolchain__literals into-coq.snap
+++ b/test-harness/src/snapshots/toolchain__literals into-coq.snap
@@ -47,13 +47,13 @@ Export Hax_lib (t_int).
 
 Record t_Foo : Type :=
   {
-    f_field : t_u8;
+    Foo_f_field : t_u8;
   }.
 Arguments t_Foo:clear implicits.
 Arguments t_Foo.
 Arguments Build_t_Foo.
 #[export] Instance settable_t_Foo : Settable _ :=
-  settable! (@Build_t_Foo) <f_field>.
+  settable! (@Build_t_Foo) <Foo_f_field>.
 
 (* NotImplementedYet *)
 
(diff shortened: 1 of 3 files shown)


cmester0:
For Coq I cannot name tuple fields, 0,1,2,.., furthermore fields need to be unique, so we should use the path in the name of the fields. This naming also has to be consistent when used elsewhere, so I changing this in concrete ident.

test-harness/src/snapshots/toolchain__reordering into-coq.snap
```diff
@@ -42,33 +42,33 @@ Import RecordSetNotations.
...
-| Foo_B.
+| t_Foo_A
+| t_Foo_B.
 Arguments t_Foo:clear implicits.
 Arguments t_Foo.
 
 Record t_Bar : Type :=
   {
-    0 : t_Foo;
+    t_Bar_0 : t_Foo;
```
cmester0, line 52: This is a solution to tuple structs.

test-harness/src/snapshots/toolchain__literals into-coq.snap
```diff
@@ -47,13 +47,13 @@ Export Hax_lib (t_int).
 
 Record t_Foo : Type :=
   {
-    f_field : t_u8;
+    t_Foo_f_field : t_u8;
```
cmester0, line 50: This is a solution to overlapping field names in structs.

W95Psp: Oh, I see!
Then you need to improve the name policies API: the binding `enum_constructor_name_transform` is what you want, but you need to add an extra argument "struct_name" or something, just as I did for `enum_constructor_name_transform`.
Do you see what I mean?

cmester0: Ahh, yes. Makes sense.

cmester0: This will also fix the issue I had, with overwriting keywords (e.g. `end`).

W95Psp: Great! :tada:

cmester0:
Closed in favor of #1108, which has more features

PR #1074, opened 2024-10-30
feat(engine): profile each phase (mem usage, time, number of items)

### LINKED ISSUE #816: Engine: investigate extensible variants to replace `Diagnostics.Phases`
Currently, phase names are declared in one monolithic file: adding a phase requires creating two files and adding a variant to another file.
Extensible variants would help there, and we could use `__MODULE__` and a PPX to generate names automatically (since modules names are unique).

### LINKED ISSUE #937: Engine: phases: record time
We should record how long each phase run, so that we can spot regressions and slow paths (we had a pretty bad complexity, fixed by @maximebuyse in #877, catching that beforehand would be useful!)

W95Psp:
This PR:
 - adds profiling information to the engine (memory usage, time, number of item processed)
 - makes phase name generated so that they are following strictly their file name 

Fixes #937, fixes #816

### DIFF
--- a/engine/lib/phases/phase_bundle_cycles.mli
+++ b/engine/lib/phases/phase_bundle_cycles.mli
@@ -1,0 +1,6 @@
+(** This phase makes sure the items don't yield any cycle,
+namespace-wise. It makes the namespaces a tree. It does so by creating
+namespaces we call bundles, in which we regroup definitions that would
+otherwise yield cycles. *)
+
+module Make : Phase_utils.UNCONSTRAINTED_MONOMORPHIC_PHASE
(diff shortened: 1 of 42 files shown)


engine/lib/phases/phase_bundle_cycles.mli
```diff
@@ -0,0 +1,6 @@
+(** This phase makes sure the items don't yield any cycle,
+namespace-wise. It makes the namespaces a tree. It does so by creating
```
maximebuyse, line 2: It can be a DAG, not necessarily a tree.

maximebuyse (approved):
Looks great!

engine/lib/phases/phase_bundle_cycles.mli
W95Psp: thanks, pushing a fix! feel free to change whatever I wrote there, btw!

PR #1077, opened 2024-10-30
fix(engine) Keep late_skip attributes in bundles

### LINKED ISSUE #1076: Cyclic bundles: `Alias` are created for items with `late_skip`
Typically, `hax_lib` attributes expand to items that are tagged with `late_skip`: a precondition is a boolean-returning function tagged with `late_skip`.
In the cyclic bundle logic, we need to persist that late skip attribute in the `Alias` item if it was present at first in the original item.

```rust
pub mod a {
    pub fn f() {
        super::b::f()
    }
}
pub mod b {
    #[hax_lib::requires(true)]
    pub fn f() {
        super::a::f()
    }
}

```
[Open this code snippet in the playground](https://hax-playground.cryspen.com/#fstar/65dc6ab578/gist=d48f698a4ebc7b0bf9e59d1a9e7a3715)

maximebuyse:
Fixes #1076

### DIFF
--- a/engine/lib/dependencies.ml
+++ b/engine/lib/dependencies.ml
@@ -383,11 +383,11 @@ module Make (F : Features.T) = struct
     let aliases =
       List.map (old_new :: variants_renamings old_new)
         ~f:(fun (old_ident, new_ident) ->
-          {
-            item with
-            v = Alias { name = old_ident; item = new_ident };
-            attrs = [];
-          })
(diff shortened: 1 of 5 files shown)


W95Psp (approved):
Can you add an example? (i.e. the one of the issue or something similar?)
Otherwise, that looks good to me!

PR #1079, opened 2024-10-30
Include defaulted items in `FullDefKind::TraitImpl`

Nadrieril:
This reworks the representation of associated impl items. The problem this solves is:
```rust
trait Trait {
    type Type: Clone = i32;
}
impl Trait for () {}
```
Here the impl implicitly reuses the type provided by the trait. The version of the type that's virtually in the impl does not have a `DefId` however, which makes it hard to get information about it. This PR adds a representation for these defaulted items, and gathers the necessary information.

Now `FullDefKind::TraitImpl.items` always has the same number of entries (and in the same order) as `FullDefKind::Trait.items`.

There remains some work to do for methods because I don't yet handle that in charon. This is also unfortunately limited wrt GATs; we'd have to construct our own `ParamEnv`s to solve that properly so I'm leaving it as future work.


Nadrieril:
Charon fix is ready

W95Psp (approved):

PR #1080, opened 2024-10-31
feat(engine): keep track of parent items in spans

W95Psp:
This PR:
 - makes spans carry a hint about their parent item;
 - makes errors carry a hint about which item the error comes from;
 - adds a `--stats` flag to the verb `into`, printing information about how many items the engine was able to process and a total number of items.

This will be quite useful for mass testing the engine: we'll be able to have statistics about how many items in a crate we can translate.


maximebuyse (approved):

W95Psp:
Note to myself: is that a regression because of span comparison/equality somewhere in the code?
Trying to define by hand the equality and comparison for spans ommitting the owner id might fix this
But we should not rely on span comparison...

We may want to use the "real" equality inside the module span, and then expose an always true equality/comparison for consumer of that module.

PR #1092, opened 2024-10-31
feat(ci): test rust by examples (extract to F*)

W95Psp:
Fixes https://github.com/hacspec/hax/issues/540, but instead of trying two snippets, it's trying 157 snippets.

41 are not going though, on a total of 198: we have about 80% success rate here.

This needs #1090.

### DIFF
--- a/.utils/rust-by-example.js
+++ b/.utils/rust-by-example.js
@@ -1,4 +1,5 @@
 // This script expects Rust By Example to be in current directory
+// (clone the repo https://github.com/rust-lang/rust-by-example)
 
 const fs = require('fs');
 const SRC_DIR = 'src';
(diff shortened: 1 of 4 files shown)


.utils/rust-by-example.js
```diff
@@ -0,0 +1,142 @@
+// This script expect Rust By Example to be in current directory
```
franziskuskiefer, line 1:
```suggestion
// This script expects Rust By Example to be in current directory
```

franziskuskiefer, line 1: Please add here where to get it from and how to use this file.

franziskuskiefer (approved):
lgtm with the comments addressed.

.utils/rust-by-example.js
W95Psp: Done!

PR #1094, opened 2024-11-04
Fix ml-kem ci job.

### LINKED ISSUE #1086: Remove/fix libcrux ml-kem CI job
The current job tries to apply the previous patches we had.
We don't have them anymore, let's drop that job and reuse the one defined in the repo of libcrux.

maximebuyse:
Fixes #1086

### DIFF
--- a/.github/workflows/mlkem.yml
+++ b/.github/workflows/mlkem.yml
@@ -40,61 +40,15 @@ jobs:
         with:
           repository: hacl-star/hacl-star
           path: hacl-star
-
-      - name: 🏃 Extract the Kyber reference code
-        run: |
-          eval $(opam env)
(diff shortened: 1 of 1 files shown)


maximebuyse:
I tried the new job with https://github.com/hacspec/hax/actions/runs/11665242466/job/32477483999 and it does what we want.

The configuration is partly copied from https://github.com/cryspen/libcrux/blob/main/.github/workflows/hax.yml
It would be nice to be able to [share](https://docs.github.com/en/actions/sharing-automations/reusing-workflows) some workflow parts with the libcrux repo but that would require a bit of refactoring.

franziskuskiefer (approved):
Thanks!
Let's get this in. We can update this with the workbench, or put a shared action into [cryspen/actions](https://github.com/cryspen/actions) later.

PR #1100, opened 2024-11-05
Fix spans in error reporting with unicode characters.

### LINKED ISSUE #1099: Unicode in Comments shifts spans and can crash hax
Here is an example where this leads to minor shifts: https://hax-playground.cryspen.com/#fstar/a80e9862e6/gist=a860cf1a1b80fe4744eb30c6417cd4ff

Here is an example where it leads to a crash, if this leads to slicing strings in the middle of a unicode glyph: https://hax-playground.cryspen.com/#fstar/a80e9862e6/gist=88c2ac3756b8420a0747bae979e5aebc

maximebuyse:
Fixes #1099


maximebuyse:
This fix only works when the unicode characters are on a different line from the error.

But something like:

```rust

fn f(x: usize) {
    let mut bar = 0;
    let mut opt = Some(x);
    /*äöü*/opt.take().map(|foo| bar = foo+1);
}
```
[Open this code snippet in the playground](https://hax-playground.cryspen.com/#fstar/a80e9862e6/gist=34d27acab982991e8cd8be85e8e09436)

is still wrong

maximebuyse:
> This fix only works when the unicode characters are on a different line from the error.
> 
> But something like:
> 
> ```rust
> fn f(x: usize) {
>     let mut bar = 0;
>     let mut opt = Some(x);
>     /*äöü*/opt.take().map(|foo| bar = foo+1);
> }
> ```
> 
> [Open this code snippet in the playground](https://hax-playground.cryspen.com/#fstar/a80e9862e6/gist=34d27acab982991e8cd8be85e8e09436)
> 
> is still wrong

I pushed a new version fixing this as well. I introduced a new dependency to take care of this which simplifies and should hopefully be more reliable.

franziskuskiefer (approved):
I don't like adding dependencies for something small like this.
But we want to surface backend errors nicely as well. So something may be needed anyway. Let's keep this on the radar to see how to deal with it.
But this fixes the issue at hand. Let's get it in.

PR #1101, opened 2024-11-05
fix(engine) Drop return/break/continue inside closures.

### LINKED ISSUE #1089: Question mark in a closure gives an error
```rust

fn test(x: Option<i32>, y: Option<i32>) -> Option<i32> {
    x.map(|i| Some(i + y?))?
}

```
[Open this code snippet in the playground](https://hax-playground.cryspen.com/#fstar/31458e07fe/gist=1a3da6e8ad09c6d1e0eebaab0f87c65c)

This gives the following error:

```
error: [HAX0002] (DropReturnBreakContinue) Fatal error: something we considered as impossible occurred! Please report this by submitting an issue on GitHub!
Details: Return/Break/Continue are expected to be gone as this point
 --> src/lib.rs:3:24
  |
3 |     x.map(|i| Some(i + y?))?
```

This is another regression from #393.

maximebuyse:
Fixes #1089.

Closures were not treated in the elimination of `return`/`break`/`continue`.

### DIFF
--- a/engine/lib/phases/phase_drop_return_break_continue.ml
+++ b/engine/lib/phases/phase_drop_return_break_continue.ml
@@ -118,6 +118,18 @@ module%inlined_contents Make (F : Features.T) = struct
                  in exit nodes. **)
       end
 
+    let closure_visitor =
+      let module Visitors = Ast_visitors.Make (F) in
+      object
+        inherit [_] Visitors.map as super
+
+        method! visit_expr' () e =
+          match e with
+          | Closure ({ body; _ } as closure) ->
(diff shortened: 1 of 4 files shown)


franziskuskiefer:
As discussed, this appears to need more tests.
Do you want to add more tests separately or in here?

maximebuyse:
> As discussed, this appears to need more tests. Do you want to add more tests separately or in here?

I think it makes more sense to do it separately (there is already a new test in here that covers the corresponding bug).

franziskuskiefer (approved):
Ok, then let's get this in and please file a follow up to increase test coverage for break/continue/return?.

PR #1115, opened 2024-11-13
Cryspen green.

maximebuyse:
This adds CI jobs for running fstar extraction on some important crates that use hax.


franziskuskiefer (approved):
Thanks

PR #1116, opened 2024-11-13
feat(frontend/cli): improve options

W95Psp:
This is useful for the workbench.
This PR:
 - adds a hidden option `cargo hax __json` that just consumes stdin and parse that as CLI options
 - implements `ToString` for a few types
 - make more fields public for types in cli_options

### DIFF
--- a/hax-types/src/cli_options/mod.rs
+++ b/hax-types/src/cli_options/mod.rs
@@ -215,9 +215,22 @@ enum InclusionKind {
 
 #[derive_group(Serializers)]
 #[derive(JsonSchema, Debug, Clone)]
-struct InclusionClause {
-    kind: InclusionKind,
-    namespace: Namespace,
+pub struct InclusionClause {
+    pub kind: InclusionKind,
+    pub namespace: Namespace,
+}
+
+impl ToString for InclusionClause {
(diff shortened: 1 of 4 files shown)


hax-types/src/cli_options/mod.rs
```diff
@@ -215,12 +215,25 @@ enum InclusionKind {
...
+pub struct InclusionClause {
+    pub kind: InclusionKind,
+    pub namespace: Namespace,
 }
 
-fn parse_inclusion_clause(
+impl ToString for InclusionClause {
+    fn to_string(&self) -> String {
+        let kind = match self.kind {
+            InclusionKind::Included(DepsKind::Transitive) => "+",
```
maximebuyse, line 226: It's a detail but it might be better to have `"+~"`, `"+:"` and the correspondance with the different cases for `InclusionKind` in a method or in a global variable that would be used both here and in `parse_inclusion_clause`. Just to avoid that somehow they end up diverging in some way.

W95Psp: That's a great idea, thanks, I pushed a fix! :)

maximebuyse (approved):

PR #1117, opened 2024-11-13
Upstream fstar core changes from cryspen-sandwich.

### LINKED ISSUE #979: Upstream fstar core additions from `cryspen-sandwich` to `main`
Some additions to fstar core were made in the `cryspen-sandwich` branch. It would be great to upstream them to the `main` branch. However some more changes were made in this branch that shouldn't be merged to main.

maximebuyse:
closes #979

### DIFF
--- a/examples/chacha20/proofs/fstar/extraction/Makefile
+++ b/examples/chacha20/proofs/fstar/extraction/Makefile
@@ -54,7 +54,7 @@ all:
 	$(MAKE) verify
 
 # Default hax invocation
-HAX_CLI = "cargo hax into fstar"
+HAX_CLI = "cargo hax into fstar --z3rlimit 40"
 
 # If $HACL_HOME doesn't exist, clone it
 ${HACL_HOME}:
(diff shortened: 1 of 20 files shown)


proof-libs/fstar/core/Std.Io.Impls.fsti
```diff
@@ -0,0 +1,2 @@
+module Std.Io.Impls
+val placeholder: unit
```
W95Psp, line 2: Why is this needed?

W95Psp (approved):
Looks good, I only have one interrogation

proof-libs/fstar/core/Std.Io.Impls.fsti
maximebuyse: I don't know, we can skip it and add it later if needed

W95Psp: Can you try to git blame maybe? so that we can maybe correlate the date and author of this change with something in sandwich for instance?

maximebuyse:
I don't understand the CI failure. The core changes are unrelated to the error, and chacha20 type-checking works locally for me (maybe updating the fstar version used by the CI could help?)

maximebuyse:
> I don't understand the CI failure. The core changes are unrelated to the error, and chacha20 type-checking works locally for me (maybe updating the fstar version used by the CI could help?)

After rebasing the CI passes. Let's merge.

PR #1120, opened 2024-11-18
doc(book): architecture of hax

### LINKED ISSUE #752: Book: document the architecture of hax
We changed hax to use 'haxmeta' files recently, we need to document that.
Ideally, we should have both a diagram that explains the interactions between the different Rust crates and the engine, with some explanations.

W95Psp:
This PR adds a page to the book, documenting the high-level architecture of hax.

Fixes #752

### DIFF
--- a/book/src/contributing/architecture.md
+++ b/book/src/contributing/architecture.md
@@ -63,7 +63,7 @@ The engine executes a sequence of **phases**, which are determined based on the 
 1. **Input:** Takes a list of items from an AST with specific feature constraints.
 2. **Output:** Transforms these items into a new AST type, potentially enabling or disabling features through type-level changes.
 
-The phases can be found in the `engin/lib/phases/` folder.
+The phases can be found in the [`Phases`](https://hacspec.org/hax/engine/hax-engine/Hax_engine/Phases/index.html) module.
 
 ### Backend Code Generation
 
--- a/book/src/contributing/architecture.md
+++ b/book/src/contributing/architecture.md
@@ -5,6 +5,8 @@ Hax is a software pipeline designed to transform Rust code into various formal v
 1. **The Frontend** (written in Rust)
 2. **The Engine** (written in OCaml)
 
+The frontend hooks into the Rust compiler, producing a abstract syntax tree for a given crate. The engine then takes this AST in input, applies various transformation, to reach in the end the language of the backend: F*, Coq...
+
 ## The Frontend (Rust)
 
 The frontend is responsible for extracting and exporting Rust code's abstract syntax trees (ASTs) in a format suitable for processing by the engine (or by other tools).
--- a/book/src/contributing/architecture.md
+++ b/book/src/contributing/architecture.md
@@ -11,7 +11,7 @@ The frontend is responsible for extracting and exporting Rust code's abstract sy
 
 ### `hax-frontend-exporter` Library
 
-This library mirrors the internal types of the Rust compiler (`rustc`) that constitute the **HIR** (High-Level Intermediate Representation), **THIR** (Typed High-Level Intermediate Representation), and **MIR** (Mid-Level Intermediate Representation) ASTs. It extends them with additional information such as attributes, trait implementations, and removes IDs indirections.
+This library mirrors the internal types of the Rust compiler (`rustc`) that constitute the **HIR** (High-Level Intermediate Representation), **THIR** (Typed High-Level Intermediate Representation), and **MIR** (Mid-Level Intermediate Representation) ASTs. It extends them with additional information such as attributes, trait implementations, and removes ID indirections.
 
 **`SInto` Trait:** The library defines an entry point for translating a given `rustc` value to its mirrored hax version using the `SInto` trait (stateful `into`). For a value `x` of type `T` from `rustc`, if `T` is mirrored by hax, then `x.sinto(s)` produces an augmented and simplified "hax-ified" AST for `x`. Here, `s` represents the state holding information about the translation process.
 
--- a/book/src/contributing/architecture.md
+++ b/book/src/contributing/architecture.md
@@ -9,11 +9,11 @@ Hax is a software pipeline designed to transform Rust code into various formal v
 
 The frontend is responsible for extracting and exporting Rust code's abstract syntax trees (ASTs) in a format suitable for processing by the engine (or by other tools).
 
-### `hax-frontend-exporter` Library
+### [`hax-frontend-exporter` Library](https://hacspec.org/hax/frontend/hax_frontend_exporter/index.html)
 
 This library mirrors the internal types of the Rust compiler (`rustc`) that constitute the **HIR** (High-Level Intermediate Representation), **THIR** (Typed High-Level Intermediate Representation), and **MIR** (Mid-Level Intermediate Representation) ASTs. It extends them with additional information such as attributes, trait implementations, and removes ID indirections.
 
-**`SInto` Trait:** The library defines an entry point for translating a given `rustc` value to its mirrored hax version using the `SInto` trait (stateful `into`). For a value `x` of type `T` from `rustc`, if `T` is mirrored by hax, then `x.sinto(s)` produces an augmented and simplified "hax-ified" AST for `x`. Here, `s` represents the state holding information about the translation process.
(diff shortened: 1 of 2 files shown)


book/src/contributing/architecture.md
```diff
@@ -0,0 +1,84 @@
+# Architecture
+
+Hax is a software pipeline designed to transform Rust code into various formal verification backends such as **F\***, **Coq**, **ProVerif**, and **EasyCrypt**. It comprises two main components:
+
+1. **The Frontend** (written in Rust)
+2. **The Engine** (written in OCaml)
+
+## The Frontend (Rust)
+
+The frontend is responsible for extracting and exporting Rust code's abstract syntax trees (ASTs) in a format suitable for processing by the engine (or by other tools).
```
franziskuskiefer, line 10: You start describing pieces below without putting them into relation. The section up here should tell what's coming and how they are related. The same goes for the top level section.

It should also contain what's public things and what are internal things.

book/src/contributing/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+2. **AST Transformation:** Applies `sinto` on each item to generate the hax-ified AST.
+3. **Output Generation:** Outputs the mirrored items into a `haxmeta` file within the `target` directory.
+
+### `cargo-hax` Binary
+
+`cargo-hax` provides a `hax` subcommand for Cargo, accessible via `cargo hax --help`. It serves as the command-line interface for hax, orchestrating both the frontend and the engine.
+
+**Workflow:**
+
+1. **Custom Build Execution:** Runs `cargo build`, instructing Cargo to use `hax-driver` instead of `rustc`.
```
franziskuskiefer, line 32: This kind of description is great but we should make sure that it doesn't go out of date when changes are done.

book/src/contributing/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+### Importing and Simplifying ASTs
+
+- **AST Importation:** Imports the hax-ified Rust THIR AST.
+- **Internal AST Conversion:** Converts the imported AST into a simplified and opinionated internal AST designed for ease of transformation and analysis.
+
+### Internal AST and Features
+
+The internal AST is defined using a **functor** that takes a list of type-level booleans, referred to as **features**, and produces the AST types accordingly.
+
+Features are for instances, mutation, loops, unsafe code. A full list is available in `engine/lib/features.ml`.
```
franziskuskiefer, line 53: Lets make that `features.ml` a link.

book/src/contributing/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+
+**Feature Witnesses:**
+
+- On relevant AST nodes, feature witnesses are included to enforce constraints at the type level.
+- **Example:** In the `loop` expression constructor, a witness of type `F.loop` is used, where `F` represents the current feature set. If `F.loop` is an empty type, constructing a `loop` expression is prohibited, ensuring that loops are disallowed in contexts where they are not supported.
```
franziskuskiefer, line 58: This is an example for the bullet point above.
I think I'd drop bullet points in this section for witnesses.

```suggestion
  **Example:** In the `loop` expression constructor, a witness of type `F.loop` is used, where `F` represents the current feature set. If `F.loop` is an empty type, constructing a `loop` expression is prohibited, ensuring that loops are disallowed in contexts where they are not supported.
```

book/src/contributing/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Transformation Phases
+
+The engine executes a sequence of **phases**, which are determined based on the target backend. Each phase:
+
+1. **Input:** Takes a list of items from an AST with specific feature constraints.
+2. **Output:** Transforms these items into a new AST type, potentially enabling or disabling features through type-level changes.
+
+The phases can be found in the `engin/lib/phases/` folder.
```
franziskuskiefer, line 67: Link

book/src/contributing/architecture.md
```diff
@@ -0,0 +1,84 @@
...
+
+After completing the transformation phases:
+
+1. **Backend Printer Invocation:** Calls the printer associated with the selected backend to generate the target code.
+2. **File Map Creation:** Produces a map from file names to their contents, representing the generated code.
+3. **Output Serialization:** Outputs the file map and additional information (e.g., errors) as JSON to `stderr`.
+
+### Communication Protocol
+
+The engine communicates asynchronously with the frontend using a protocol defined in `hax_types::engine_api::protocol`. This communication includes:
```
franziskuskiefer, line 79: Link

franziskuskiefer (changes requested):
Thanks. Generally looks good with a couple nits.

book/src/contributing/architecture.md
```diff
@@ -0,0 +1,84 @@
+# Architecture
+
+Hax is a software pipeline designed to transform Rust code into various formal verification backends such as **F\***, **Coq**, **ProVerif**, and **EasyCrypt**. It comprises two main components:
+
+1. **The Frontend** (written in Rust)
+2. **The Engine** (written in OCaml)
```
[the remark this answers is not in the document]
W95Psp: Yes, I think we should drop that.
For me, the engine _contains_ the backends.

book/src/contributing/architecture.md
W95Psp: I agree, but I don't see how we could enforce that here (since it's a structural description)

book/src/contributing/architecture.md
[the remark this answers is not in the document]
franziskuskiefer: This is a bigger change than just this document. We are using three parts everywhere so far, including [your blog post](https://hacspec.org/blog/posts/hax-v0-1/). And I'm not sure I agree with dropping it. There should be a distinction between transformations and printing. Not doing that just mushes everything into one big module.

book/src/contributing/architecture.md
franziskuskiefer: Hm, I don't see a new paragraph here.

book/src/contributing/architecture.md
franziskuskiefer: This is a process question. I'll add this to the new processes we are implementing.

franziskuskiefer (changes requested):
Why are there code changes in here now?

W95Psp:
Huh, sorry, while doing something else I messed with this branch (I think I rebased that branch with a local main behind origin)... that thing is fixed now

book/src/contributing/architecture.md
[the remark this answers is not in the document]
W95Psp: Mmh, I see what you mean, but in terms of software architecture, the engine does contain the backends. So, in this architecture document, I think we should not see backends as an artifact outside of the engine. Maybe I can detail that the engine is composed of a library (the phases, the AST), `n` backends, and a binary that wraps everything together?
I could also add a small diagram for the whole architecture. That'd make things very clear.
WDYT?

book/src/contributing/architecture.md
W95Psp: Sorry, messed up with the history, I re-commited, that's commit afff4b381b6454b5d53521d4c03729ac007dd8c7 now

book/src/contributing/architecture.md
W95Psp: Makes sense. So nothing to do in this PR, right? Let's resolve if we agree on this being "future work" (outside of this PR)

book/src/contributing/architecture.md
W95Psp: Should be good now

book/src/contributing/architecture.md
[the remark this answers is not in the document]
franziskuskiefer: Yeah that would be good. Do you want to do that in here or in a follow up?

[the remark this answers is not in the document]
W95Psp: I can do that diagram here!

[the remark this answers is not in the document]
franziskuskiefer: Did you want to do it? I'm not seeing one :)

franziskuskiefer (approved):
Not all comments resolved, yet, but generally lgtm. Please file follow ups for the ones you don't want to address here.

book/src/contributing/architecture.md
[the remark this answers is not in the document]
W95Psp: Yeah, I started something quickly but had no time. The week is pretty dense already, I'll file a follow up.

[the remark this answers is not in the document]
W95Psp: https://github.com/hacspec/hax/issues/1164

PR #1121, opened 2024-11-20
fix typos in the book

ruipedro16:


W95Psp (approved):
Thanks, that's great! Let's merge! :)

PR #1123, opened 2024-11-21
Engine: source maps: fix bugs, add features, enable for Coq

W95Psp:
This PR:
 - makes cargo hax save source maps as files if the engine produce any
 - fix a bug in the source map encoding
 - output source maps for the Coq backend

To test the source maps for Coq:
 - `cargo hax into coq` on a crate
 - https://evanw.github.io/source-map-visualization/, upload one `.v` file and it's corresponding `.v.map` file (no need for the rust source files, they are embedded in sourcemaps)


cmester0 (approved):
LGTM

PR #1134, opened 2024-11-25
Opaque extensions

### LINKED ISSUE #1119: Allow `impl`s to be opaque
This is needed for https://github.com/hacspec/hax/issues/1070#issuecomment-2476182283, we already have opaque types of which this will be an extension.

maximebuyse:
Fixes #1119

We implement [this design](https://docs.google.com/document/d/1D_Cqa3JU_1ktm4LMZaLY6gowq2-2xdthhoQ3lPJT9_Q/edit?tab=t.kk9pdzl7lqgf). The attribute `opaque` is introduced and allows to drop the content (erase) of functions, types, trait impls, consts, etc. to make them abstract. In this case we produce a type signature and an abstract definition (using `assume` in fstar). The interface-only include flag (`cargo hax into -i '+:.....`) allows to erase easily many items. It relies on default rules that say if a selected item should be erased or not (see design doc).


maximebuyse:
Adding a note here that I noticed while reviewing the test result for `tests/cli/interface-only`. This previously produced only an `fsti` file, it now produces only an `fst` file. This makes sense because it uses the `-i '+:'` flag but no `--interfaces` flag for the fstar backend. As we now provide definitions for erased items we get them in an fst file. The definitions can still be discarded using `--interfaces +!**`. Maybe it would be worth documenting this but I am not sure where (I started with updating the documentation of the `-i` flag).

engine/lib/import_thir.ml
```diff
@@ -1440,26 +1440,49 @@ let cast_of_enum typ_name generics typ thir_span
...
+             |> not ->
+          true
+      | _ -> false
+    in
+    let attrs =
+      if erased_by_type_only && not erased_by_user then
+        Attr_payloads.to_attr Erased span :: attrs
+      else attrs
+    in
+    let erased = erased_by_user || erased_by_type_only in
```
W95Psp: Then, can you add documentation explaining roughly what is the logic, on bindings `erased_by_user`, `erased_by_type_only` and `erased`?
Similarly, please document the `if erased_by_type_only && not erased_by_user then`: the logic is if `erased` is true, then we add a attribute `erased` unless it already exists, right?

engine/lib/import_thir.ml
```diff
@@ -1681,11 +1710,65 @@ and c_item_unwrapped ~ident ~drop_body (item : Thir.item) : item list =
             ~f:(fun { attributes; _ } -> not (should_skip attributes))
             items
         in
-        let has_type =
-          List.exists items ~f:(fun item ->
-              match item.kind with Type _ -> true | _ -> false)
+        let items =
+          if erased then
+            [
```
W95Psp, line 1720: Can you rewrite this from `[ the_item_expr ]` to `let dummy_assoc_item = the_item_expr in [dummy_assoc_item]`?
Or add a comment.
But we should convey the idea that here we're producing a dummy associated item on purpose

hax-lib/macros/types/src/lib.rs
```diff
@@ -130,8 +130,8 @@ pub enum AttrPayload {
     PVConstructor,
     PVHandwritten,
     TraitMethodNoPrePost,
-    /// Make a type opaque
-    OpaqueType,
+    /// Make something opaque
```
W95Psp, line 133:
```suggestion
    /// Make an item opaque
```

maximebuyse:
> Thanks, I'm reviewing now. Some snapshots are conflicting though: can you update them?

Done. I also pushed a small commit to allow making a single method opaque inside an impl (regular non-trait impl).

test-harness/src/snapshots/toolchain__attribute-opaque into-fstar.snap
```diff
@@ -23,19 +23,70 @@ info:
...
+  = note: `#[warn(deprecated)]` on by default
+
+warning: use of deprecated macro `hax_lib::opaque_type`: Please use 'opaque' instead
+ --> attribute-opaque/src/lib.rs:7:3
+  |
+7 | #[hax_lib::opaque_type]
+  |   ^^^^^^^^^^^^^^^^^^^^
+
+warning: `attribute-opaque` (lib) generated 2 warnings
+    Finished `dev` profile [unoptimized + debuginfo] target(s) in XXs'''
```
W95Psp, line 42: Let's rename all occurences of `opaque_type` into `opaque`.

W95Psp (commented):
Thanks, that looks good to me :)
I have some minor comments.

I still need to test the PR on snippets of code.

W95Psp:
> > Thanks, I'm reviewing now. Some snapshots are conflicting though: can you update them?
> 
> Done. I also pushed a small commit to allow making a single method opaque inside an impl (regular non-trait impl).

Oh great, that's a nice addition!

engine/lib/import_thir.ml
maximebuyse: Yes that's right. I'll add comments.

engine/backends/fstar/fstar_backend.ml
```diff
@@ -971,10 +994,7 @@ struct
...
                    ty );
                ] )
-    | Type { name; generics; _ }
-      when Attrs.find_unique_attr e.attrs
-             ~f:([%eq: Types.ha_payload] OpaqueType >> Fn.flip Option.some_if ())
-           |> Option.is_some ->
+    | Type { name; generics; _ } when is_erased ->
         if not ctx.interface_mode then
           Error.raise
           @@ {
```
W95Psp, line 1000: We should not error out here now: we should allow type definition to be opaque, whether we are producing a fsti, a fst or both

W95Psp:
We also need the opaque attribute to work on traits I think. We forgot that case in the design document.

```rust
#[hax_lib::opaque]
trait Foo<X, Y> {
  ...
}
```

Should be translated to:
```fstar
assume type foo: Type -> Type -> Type
```
(and for fsti `val foo: Type -> Type -> Type`)

W95Psp:
```rust
#[hax_lib::opaque]
fn opaque_function(){}

fn regular_function(){}
```
[Open this code snippet in the playground](https://hax-playground.cryspen.com/#fstar/d5bf025e17/gist=2c8433ec799e218e8f4a9ae3db084cff)

Gets translated to:
```fstar
val opaque_function: Prims.unit -> Prims.unit
assume val opaque_function': Prims.unit -> Prims.unit
let opaque_function = opaque_function'

let regular_function (_: Prims.unit): Prims.unit = ()
```

I don't think we want the `val opaque_function`, do we?

maximebuyse:
> We also need the opaque attribute to work on traits I think. We forgot that case in the design document.
> 
> ```rust
> #[hax_lib::opaque]
> trait Foo<X, Y> {
>   ...
> }
> ```
> 
> Should be translated to:
> 
> ```fstar
> assume type foo: Type -> Type -> Type
> ```
> 
> (and for fsti `val foo: Type -> Type -> Type`)

Yes that can be nice when we have unallowed `&mut` in method signatures. I implemented that but maybe we should be careful about an opaque trait that has non-opaque `impl`s. I think it could be nice to somehow enforce this (but maybe it is not a priority.

W95Psp:
> Yes that can be nice when we have unallowed `&mut` in method signatures. I implemented that but maybe we should be careful about an opaque trait that has non-opaque `impl`s. I think it could be nice to somehow enforce this (but maybe it is not a priority.

Indeed, let's file an issue with a good description then. I think we don't care for this PR.

maximebuyse:
> > Yes that can be nice when we have unallowed `&mut` in method signatures. I implemented that but maybe we should be careful about an opaque trait that has non-opaque `impl`s. I think it could be nice to somehow enforce this (but maybe it is not a priority.
> 
> Indeed, let's file an issue with a good description then. I think we don't care for this PR.

Yes, let's do that. I am not sure what you described worked by the way. With:
```
val t_T (v_Self v_U: Type0) : Type0

[@@ FStar.Tactics.Typeclasses.tcinstance]
val impl:t_T u8 u8
```
fstar complains with:
```
Instances must define instances of `class` types.
  - Type
    Opaque_impl.t_T Rust_primitives.Integers.u8 Rust_primitives.Integers.u8
    is not a class.
```

W95Psp:
Ah, you are right, there's a catch, you need an attribute to mark the `val` (or `assume val`) as a typeclass, as follows:
```fstar
[@@ FStar.Tactics.Typeclasses.tcclass]
val t_T (v_Self v_U: Type0) : Type0

[@@ FStar.Tactics.Typeclasses.tcinstance]
val impl:t_T u8 u8
```

maximebuyse:
> ```rust
> #[hax_lib::opaque]
> fn opaque_function(){}
> 
> fn regular_function(){}
> ```
> 
> [Open this code snippet in the playground](https://hax-playground.cryspen.com/#fstar/d5bf025e17/gist=2c8433ec799e218e8f4a9ae3db084cff)
> 
> Gets translated to:
> 
> ```fstar
> val opaque_function: Prims.unit -> Prims.unit
> assume val opaque_function': Prims.unit -> Prims.unit
> let opaque_function = opaque_function'
> 
> let regular_function (_: Prims.unit): Prims.unit = ()
> ```
> 
> I don't think we want the `val opaque_function`, do we?

@W95Psp This is fixed now

maximebuyse:
We just noticed a few problems here:
- For the case of types we should generate something similar as other cases in the fst (`assume val` with a let alias)
- The let alias should take and pass the generic arguments otherwise fstar fails. 
- We should have tests for all of that, also with trait bounds.

maximebuyse:
> We just noticed a few problems here:
> 
>     * For the case of types we should generate something similar as other cases in the fst (`assume val` with a let alias)
> 
>     * The let alias should take and pass the generic arguments otherwise fstar fails.
> 
>     * We should have tests for all of that, also with trait bounds.

All of this is now fixed. @W95Psp You can review.

W95Psp:
I've spawned a ML-KEM job for this PR: https://github.com/cryspen/libcrux/actions/runs/12292989138

That looks good to me. I think we can improve a bit the tests. Can you add:
 - an associated type in trait `T`
 - an associated constant in `T`
 - an associated method (with a self) in `T`, with a precondition (say, a method that takes a self and a u8, with a precondiotion on the u8)
 - in the impl of T for u8, add the same precondition and a postcondition
 - an inherent impl

This way I think we will have all the possible items kind covered :)

W95Psp:
The Ci above failed: after chatting with @maximebuyse, we found out the issue. This PR is a breaking change for haxlib, thus we need to update the pin in  libcrux.

W95Psp:
We found a small regression: when extracting an impl without associated type in F* interface mode, we used to expose a single opaque `val` for the impl. This is not the case any more, but @maximebuyse is taking care of fixing that.
After that, we will re-trigger a libcrux job, if this job is green, we merge.

maximebuyse:
> We found a small regression: when extracting an impl without associated type in F* interface mode, we used to expose a single opaque `val` for the impl. This is not the case any more, but @maximebuyse is taking care of fixing that. After that, we will re-trigger a libcrux job, if this job is green, we merge.



> We found a small regression: when extracting an impl without associated type in F* interface mode, we used to expose a single opaque `val` for the impl. This is not the case any more, but @maximebuyse is taking care of fixing that. After that, we will re-trigger a libcrux job, if this job is green, we merge.

I worked on a fix, I think it does what we want. It would be nice to add a test for that but it affects the interface mode (not interface-only) and I don't think we have a test that does that. Anyway I restarted the libcrux job: https://github.com/cryspen/libcrux/actions/runs/12296788479/job/34316515254

W95Psp (approved):
Libcrux CI is green, let's merge!

PR #1140, opened 2024-11-26
fix(engine): item ordering

### LINKED ISSUE #771: Engine: dependencies: the sorting seems unstable
See https://github.com/cryspen/libcrux/blob/ml-kem-lax-check/libcrux-ml-kem/proofs/fstar/extraction/Libcrux_ml_kem.Vector.Portable.Serialize.fst

In this module, functions are in big part independent, but still, they get reordered.

W95Psp:
This PR fixes #771: now the items are sorted topologically *but* kept in definition order if equal topologically speaking.

Technically this PR:
 - kills the `original_items` parameter in `dependencies.ml` functions
 - introduce a functor `WithItems` instead, that takes a list of items
 - add a module `Ordered_concrete_ident` that defines a comparison operator sensible to the list of items taken by `WithItems`
 - apply the functor automatically to provide the same interface as we had before


maximebuyse (approved):
Great :tada:

PR #1146, opened 2024-11-27
Improve `setup.sh`

### LINKED ISSUE #1145: Build error: binaries seems cached to eagerly
@franziskuskiefer reporting the following error after updating a local checkout of hax and running `./setup.sh`:

```
Caused by:
  process didn't exit successfully: `/Users/franziskus/cryspen/workbench/hax/target/release/build/hax-engine-names-extract-585571f63244d874/build-script-build` (exit status: 101)
  --- stderr
      Finished `dev` profile [unoptimized + debuginfo] target(s) in 2.51s
  thread 'main' panicked at /Users/franziskus/cryspen/workbench/hax/hax-types/src/driver_api.rs:52:72:
  called `Result::unwrap()` on an `Err` value: Message("unknown variant `Abi`, expected one of `Rust`, `C`, `Other`")
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

  thread 'main' panicked at engine/names/extract/build.rs:113:19:
  error while parsing JSON: EOF while parsing a value at line 1 column 0

  String was:
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This seems to be a problem of caching: the build script of the crate `names` is parsing JSON using the schemas defined post https://github.com/hacspec/hax/pull/1122, which adds the type `Abi`, while the JSON received appears to have no such `Abi` enum.

W95Psp:
This PR:
 - fixes #1145 by cleaning both cargo and dune before doing anything
 - adds a `--no-cleanup` flag to disable this cleaning behavior if wanted
 - adds a `--help` flag and a help message

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -5,6 +5,7 @@ set -eu
 SCRIPTPATH="$( cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
 
 opam_jobs=4
+CLEANUP_WORKSPACE=on
 
 # Parse command line arguments.
 all_args=("$@")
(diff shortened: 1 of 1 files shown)


karthikbhargavan:
hm, this is going to be hard to test since I have now cleaned up my state. shall we find someone with a dirty state?

W95Psp:
Yes, I think that's a bit complicated to test :sweat_smile:
Let's merge then, thanks for the fast review :)

PR #1148, opened 2024-11-27
Book: small cleanup

W95Psp:
This PR cleans up the book a little, essentially removing greyed-out sections.

As we discussed @karthikbhargavan, see https://github.com/cryspen/home/issues/112.


karthikbhargavan (approved):
The text could likely use some polish, but let's do that as part of the paper writing and copy the final text over here.

PR #1166, opened 2024-12-03
Clean up of F* libraries

karthikbhargavan:
This PR adds some missing functions and removes all assumes and admits from the F* library.


karthikbhargavan:
We need to test this on all our projects before merging.

W95Psp (approved):
The code looks good to me, but we still need to make sure that's fine with ML-KEM.

PR #1174, opened 2024-12-09
Remove rejected type ascription patterns from generated f*.

### LINKED ISSUE #1170: Wrong type ascription pattern in generated F*
```rust
struct Other<'a>(&'a i32);

enum Test<'a> {
    C1(Other<'a>)
}

impl<'a> Test<'a> {
    fn test(&self) -> i32 {
        match self {
            Self::C1(c) => *c.0

        }
    }
}
```
[Open this code snippet in the playground](https://hax-playground.cryspen.com/#fstar/de59826b83/gist=b1f6214c71fbe3b0a3dbacf28b9f5b77)

The generated F* for the `test` method from the example is: 
```
let impl__test (self: t_Test) : i32 =
  match self with
  | (Test_C1 c: t_Test) -> c._0
```
F* rejects it with `Type ascriptions within patterns are only allowed on variables`

maximebuyse:
Fixes #1170 

This mostly ignores type ascription patterns in the F* backend (maybe we should completely ignore them, see https://github.com/hacspec/hax/issues/1170#issuecomment-2527433453). These patterns are translated as their subpattern. This fixes the issue from #1170 

The main difference as a side effect is that for 

```
let _: i8 = 1 + 1;
```
We will now generate
```
let _:i8 = 1y +! 1y in
```
instead of 
```
let (_: i8):i8 = 1y +! 1y in
```
Both are valid F* and the new version is simpler, and clearer.

### DIFF
--- a/engine/backends/fstar/fstar_backend.ml
+++ b/engine/backends/fstar/fstar_backend.ml
@@ -401,8 +401,9 @@ struct
     let ppat = ppat' false in
     match p.p with
     | PWild -> F.wild
-    | PAscription { typ; pat } ->
+    | PAscription { typ; pat = { p = PBinding _; _ } as pat } ->
         F.pat @@ F.AST.PatAscribed (ppat pat, (pty p.span typ, None))
+    | PAscription { pat; _ } -> ppat pat
     | PBinding
         {
           mut = Immutable;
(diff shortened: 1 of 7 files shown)


W95Psp (approved):
Thanks, that looks good to me!

PR #1184, opened 2024-12-12
Skip generated impls using `Erased` attribute.

### LINKED ISSUE #1070: Cyclic dependency introduced by bundling with `let open`
```rust
pub mod a {
    #[derive(Debug)]
    pub struct S {}
    pub struct Sa(super::b::Sb);
    impl std::fmt::Debug for Sa {
        fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
            write!(f, "{:?}", self.0)
        }
    }

}

pub mod b {
    pub struct Sb(super::a::S);
    impl std::fmt::Debug for Sb {
        fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
            write!(f, "{:?}", self.0)
        }
    }
}
```
[Open this code snippet in the playground](https://hax-playground.cryspen.com/#fstar/001a27e207/gist=5c43e496c5fdf2dd5a8f47540b2ba269)

In the situation above, the recursive bundle contains a `let open` for one of the original modules creating a cycle.

maximebuyse:
Fixes #1070 

Instead of just skipping generated impls, this adds the `Erased` attribute to them so that the body is dropped but they still exist in the engine. This allows for them to be moved by bundling thus removing the spurious `let open`.

### DIFF
--- a/engine/backends/coq/coq/coq_backend.ml
+++ b/engine/backends/coq/coq/coq_backend.ml
@@ -407,14 +407,19 @@ struct
       method item'_Impl ~super ~generics ~self_ty ~of_trait ~items
           ~parent_bounds:_ ~safety:_ =
         let name, args = of_trait#v in
-        CoqNotation.instance
-          (name#p ^^ string "_" ^^ string (Int.to_string ([%hash: item] super)))
-          generics#p []
-          (name#p ^^ concat_map (fun x -> space ^^ parens x#p) args)
-          (braces
-             (nest 2
-                (concat_map (fun x -> break 1 ^^ name#p ^^ !^"_" ^^ x#p) items)
-             ^^ break 1))
+        if Attrs.is_erased super.attrs then empty
+        else
+          CoqNotation.instance
+            (name#p ^^ string "_"
+            ^^ string (Int.to_string ([%hash: item] super)))
+            generics#p []
+            (name#p ^^ concat_map (fun x -> space ^^ parens x#p) args)
+            (braces
+               (nest 2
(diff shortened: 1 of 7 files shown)


maximebuyse:
Two comments here:
- the `skip` attribute seems to have been removed so maybe we can remove the corresponding logic in `import_thir`?
- there is a coq test that fails because of this (it generates a type class instance with dummy methods), maybe this means we should also implement the `Erased` attribute in the other backends.

maximebuyse:
> Two comments here:
> 
>     * the `skip` attribute seems to have been removed so maybe we can remove the corresponding logic in `import_thir`?
> 
>     * there is a coq test that fails because of this (it generates a type class instance with dummy methods), maybe this means we should also implement the `Erased` attribute in the other backends.

For coq, I added a quick fix: it filters out trait impls that have the `Erased` attribute. I opened https://github.com/hacspec/hax/issues/1185 to add full support for `Erased` in the coq backend.

maximebuyse:
> Two comments here:
> 
>     * the `skip` attribute seems to have been removed so maybe we can remove the corresponding logic in `import_thir`?
> 
>     * there is a coq test that fails because of this (it generates a type class instance with dummy methods), maybe this means we should also implement the `Erased` attribute in the other backends.

For `skip` we discussed it and it is indeed unused so I removed the logic. @W95Psp You can finish your review, I am done with my changes.

W95Psp (approved):
Looks great, thanks!

PR #1189, opened 2024-12-16
fix(engine) Opaque consts

### LINKED ISSUE #1131: Engine: F*: interface mode: `dropped_body` on associated constants of inherent impls
Consider a struct `X` and an inherent `impl` block for `X`, as follows:

```rust
struct X;
impl X {
    const A: u8 = 0;
}
```

`A` is an associated constant for `X`.

When extracted with interface flags on to F*, this results in a `let` binding with a `dropped_body ()` body.

**Why this is not normal:**
The current policy is that constant items that are marked "extract type-only" are actually fully translated: extracting a constant almost never yields issues, and is almost always useful in proof settings.

**Plausible explanation for this bug:**  
Here, this policy doesn't apply probably because associated constants in inherent impls are translated with `c_expr` instead of `c_body` in `thir_export.ml`.

### LINKED ISSUE #1152: Spuritous dropped body in associated constants in inherent impls
We need a reproducer here.

maximebuyse:
Fixes #1131. Fixes #1152.
This is a followup of #1134. It fixes 2 problems:
- `const`s inside (non-trait) impls are not erased by type-only flag (so that the behaviour is the same as for top-level `const`s)
- For erased `const`s (when an `opaque` flag is added by the user) we now produce a `val` in the `fsti` (it was missing before).

### DIFF
--- a/engine/backends/fstar/fstar_backend.ml
+++ b/engine/backends/fstar/fstar_backend.ml
@@ -929,10 +929,11 @@ struct
           F.decl ~fsti:false
           @@ F.AST.TopLevelLet (qualifier, [ (pat, pexpr body) ])
         in
-        let interface_mode = ctx.interface_mode && not (List.is_empty params) in
+        let is_const = List.is_empty params in
         let ty =
-          add_clauses_effect_type ~no_tot_abbrev:interface_mode e.attrs
-            (pty body.span body.typ)
+          add_clauses_effect_type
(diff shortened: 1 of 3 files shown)


W95Psp (approved):
LGTM. Let's merge if https://github.com/cryspen/libcrux/actions/runs/12357942046 is green! :)

W95Psp:
It's green, I'm pressing the button!

PR #1193, opened 2024-12-17
fix(frontend) Only include direct parent bounds for associated types in traits.

### LINKED ISSUE #1159: Super-parent bounds of associated types in impls are inlined
```rust
trait A{}
trait B{}
trait C:A+B{}

trait D {
   type T:C;
}

```
[Open this code snippet in the playground](https://hax-playground.cryspen.com/#fstar/102ae9fdc9/gist=92dc559a6c2d45b8d440f4dc2aa977ed)

Here we translate to the following F*:
```fstar

class t_C (v_Self: Type0) = {
  [@@@ FStar.Tactics.Typeclasses.no_method]_super_14273389043514592625:t_A
  v_Self;
  [@@@ FStar.Tactics.Typeclasses.no_method]_super_14858733883632685556:t_B
  v_Self
}

class t_D (v_Self: Type0) = {
  f_T:Type0;
  f_T_16407784479303782286:t_C f_T;
  f_T_4589724874609156387:t_B f_T;
  f_T_3674199358433564065:t_A f_T
}
```

The bounds on `T` contain the super bounds of `C`, this is wrong.

maximebuyse:
Fixes #1159 

In the case of an associated type inside a trait that has indirect bounds, we don't include the indirect bounds anymore (only the direct ones). This behaves better with backends and the information can still be reconstructed.


frontend/exporter/src/traits.rs
```diff
@@ -211,7 +211,6 @@ fn solve_item_traits_inner<'tcx, S: UnderOwnerState<'tcx>>(
     use crate::rustc_middle::ty::ToPolyTraitRef;
     let tcx = s.base().tcx;
     let param_env = s.param_env();
-
```
W95Psp, line 214: Can you revert that line change? so that we get nice history :)

W95Psp (approved):
LGTM, just on small thing

PR #1197, opened 2024-12-18
Engine: get rid of most of exception throwing function calls, improve parsers and error reports

### LINKED ISSUE #1187: F* Backend: catch Failure and print readable error message
The F* backend has a number of calls to "failwith" notably when parsing things as F* but also in other expected things.
These result in the the hax invocation failing with a long and somewhat unreadable backtrace with no indication of the "span" in which this error was triggered. We should systematically catch these "Failure" exceptions in the backend and print nicer messages with the span of the source code.

W95Psp:
This PR:
 - adds a version to `haxmeta` files
 - adds a version to the engine
 - makes the frontend check haxmeta versions
 - makes the engine check the frontend version
 - improve parsing error reporting
 - makes generated parsers conform to yojson style
 - report parsing errors in the F* backend
 - removes most of `*_exn` calls
 - removes most of failwith calls

Fixes https://github.com/hacspec/hax/issues/1150, fixes https://github.com/hacspec/hax/issues/1188, fixes #1187

Context: this PR goes toward better user experience for releasing hax v0.1, it gets rid of most of remaining unhandled explicit failures

### DIFF
--- a/engine/lib/attr_payloads.ml
+++ b/engine/lib/attr_payloads.ml
@@ -6,7 +6,18 @@ let payloads : attrs -> (Types.ha_payload * span) list =
   let parse =
     (* we have to parse ["JSON"]: first a string, then a ha_payload *)
     function
-    | `String s -> Yojson.Safe.from_string s |> [%of_yojson: Types.ha_payload]
+    | `String s -> (
+        match Yojson.Safe.from_string s |> Types.safe_ha_payload_of_yojson with
+        | Error _ ->
+            Stdlib.prerr_endline
+              [%string
+                {|
+The hax engine could not parse a hax attribute.
+This means that the crate being extracted and the version of hax engine are incompatible.
+Please make sure the `hax-lib` dependency of the extracted crate matches hax-engine's version (%{Types.hax_version}).
+|}];
+            Stdlib.exit 1
+        | Ok value -> value)
     | x ->
         Stdlib.failwith
         @@ "Attr_payloads: payloads: expected a string while parsing JSON, got "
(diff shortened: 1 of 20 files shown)


maximebuyse (approved):
Great :tada:

engine/lib/attr_payloads.ml
```diff
@@ -6,7 +6,18 @@ let payloads : attrs -> (Types.ha_payload * span) list =
...
     (* we have to parse ["JSON"]: first a string, then a ha_payload *)
     function
-    | `String s -> Yojson.Safe.from_string s |> Types.parse_ha_payload
+    | `String s -> (
+        match Yojson.Safe.from_string s |> Types.safe_ha_payload_of_yojson with
+        | Error _ ->
+            Stdlib.prerr_endline
+              [%string
+                {|
+The hax engine could not parse a hax attribute.
```
karthikbhargavan, line 15: Couldn't this also happen if the hax_lib is right but somehow the content of the attibute is malformed?

W95Psp: No, this situation should never happen, the hax lib attributes are always JSON serializations of a Rust enum, and the OCaml parser used here is directly generated from the Rust enum definition.

So, we could still have a bug in generated parsers, but this almost never happened (it happened once because of OCaml ints being smaller than Rust ints)

karthikbhargavan (approved):
Looks great. Did you try on libcrux? If not, I can tomorrow.

W95Psp:
The job is successful, I'm mergining

PR #1205, opened 2024-12-19
Bundle naming fixes

### LINKED ISSUE #1157: clash of impl names in rec bundles (disambiguators are no longer valid)
impls coming from different modules in a recursive bundles don't get a new disambiguator so the names can clash.

### LINKED ISSUE #1200: Stable name for rec bundles
Rec bundles have names that contain a hash. This is problematic because if the content of the bundle changes the name of the generated file also changes which makes it hard to manage projects. We should instead generate a name using the names of the modules that compose the bundle, with a disambiguation only if needed.

maximebuyse:
This PR:
- Closes #1200. Makes bundle names stable, by computing the hash based on the names of the modules going inside the bundle (instead of the items)
- Fixes #1157. By adding a disambiguator when names clash, instead of modifying the last string in the path (which fails for impls that don't have a string at the en of their paths)

### DIFF
--- a/engine/lib/concrete_ident/concrete_ident.ml
+++ b/engine/lib/concrete_ident/concrete_ident.ml
@@ -596,14 +596,13 @@ module Create = struct
     let len x = List.length x.def_id.path in
     let compare x y = len x - len y in
     let id = List.min_elt ~compare from |> Option.value_exn in
-    let parent = parent id in
     {
       kind = Kind.Value;
       def_id =
         {
-          parent.def_id with
+          id.def_id with
           path =
-            parent.def_id.path
+            id.def_id.path
             @ [
                 {
                   data = TypeNs "rec_bundle";
--- a/engine/lib/concrete_ident/concrete_ident.ml
+++ b/engine/lib/concrete_ident/concrete_ident.ml
@@ -605,7 +605,7 @@ module Create = struct
             id.def_id.path
             @ [
                 {
-                  data = TypeNs "rec_bundle";
+                  data = TypeNs "cyclic_bundle";
                   disambiguator = [%hash: t list] from;
                 };
               ];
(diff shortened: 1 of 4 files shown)


engine/lib/concrete_ident/concrete_ident.ml
```diff
@@ -635,6 +634,14 @@ module Create = struct
   let constructor name =
     let path = name.def_id.path @ [ { data = Ctor; disambiguator = 0 } ] in
     { name with def_id = { name.def_id with path } }
+
+  let add_disambiguator name disambiguator =
+    let path = name.def_id.path in
+    let last = List.last_exn path in
+    let path =
+      List.drop_last_exn path @ [ { data = last.data; disambiguator } ]
```
karthikbhargavan, line 642: Shopuld we be using an "_exn" function here? If needed, let's put a comment saying why this cannot fail.

karthikbhargavan (commented):
Generally looks good, but two comments, since we are touching this code anyway.

First, can we rename "Rec_bundle" to "Cyclic_bundle"? This would be more precise.
Second, could we list the files that are part of the bundle in the top of the file somehow?

I can imagine the second ask to be harder.

maximebuyse:
libcrux CI job: https://github.com/cryspen/libcrux/actions/runs/12480998112/job/34832756469

engine/lib/concrete_ident/concrete_ident.ml
maximebuyse: Done

maximebuyse:
> Generally looks good, but two comments, since we are touching this code anyway.
> 
> First, can we rename "Rec_bundle" to "Cyclic_bundle"? This would be more precise. Second, could we list the files that are part of the bundle in the top of the file somehow?
> 
> I can imagine the second ask to be harder.

Done for the first one. The list of files (or maybe modules) is a good idea, but let's leave it for later. I think trying to name the bundle with a name containing the names of the modules it contains (which is part of the naming refactor) would already help.

karthikbhargavan (approved):
Looks good, let's merge if libcrux CI passes.

PR #1206, opened 2024-12-19
fix(engine) Make sub-parts of `Quote` visited by visitors

### LINKED ISSUE #1196: No rewriting of `$` names in `fstar!` macro with bundling
```rust
mod a {
    #[hax_lib::requires(fstar!("$super::b::cond()"))]
    pub fn f(){super::b::g()}
}
mod b {
    pub fn g(){super::a::f()}
    pub fn cond() -> bool {true}
}

```
[Open this code snippet in the playground](https://hax-playground.cryspen.com/#fstar/a598467a5d/gist=fb5e07a4021ac7e68906a67472630599)

The example above has the following code in its fstar bundle: 
```
let cond (_: Prims.unit) : bool = true

let rec f (_: Prims.unit)
    : Prims.Pure Prims.unit
      (requires Playground.B.cond ())
      (fun _ -> Prims.l_True) = g ()
```

Here we find `Playground.B.cond` in the precondition which should be transformed to just `cond` to avoid a cyclic dependency. I suspect the macro does something at the string level which will make it hard to fix.

maximebuyse:
Fixes #1196 

The investigation for #1196 revealed that subparts of `Quote` variant in the AST were ignored by visitors. The easiest way to change that is to use a `quote_contents` ADT inside instead of an inlined polymorphic variants. This involves quite a few changes here and there.

### DIFF
--- a/engine/backends/coq/coq/coq_backend.ml
+++ b/engine/backends/coq/coq/coq_backend.ml
@@ -151,6 +151,16 @@ struct
       method borrow_kind_Unique = default_document_for "borrow_kind_Unique"
       method common_array x1 = brackets (separate (semi ^^ space) x1)
 
+      method quote_content_Verbatim _v =
+        default_document_for "quote_content_Verbatim"
+
+      method quote_content_Expr _e = default_document_for "quote_content_Expr"
+
+      method quote_content_Pattern _p =
+        default_document_for "quote_content_Pattern"
+
+      method quote_content_Typ _t = default_document_for "quote_content_Typ"
+
       method dyn_trait_goal ~trait:_ ~non_self_args:_ =
         default_document_for "dyn_trait_goal"
 
--- a/engine/backends/coq/coq/coq_backend.ml
+++ b/engine/backends/coq/coq/coq_backend.ml
@@ -151,16 +151,6 @@ struct
       method borrow_kind_Unique = default_document_for "borrow_kind_Unique"
       method common_array x1 = brackets (separate (semi ^^ space) x1)
 
-      method quote_content_Verbatim _v =
-        default_document_for "quote_content_Verbatim"
-
-      method quote_content_Expr _e = default_document_for "quote_content_Expr"
-
-      method quote_content_Pattern _p =
-        default_document_for "quote_content_Pattern"
-
-      method quote_content_Typ _t = default_document_for "quote_content_Typ"
-
       method dyn_trait_goal ~trait:_ ~non_self_args:_ =
         default_document_for "dyn_trait_goal"
 
--- a/engine/backends/fstar/fstar_backend.ml
+++ b/engine/backends/fstar/fstar_backend.ml
@@ -692,10 +692,10 @@ struct
   and pquote span { contents; _ } =
     List.map
       ~f:(function
-        | `Verbatim code -> code
-        | `Expr e -> pexpr e |> term_to_string
-        | `Pat p -> ppat p |> pat_to_string
-        | `Typ p -> pty span p |> term_to_string)
+        | Verbatim code -> code
+        | Expr e -> pexpr e |> term_to_string
+        | Pattern p -> ppat p |> pat_to_string
+        | Typ p -> pty span p |> term_to_string)
       contents
     |> String.concat
 
--- a/engine/lib/generic_printer/generic_printer.ml
+++ b/engine/lib/generic_printer/generic_printer.ml
@@ -231,14 +231,15 @@ module Make (F : Features.T) = struct
       [local] is true, otherwise it prints the full path, separated by
       `module_path_separator`. *)
 
-      method quote (quote : quote) : document =
+      method quote ~contents ~witness:_ : document =
         List.map
-          ~f:(function
-            | `Verbatim code -> string code
-            | `Expr e -> self#print_expr AstPosition_Quote e
-            | `Pat p -> self#print_pat AstPosition_Quote p
-            | `Typ p -> self#print_ty AstPosition_Quote p)
-          quote.contents
+          ~f:(fun doc ->
+            match doc#v with
+            | Verbatim code -> string code
+            | Expr e -> self#print_expr AstPosition_Quote e
+            | Pattern p -> self#print_pat AstPosition_Quote p
+            | Typ t -> self#print_ty AstPosition_Quote t)
+          contents
         |> concat
 
       (** {2:specialize-expr Specialized printers for [expr]} *)
@@ -628,10 +629,6 @@ module Make (F : Features.T) = struct
                   ^ [%show: global_ident] id
                   ^ "]"))
           ast_position id
-
-      method _do_not_override_lazy_of_quote ast_position (value : quote)
-          : quote lazy_doc =
-        lazy_doc (fun (value : quote) -> self#quote value) ast_position value
 
       method! _do_not_override_lazy_of_item ast_position (value : item)
           : item lazy_doc =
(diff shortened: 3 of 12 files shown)


maximebuyse:
I succeeded in making a version that fixes the bug for F*. But I am not sure that it does the right thing for the generic printer (at least it builds correctly and doesn't break tests).

engine/backends/coq/coq/coq_backend.ml
```diff
@@ -151,6 +151,16 @@ struct
       method borrow_kind_Unique = default_document_for "borrow_kind_Unique"
       method common_array x1 = brackets (separate (semi ^^ space) x1)
 
+      method quote_content_Verbatim _v =
```
karthikbhargavan, line 154: This may need a review from @cmester0

engine/backends/fstar/fstar_backend.ml
```diff
@@ -692,10 +692,10 @@ struct
...
     List.map
       ~f:(function
-        | `Verbatim code -> code
-        | `Expr e -> pexpr e |> term_to_string
-        | `Pat p -> ppat p |> pat_to_string
-        | `Typ p -> pty span p |> term_to_string)
+        | Verbatim code -> code
+        | Expr e -> pexpr e |> term_to_string
+        | Pattern p -> ppat p |> pat_to_string
+        | Typ p -> pty span p |> term_to_string)
```
karthikbhargavan, line 698: Generally, the use of explicit constructors is better so this change seem fine, unless we were relying on subtyping or on one of these constructor names also being used in another type. As long as the code compiles and typechecks this should not be the case.

engine/lib/generic_printer/generic_printer.ml
```diff
@@ -231,14 +231,15 @@ module Make (F : Features.T) = struct
       [local] is true, otherwise it prints the full path, separated by
       `module_path_separator`. *)
 
-      method quote (quote : quote) : document =
+      method quote ~contents ~witness:_ : document =
```
karthikbhargavan, line 234: why not take an argument of type `quote` here, which has both the contents and witness?

engine/lib/generic_printer/generic_printer_template.ml
```diff
@@ -35,6 +35,11 @@ struct
       method borrow_kind_Mut _x1 = default_document_for "borrow_kind_Mut"
       method borrow_kind_Shared = default_document_for "borrow_kind_Shared"
       method borrow_kind_Unique = default_document_for "borrow_kind_Unique"
+      method cf_kind_BreakOnly = default_document_for "cf_kind_BreakOnly"
+
+      method cf_kind_BreakOrReturn =
+        default_document_for "cf_kind_BreakOrReturn"
+
```
karthikbhargavan, line 42: Is this a drive-by fix, how is it related to Quote?

engine/lib/phases/phase_transform_hax_lib_inline.ml
```diff
@@ -134,33 +133,22 @@ module%inlined_contents Make (F : Features.T) = struct
...
           let verbatim = split_str ~on:"SPLIT_QUOTE" str in
           let contents =
-            let rec f verbatim
-                (code :
-                  [ `Verbatim of string
-                  | `Expr of B.expr
-                  | `Pat of B.pat
-                  | `Typ of B.ty ]
-                  list) =
+            let rec f verbatim (code : B.quote_content list) =
```
karthikbhargavan, line 141: Here, your code looks nicer than what was there before.

test-harness/src/snapshots/toolchain__loops into-fstar.snap
```diff
@@ -651,7 +651,7 @@ let enumerated_chunked_slice (#v_T: Type0) (slice: t_Slice v_T) : u64 =
     (fun count i ->
         let count:u64 = count in
         let i:usize = i in
-        i <= Core.Slice.impl__len #v_T slice)
+        i <= Core.Slice.impl__len #v_T slice <: usize)
```
karthikbhargavan, line 654: This diff seems strange, but not wrong?

karthikbhargavan (approved):
Generally, this looks good to me. Let's test on libcrux, and maybe you can resolve some of my comments.

engine/lib/generic_printer/generic_printer_template.ml
maximebuyse: This is generated code. I used the tool to regenerate this and this was part of it. This is because we changed the AST as part of https://github.com/hacspec/hax/pull/988, and this file hasn't been regenerated since.

engine/lib/generic_printer/generic_printer.ml
maximebuyse: My understanding is that the method signature is generated from the AST so we don't get to choose. But maybe I am wrong and there is a better way. @W95Psp will know.

test-harness/src/snapshots/toolchain__loops into-fstar.snap
W95Psp: I think that makes sense: https://github.com/hacspec/hax/blob/f47b2ae440333850f9990c59f7200e79802b64f5/tests/loops/src/lib.rs#L26 introduces a quote.
The contents of the quote were previously not visited by any visitor, now they are, and the subtyping insertion is from a visitor.

engine/backends/coq/coq/coq_backend.ml
```diff
@@ -151,6 +151,16 @@ struct
...
+        default_document_for "quote_content_Verbatim"
+
+      method quote_content_Expr _e = default_document_for "quote_content_Expr"
+
+      method quote_content_Pattern _p =
+        default_document_for "quote_content_Pattern"
+
+      method quote_content_Typ _t = default_document_for "quote_content_Typ"
+
```
W95Psp, line 163: Those methods should be defined in `generic_printer` I think, with good defaults:
 - verbatim just prints the string verbatim
 - `quote_content_* value` should be `value#p`

engine/lib/generic_printer/generic_printer.ml
```diff
@@ -231,14 +231,15 @@ module Make (F : Features.T) = struct
...
-            | `Expr e -> self#print_expr AstPosition_Quote e
-            | `Pat p -> self#print_pat AstPosition_Quote p
-            | `Typ p -> self#print_ty AstPosition_Quote p)
-          quote.contents
+          ~f:(fun doc ->
+            match doc#v with
+            | Verbatim code -> string code
+            | Expr e -> self#print_expr AstPosition_Quote e
+            | Pattern p -> self#print_pat AstPosition_Quote p
+            | Typ t -> self#print_ty AstPosition_Quote t)
```
W95Psp, line 241: Here, we should directly use `doc#p`: this will call into the underlying methods `self#quote_content_{Verbatim,Expr,Pattern,Typ}`

engine/lib/generic_printer/generic_printer_template.ml
```diff
@@ -318,6 +319,16 @@ struct
...
+      method quote_content_Expr _x1 = default_document_for "quote_content_Expr"
+
+      method quote_content_Pattern _x1 =
+        default_document_for "quote_content_Pattern"
+
+      method quote_content_Typ _x1 = default_document_for "quote_content_Typ"
+
+      method quote_content_Verbatim _x1 =
+        default_document_for "quote_content_Verbatim"
+
```
W95Psp, line 331: Related to my [other comment in `coq/coq_backend.ml`](https://github.com/hacspec/hax/pull/1206/files#r1903775169):

Those will disappear after implementing default methods in generic_printer.ml, don't forget to regenerate at that point!

engine/backends/coq/coq/coq_backend.ml
maximebuyse: This is gone now that we have default methods for this.

maximebuyse:
> Thanks, there's a few things to change, but that looks good :)

Thanks, everything should be good now!

W95Psp (approved):
LGTM, let's merge!

PR #1208, opened 2024-12-23
fix(engine) Add ascription to match scrutinees.

### LINKED ISSUE #1207: Missing type ascription in match scrutinee with array indexing
While I would expect it to, the following snippet does not typecheck:

```rust
const FOO_SIZE: usize = 5;

const FOO: [u8; FOO_SIZE] = [0, 0, 0, 0, 0];

#[hax_lib::requires(i < FOO_SIZE)]
fn foo(i: usize) {
    let _ = match FOO[i] {
        0xff => 42,
        _ => 41,
    };
}
```

[view on playground](https://hax-playground.cryspen.com/#fstar/9d78af6a1a/gist=83aa686febdca1aa1d35bdb7b8e7788c)

However, when replacing the `match` with an `if`, it does, as expected:

```rust
const FOO_SIZE: usize = 5;

const FOO: [u8; FOO_SIZE] = [0, 0, 0, 0, 0];

#[hax_lib::requires(i < FOO_SIZE)]
fn foo(i: usize) {
    let _ = if FOO[i] == 0xff {
        42
    } else {
        41
    };
}
```

[view on playground](https://hax-playground.cryspen.com/#fstar/9d78af6a1a/gist=628b4d1a9aa9414f1f6d8fdf40c870ee)

It seems that the array indexing in the first case is missing a `<: u8` type ascription, which gets generated only in the `if` case.

Is this some specific interaction between the `match` and array indexing, or just the symptom of something else?

Credits go to @chrysn for finding this.

maximebuyse:
Fixes #1207 

This adds an ascription to pattern matching scrutinees. Maybe this could be made more fine-grained to do it only with scrutinees that can cause issues (this would avoid making the generated code too verbose)

### DIFF
--- a/engine/lib/ast_utils.ml
+++ b/engine/lib/ast_utils.ml
@@ -367,6 +367,10 @@ module Make (F : Features.T) = struct
                           bounds_impls;
                         };
                   }
+            (* Match scrutinees need to be ascribed as well
+               (see https://github.com/hacspec/hax/issues/1207).*)
+            | Match { scrutinee; arms } ->
+                { e with e = Match { scrutinee = ascribe scrutinee; arms } }
             | _ ->
                 (* Ascribe the return type of a function application & constructors *)
                 if (ascribe_app && is_app e.e) || [%matches? Construct _] e.e
--- a/test-harness/src/snapshots/toolchain__cyclic-modules into-fstar.snap
+++ b/test-harness/src/snapshots/toolchain__cyclic-modules into-fstar.snap
@@ -254,22 +254,22 @@ type t_T =
   | T_t2 : t_T
 
 let t_T_cast_to_repr (x: t_T) : isize =
-  match x with
+  match x <: t_T with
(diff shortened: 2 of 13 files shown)


engine/lib/ast_utils.ml
```diff
@@ -367,6 +367,10 @@ module Make (F : Features.T) = struct
                           bounds_impls;
                         };
                   }
+            (* Match scrutinees need to be ascribed as well
+               (see https://github.com/hacspec/hax/issues/1207).*)
+            | Match { scrutinee; arms } ->
```
karthikbhargavan, line 372: Nice, minimal change.

karthikbhargavan (approved):
Great, thanks. Let's get this in and close the issue.

maximebuyse:
I started a libcrux job to test this branch: https://github.com/cryspen/libcrux/actions/runs/12480045327/job/34830201797

PR #1209, opened 2024-12-23
fix(engine) Check attributes of descendants to build dependency graph.

### LINKED ISSUE #1016: F* term ordering issue with hax attribute
It seems to me that F* terms are not issued in the right order in some cases involving hax attributes.

For instance, the following Rust snippet

```rust
const FOO_LEN: usize = 42;

#[hax_lib::attributes]
struct Foo {
    #[refine(foo < FOO_LEN)]
    foo: usize,
}
```

[view in hax playground](https://hax-playground.cryspen.com/#fstar/e793beee0a/gist=19113ac424b694d69c1fded7e405a2a2)

results in the following F*:

```fstar
type t_Foo = {
  f_foo:f_foo: usize{f_foo <. v_FOO_LEN}
}

let v_FOO_LEN: usize = sz 42
```

As the declaration of `FOO_LEN` gets issued after its use, lax checking fails with the following error:

```
* Error 72 at Playground.fst(7,30-7,39):
  - Identifier not found: [v_FOO_LEN]
```

My understanding is that there is already a dependency detection mechanism inside the engine (in [engine/lib/dependencies.ml](https://github.com/hacspec/hax/blob/e793beee0acf5a573ef9820c1e473ebd4a69c06b/engine/lib/dependencies.ml) I think) where dependencies between items are modeled as graph vertices, which then allows backends to issue terms in topological order (I believe some of it happens in [engine/backends/fstar/fstar_backend.ml](https://github.com/hacspec/hax/blob/main/engine/backends/fstar/fstar_backend.ml) but I am not sure where/how these vertices are inserted into the graph).

**Is my understanding correct? Could you quickly explain where these vertices are inserted?**

I suppose that dependencies resulting from hax attributes are not currently modeled in that graph, hence why these terms do not get issued in the correct order.

maximebuyse:
Fixes #1016 

When creating the dependency graph, we were looking for associated items in the attributes of the item. But associated items can also be held by attributes of subparts of the item (like variants in the case of a struct).

### DIFF
--- a/engine/lib/ast_utils.ml
+++ b/engine/lib/ast_utils.ml
@@ -655,6 +655,13 @@ module Make (F : Features.T) = struct
               self#zero
           | _ -> super#visit_expr' () e
       end
+
+    let collect_attrs =
+      object (_self)
+        inherit [_] Visitors.reduce
+        inherit [_] expr_list_monoid
+        method! visit_attrs () attrs = attrs
+      end
   end
 
   (** Produces a local identifier which is locally fresh **with respect
--- a/engine/lib/dependencies.ml
+++ b/engine/lib/dependencies.ml
@@ -88,8 +88,9 @@ module Make (F : Features.T) = struct
         =
       List.concat_map
         ~f:(fun i ->
+          let attrs = U.Reducers.collect_attrs#visit_item () i in
           let assoc =
-            uid_associated_items i.attrs |> List.map ~f:(fun i -> i.ident)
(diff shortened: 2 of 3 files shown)


karthikbhargavan (approved):
This one should be tested on a larger codebase, but the change looks good.

maximebuyse:
https://github.com/cryspen/libcrux/actions/runs/12480081391/job/34830360886 succeeded, let's merge

Issue #485, opened 2024-01-31
Tests: improve testing generally


W95Psp:
Tests are a bit complicated currently.
 - They are coarse-grained: one crate in `tests` is one test. Adding a test should be "free", currently it's quite bothering.
 - We don't have any way of testing the extraction with special flags on, making it impossible to test, i.e., the `-i` flag.

W95Psp:
Still relevant

maximebuyse:
As described in https://github.com/hacspec/hax/pull/1210#issuecomment-2561006649 it would be nice to be able to test the behaviour of hax on unsupported code.

PR #1210, opened 2024-12-23
fix(backend) Filter Impl after adding comments.

### LINKED ISSUE #1202: Interface only mode still produces an fst file if there are errors
With the following rust code: 
```
fn bad<'a>(x: (& 'a mut i32, & 'a mut i32)) -> & 'a mut i32 {
    x.0
}

fn ok() {}
```

Extracting with `cargo hax into -i '+:**' fstar --interfaces '+!**'` produces an fst file and an fsti. The fsti contains:
```
val ok: Prims.unit -> Prims.Pure Prims.unit Prims.l_True (fun _ -> Prims.l_True)
(* item error backend: (DirectAndMut) The mutation of this [1m&mut[0m is not allowed here.
...
*)
```
and the fsti only contains a comment about the error.
This is annoying because F* will complain that `ok` is not implemented in the fst file. It would be better to produce no fst file in this case.

maximebuyse:
Fixes #1202.

This avoids creating an fst file with only comments in it in fsti-only mode. The filtering of `Impl` elements is moved to after treating comments.

### DIFF
--- a/engine/backends/fstar/fstar_backend.ml
+++ b/engine/backends/fstar/fstar_backend.ml
@@ -1608,7 +1608,6 @@ let strings_of_item (bo : BackendOptions.t) m items (item : item) :
     [%matches? (Types.Included None' : Types.inclusion_kind)] interface_mode'
   in
   Print.pitem item
-  |> List.filter ~f:(function `Impl _ when no_impl -> false | _ -> true)
   |> List.concat_map ~f:(function
        | `Impl i -> [ (mk_impl (Print.decl_to_string i), `Newline) ]
        | `Intf i -> [ (mk_intf (Print.decl_to_string i), `Newline) ]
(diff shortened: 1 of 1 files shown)


engine/backends/fstar/fstar_backend.ml
```diff
@@ -1618,6 +1617,7 @@ let strings_of_item (bo : BackendOptions.t) m items (item : item) :
            let s = "(* " ^ s ^ " *)" in
            if interface_mode then [ (`Impl s, `Newline); (`Intf s, `Newline) ]
            else [ (`Impl s, `Newline) ])
+  |> List.filter ~f:(function `Impl _, _ when no_impl -> false | _ -> true)
```
karthikbhargavan, line 1620: Yes, this change looks right, but we need some tests.

karthikbhargavan (approved):
Can we add some tests?

maximebuyse:
> Can we add some tests?

Done

maximebuyse:
> > Can we add some tests?
> 
> Done

Actually the test I added breaks the CI because it tests the running hax on unsupported rust code (where hax produces a result but with an exit code of 1). I think we should merge anyway, but it would be nice to be able to test what happens with unsupported code. I'll add a comment in #485

PR #1212, opened 2024-12-24
fix(CI) Update F* version to fix mlkem CI job 

maximebuyse:
This PR updates the F* version used by the ML-KEM CI job. The job was failing recently (see for example https://github.com/hacspec/hax/actions/runs/12482298578/job/34836176279, because of a syntax error). Using a more recent F* fixes this (https://github.com/hacspec/hax/actions/runs/12483217088/job/34838637301)

### DIFF
--- a/.github/workflows/mlkem.yml
+++ b/.github/workflows/mlkem.yml
@@ -33,7 +33,7 @@ jobs:
           nix profile install ./hax
 
       - name: ⤵ Install FStar
-        run: nix profile install github:FStarLang/FStar/v2024.01.13
+        run: nix profile install github:FStarLang/FStar/v2024.09.05
 
       - name: ⤵ Clone HACL-star repository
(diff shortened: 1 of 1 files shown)


.github/workflows/mlkem.yml
```diff
@@ -33,7 +33,7 @@ jobs:
           nix profile install ./hax
 
       - name: ⤵ Install FStar
-        run: nix profile install github:FStarLang/FStar/v2024.01.13
+        run: nix profile install github:FStarLang/FStar/v2024.09.05
```
franziskuskiefer, line 36: libcrux moved to 2024.12.03 in https://github.com/cryspen/libcrux/pull/729

maximebuyse: I changed to use the same

maximebuyse:
The version is the same as libcrux, let's merge.

PR #1220, opened 2025-01-07
Visit trait goals to rename impl expr they may contain.

### LINKED ISSUE #1218: Naming bug with generic constraint on associated types
```rust
trait T1 {
    type Assoc;
}

trait T2 {}

impl<U: T1> T2 for U where U::Assoc : T2 {}


```
[Open this code snippet in the playground](https://hax-playground.cryspen.com/#fstar+tc/f47b2ae440/gist=c23e295f24de541a0e87120618b620f8)

maximebuyse:
Fixes #1218.

The bug happens when an impl expr is contained inside the trait goal as trait goals are not visited by the visitor. The fix is simply to visit trait goals with the visitor so that impl expr renaming also happens there.


W95Psp (approved):
Great, tahnks, let's merge!

PR #1225, opened 2025-01-13
Hax home page using mkdocs

W95Psp:
This PR drops the mdbook book in favor of a unique entry point website, generated using mkdocs.

This PR:
 - inits a mkdocs static website
 - moves the book
 - restore most of the hax playground interaction features

I fixed some broken links and broken contents, but I suspect there is more!

Also, there is a number of TODO left (let's see if we want those in this PR or not):
 - [x] restore publishing of Rust/OCaml docs
 - [x] move "Manual > Contributing" to "Dev"


franziskuskiefer (changes requested):
Thanks!
Let's drop everything that's not properly populated yet.
- Manual > Examples look empty
- RFCs we could drop. But I'm fine keeping it as is.
- Engine and Frontend should maybe only link to the rust/odocs for now? Those graphs there don't make much sense 😉 
- Keep the blog but drop my test entry there.
- It looks like you dropped the default mkdocs page from dev. But it also had some useful information on how to use mkdocs and dependencies that I think we should keep.

franziskuskiefer:
> * restore most of the hax playground interaction features

Oh and on this I'm not sure what that means. I don't see interaction options. I'm fine with dropping them for now and adding them back later. But we should make clear what's working and what not.

W95Psp:
👍 
Will address those points!

For the interaction, you don't see any "Play" buttons on the right corner of code blocks in e.g. "Tutorial > Panic freedom"?

Capture d’écran 2025-01-13 à 16 40 35

.github/workflows/gh_pages.yml
```diff
@@ -38,75 +37,4 @@ jobs:
...
-        } > README.md
-
-        git add -A
-        if git commit -m "Book update (generated from hacspec/hax@$HAX_COMMIT)"; then
-          git push https://$USERNAME:$PUSH_HACSPEC_GITHUB_TOKEN@github.com/hacspec/book
-        fi
-      env:
-        PUSH_HACSPEC_GITHUB_TOKEN: ${{secrets.PUSH_HACSPEC_GITHUB_TOKEN}}
-        USERNAME: github-actions[bot]
+        uses: actions/deploy-pages@v4 # or specific "vX.X.X" version tag for this action
```
franziskuskiefer, line 40:
```suggestion
        uses: actions/deploy-pages@v4 # or specific "vX.X.X" version tag for this action

```

.github/workflows/gh_pages.yml
```diff
@@ -6,24 +6,23 @@ on:
...
-      uses: actions/upload-pages-artifact@v2
-
-  # deploys the result of `build`
-  # this job is a copy paste from <https://github.com/actions/deploy-pages>
-  deploy-documentation:
-    # Add a dependency to the build job
-    needs: build-documentation
-
-    # Grant GITHUB_TOKEN the permissions required to make a Pages deployment
+      - uses: actions/checkout@v3
```
franziskuskiefer, line 13:
```suggestion
      - uses: actions/checkout@v4
```

docs/frontend/cli/index.md
```diff
@@ -0,0 +1,3 @@
+# Cli
```
franziskuskiefer, line 1: There's nothing in here. Let's drop?

docs/frontend/index.md
```diff
@@ -0,0 +1,3 @@
+# Frontend
+
```
franziskuskiefer, line 2: The second link here is dead.

franziskuskiefer (approved):
lgtm with the few nits addressed.

I'm seeing warnings
```
WARNING -  Doc file 'engine/index.md' contains a link './docs/hax-engine/index.html', but the
           target 'engine/docs/hax-engine/index.html' is not found among documentation files.
WARNING -  Doc file 'frontend/index.md' contains a link
           './docs/hax_frontend_exporter/index.html', but the target
           'frontend/docs/hax_frontend_exporter/index.html' is not found among documentation
           files.
```

docs/frontend/cli/index.md
W95Psp: 👍

mkdocs.yml
```diff
@@ -0,0 +1,76 @@
+site_name: hax
```
franziskuskiefer, line 1: We probably need a `repo_url` here for deployment.

W95Psp:
The warnings are because those paths are generated via rustdoc and odoc: this happens only in the CI

docs/frontend/index.md
W95Psp: It's dead only locally, the CI populates it with (rustdoc / odoc) contents.

